PR #31, opened 2021-06-15
feat: add upsert object handler

ankitjena:

Feature

The create object handler throws duplicate not allowed error, if object already exists

## What is the new behavior?

Adds route
```
POST /object/upsert/:bucketName/*
```
which inserts an object if it doesn't exist. If it does exist it updates the object with new file.

### Screen recording
Peek 2021-06-15 22-23


src/routes/object/upsertObject.ts
```diff
@@ -0,0 +1,209 @@
...
+      } catch (err) {
+        request.log.error(err)
+        return response.status(400).send({
+          statusCode: '400',
+          error: err.message,
+          message: err.message,
+        })
+      }
+
+      const objectResponse = await postgrest
```
inian: Can you use the [upsert function](https://supabase.io/docs/reference/javascript/upsert) of postgrest? Saves us a API call

inian (changes requested):

inian:
Hi Ankit, thank you so much for the PR! Looking at the code, once you use the `upsert` call of postgrest-js, the code will be almost exactly the same as `createObject`. 

What do you think about adding upsert functionality into createObject itself? By default, we won't upsert but if there is a header called `x-upsert: true`, we upsert instead of insert. 

Another advantage of this, we wont need to create a [new API endpoint](https://github.com/supabase/storage-api/pull/31/files#diff-0c165bea0a64be46e0ee17e68fa7ff006cb57e7b69b5dabb62ff23ac14bdc524R41). The API endpoint `/upsert/:bucketName/*` will probably clash with the current endpoint if the user has a bucket name called `upsert` which is a problem

ankitjena:
Right, I was thinking of the same thing about having upsert inside `createObject`, having a discussion on that would have been better I guess. Also I didn't think of the clash with bucketName for the upsert endpoint. Thanks Inian. I will work on that and make the changes.

inian:
Thanks Ankit :)

> I didn't think of the clash with bucketName for the upsert endpoint.

I learnt this the hard way as well haha https://github.com/supabase/storage-api/issues/22

inian:
Also can you merge in the changes from master Ankit? I just made some changes to the createObject file.

ankitjena:
> Also can you merge in the changes from master Ankit? I just made some changes to the createObject file.

Sure.

@inian I was trying the upsert now, for some reason it doesn't work, looked up the documentation in postgrest-js
image
```
const { data: results, error, status } = await postgrest
        .from<Obj>('objects')
        .insert(
          [
            {
              name: objectName,
              owner: owner,
              bucket_id: bucketName,
            },
          ],
          {
            returning: 'minimal',
            upsert: true,
          }
        )
        .single()
```
I get the same error
```
{
  "statusCode": "23505",
  "error": "Key (bucket_id, name)=(avatars, folder/cat-new.png) already exists.",
  "message": "duplicate key value violates unique constraint \"bucketid_objname\""
}
```

ankitjena:
@inian Any suggestion for where I should look at for resolving this?

ankitjena:
Apparently, the postgrest client wasn't exposing the upsert method in query builder, I upgraded the package and using it worked. I was trying out the deprecated insert operation with upsert sadly.

ankitjena:
I messed up the merge somehow, closing this PR and creating another one.

PR #44, opened 2021-07-13
fix: #43 use custom content type parser for application/json and text/plain

thebengeu:

Pipe the request stream for these content types instead of the default handling. See https://www.fastify.io/docs/latest/ContentTypeParser/.

`text/plain` and `application/json` MIME type are uploaded in Storage Buckets as empty files.

## What is the new behavior?

`text/plain` and `application/json` MIME type are uploaded in Storage Buckets correctly.


inian:
Nice @thebengeu 🔥

inian (approved):

PR #59, opened 2021-07-29
feat: use X-Forwarded-Host to construct PostgREST URL for multitenant setups

thebengeu:
Depends on #58.


.env.sample
```diff
@@ -10,3 +10,9 @@ PGOPTIONS="-c search_path=storage"
 FILE_SIZE_LIMIT=52428800
 STORAGE_BACKEND=s3
 FILE_STORAGE_BACKEND_PATH=./data
+
+# Multitenant
+X_FORWARDED_HOST_REGEXP=
+SUPABASE_URL=
```
inian: reminder to use the read replica URL here since we are only reading always

inian:
Looks good Beng! 🔥 Can you add some tests for this too?

src/utils/project.ts
```diff
@@ -0,0 +1,78 @@
...
+    throw new Error('SUPABASE_API_KEY environment variable is not set')
+  }
+  const supabase = createClient(supabaseUrl, supabaseApiKey)
+  const {
+    data: {
+      jwt_secret_encrypted,
+      services: [{ service_api_keys }],
+    },
+  } = await supabase
+    .from('projects')
```
inian: hmm this couples the storage-api code to our internal middleware schema, lets see if we can come up with a better abstraction for this later on for those who want to run self hosted storage in multitenant mode

.env.sample
thebengeu: Renamed SUPABASE_URL to SUPABASE_READ_ONLY_URL for clarity.

src/utils/project.ts
thebengeu: As discussed offline, for a better abstraction, I'll add endpoints to add and delete per-tenant config.

thebengeu: Done, please take a look.

thebengeu:
@inian To get CI to pass, could you add `TENANT_ID` as a secret, with the same value as `PROJECT_REF`?

migrations/multitenant/0001-initialmigration.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS storage.tenants (
+  id text PRIMARY KEY,
+  config jsonb NOT NULL
```
inian: Can we have explicit columns here Beng? Will be easier to type etc.  

And for the multitenant db, this can be in the public schema itself right? If you want to create it in a storage schema, where is the schema created?

src/routes/tenant/index.ts
```diff
@@ -0,0 +1,114 @@
...
+interface tenantRequestInterface extends RequestGenericInterface {
+  Body: FromSchema<typeof schema.body>
+  Params: {
+    tenantId: string
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export default async function routes(fastify: FastifyInstance) {
+  fastify.get('/', async () => {
```
inian, line 32: How are these admin routes protected? Can we have a simple key based auth here?

src/routes/tenant/index.ts
```diff
@@ -0,0 +1,114 @@
...
+    const result = await pool.query(
```
inian, line 33: Add a proper error handling here? Eg, if the db is not available etc

src/utils/tenant.ts
```diff
@@ -0,0 +1,79 @@
...
+
+interface TenantConfig {
+  anonKey: string
+  databaseUrl: string
+  jwtSecret: string
+  serviceKey: string
+}
+
+const tenantConfigCache: {
+  [tenantId: string]: TenantConfig
```
inian: Make this into a Map instead of an object? Faster and more explicit.

src/utils/tenant.ts
```diff
@@ -0,0 +1,79 @@
...
+    SELECT
+      id,
+      config
+    FROM
+      storage.tenants
+    `
+  )
+  for (const tenant of result.rows) {
+    const { id, config } = tenant
+    await cacheTenantConfigAndRunMigrations(id, config)
```
inian: Should we do this in batches of 100 or so? Connecting to 1000s of dbs individually and running migrations sequentially will take a considerable amount of time

src/test/db/docker-compose.yml
```diff
@@ -25,3 +25,13 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
       POSTGRES_PORT: 5432
+  multitenant_db:
+    image: postgres:12
```
inian: lets use 13

src/test/db/multitenant/00-storage-schema.sql
```diff
@@ -0,0 +1 @@
+CREATE SCHEMA IF NOT EXISTS storage;
```
inian: ah this should be in /migrations/multitenant/0001-initialmigration.sql (or just use public schema)

inian:
> @inian To get CI to pass, could you add TENANT_ID as a secret, with the same value as PROJECT_REF?

Done

src/test/db/multitenant/00-storage-schema.sql
thebengeu: Switched to public schema.

migrations/multitenant/0001-initialmigration.sql
thebengeu: Switched to explicit columns and public schema.

src/routes/tenant/index.ts
thebengeu: Added auth based on `apikey` header.

src/utils/tenant.ts
thebengeu: Done.

src/utils/tenant.ts
thebengeu: Done.

src/test/db/docker-compose.yml
thebengeu: Done.

src/routes/tenant/index.ts
thebengeu: This is an internal endpoint, we can just let it throw and be handled by default error handling, which does log uncaught errors even in production mode.

.github/workflows/ci.yml
```diff
@@ -36,16 +36,19 @@ jobs:
...
-          echo PROJECT_REF=${{ secrets.PROJECT_REF }} >> .env
+          echo TENANT_ID=${{ secrets.TENANT_ID }} >> .env
           echo REGION=${{ secrets.REGION }} >> .env
           echo POSTGREST_URL=${{ secrets.POSTGREST_URL }} >> .env
           echo GLOBAL_S3_BUCKET=${{ secrets.GLOBAL_S3_BUCKET }} >> .env
           echo PGRST_JWT_SECRET=${{ secrets.PGRST_JWT_SECRET }} >> .env
           echo AUTHENTICATED_KEY=${{ secrets.AUTHENTICATED_KEY }} >> .env
           echo DATABASE_URL=postgresql://postgres:postgres@127.0.0.1/postgres >> .env
-          echo PGOPTIONS='-c search_path=storage' >> .env
+          echo PGOPTIONS='-c search_path=storage,public' >> .env
```
darora: could be cleaned up by specifying the entire file in a heredoc

.github/workflows/ci.yml
```diff
@@ -36,16 +36,19 @@ jobs:
...
           echo FILE_SIZE_LIMIT=52428800 >> .env
           echo STORAGE_BACKEND=s3 >> .env
+          echo MULTITENANT_DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:5433/postgres >> .env
+          echo POSTGREST_URL_SUFFIX=/rest/v1 >> .env
+          echo API_KEY=apikey >> .env
```
darora: `ADMIN_API_KEY`?

src/plugins/apikey.ts
```diff
@@ -0,0 +1,11 @@
+import fastifyPlugin from 'fastify-plugin'
+import { getConfig } from '../utils/config'
+
+export default fastifyPlugin(async (fastify) => {
+  const { apiKey } = getConfig()
+  fastify.addHook('onRequest', async (request, reply) => {
+    if (request.headers.apikey !== apiKey) {
```
darora: Suggest making this a set membership check rather than equality; gives us a cheap mechanism to rotate these

src/plugins/postgrest.ts
```diff
@@ -1,16 +1,57 @@
...
+    superUserPostgrest: PostgrestClient
   }
 }
 
-export default fastifyPlugin(async (fastify) => {
+async function getPostgrestClient(request: FastifyRequest, jwt: string): Promise<PostgrestClient> {
+  const { anonKey, postgrestURL, postgrestURLSuffix, xForwardedHostRegExp } = getConfig()
+  let url = postgrestURL
+  let apiKey = anonKey
+  if (xForwardedHostRegExp) {
```
darora: can we test `isMultitenant` here to make the intent clearer? That's what we're using the current test as a proxy for, yeah?

src/plugins/tenant-id.ts
```diff
@@ -0,0 +1,21 @@
...
+  interface FastifyRequest {
+    tenantId: string
+  }
+}
+
+export default fastifyPlugin(async (fastify) => {
+  const { tenantId, xForwardedHostRegExp } = getConfig()
+  fastify.decorateRequest('tenantId', tenantId)
+  fastify.addHook('onRequest', async (request) => {
+    if (!xForwardedHostRegExp) return
```
darora: same suggestion of testing on `isMultitenant`

.github/workflows/ci.yml
thebengeu: Switched to using the `env` context instead, which is even cleaner.

.github/workflows/ci.yml
thebengeu: Done.

src/plugins/apikey.ts
thebengeu: Done.

src/plugins/postgrest.ts
thebengeu: The current test also serves as truthiness narrowing since `xForwardedHostRegExp` may be undefined. But adding `isMultitenant` does make the intent clearer, done.

src/plugins/tenant-id.ts
thebengeu: Done.

inian:
Lets encrypt the JWT secret, anon and service keys too Beng. We already do that in our middleware db.

thebengeu:
Encrypted the JWT secret, anon and service keys.

inian (approved):

PR #66, opened 2021-08-16
Forward 304 and If-Modified-Since/If-None-Match headers

### LINKED ISSUE #60: Handle If-Modified-Since or If-None-match
Right now we send cache-control with a max-age header. This is enough if there is a CDN in front of our API which take care of handling 304s if the object hasn't changed. 

But since we don't have a CDN yet, we need to implement this logic ourselves. The fix should be pretty simple - just pass the If-Modified-Since or If-None-match headers from the request on to S3. S3 Should send back a 304 based on that which we can stream on to the client.

thebengeu:
Resolves #60.

### DIFF
--- a/src/routes/object/getObject.ts
+++ b/src/routes/object/getObject.ts
@@ -41,8 +41,6 @@ async function requestHandler(
     unknown
   >
 ) {
-  const range = request.headers.range
-
   const { bucketName } = request.params
   const objectName = request.params['*']
 
@@ -71,7 +69,11 @@ async function requestHandler(
   const s3Key = `${projectRef}/${bucketName}/${objectName}`
   request.log.info(s3Key)
   try {
-    const data = await storageBackend.getObject(globalS3Bucket, s3Key, range)
+    const data = await storageBackend.getObject(globalS3Bucket, s3Key, {
+      ifModifiedSince: request.headers['if-modified-since'],
+      ifNoneMatch: request.headers['if-none-match'],
+      range: request.headers.range,
+    })
 
     response
       .status(data.metadata.httpStatusCode ?? 200)
@@ -84,6 +86,9 @@ async function requestHandler(
     }
     return response.send(data.body)
   } catch (err) {
+    if (err.$metadata?.httpStatusCode === 304) {
+      return response.status(304).send()
+    }
     request.log.error(err)
     return response.status(400).send(createResponse(err.message, '400', err.name))
   }
--- a/src/backend/file.ts
+++ b/src/backend/file.ts
@@ -30,7 +30,7 @@ export class FileBackend implements GenericStorageBackend {
     return xattr.set(file, attribute, value)
   }
 
-  async getObject(bucketName: string, key: string, range?: string): Promise<ObjectResponse> {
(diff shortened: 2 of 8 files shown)


inian (approved):

src/routes/object/getObject.ts
```diff
@@ -84,6 +86,9 @@ async function requestHandler(
     }
     return response.send(data.body)
   } catch (err) {
+    if (err.$metadata?.httpStatusCode === 304) {
```
inian, line 89: Does AWS sdk v2 also behave like this? Any chance of fixing it upstream? If not, add a comment here that aws sdk throws for 304

inian:
Amazing Beng!

src/routes/object/getObject.ts
thebengeu: I made a comment in aws/aws-sdk-js-v3#2635 questioning whether 304 should be thrown as an exception. However, I tested AWS SDK v2:

```js
const AWS = require('aws-sdk')
const s3 = new AWS.S3()
s3.getObject({Bucket: '...', Key: '...', IfNoneMatch: '...'}, (err) => console.log(err))
```

And got error:

```
NotModified: null
    at Request.extractError (/Users/beng/supabase/node_modules/aws-sdk/lib/services/s3.js:694:35)
    at Request.callListeners (/Users/beng/supabase/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/Users/beng/supabase/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/Users/beng/supabase/node_modules/aws-sdk/lib/request.js:688:14)
    at Request.transition (/Users/beng/supabase/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/beng/supabase/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/beng/supabase/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/Users/beng/supabase/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/Users/beng/supabase/node_modules/aws-sdk/lib/request.js:690:12)
    at Request.callListeners (/Users/beng/supabase/node_modules/aws-sdk/lib/sequential_executor.js:116:18) {
  code: 'NotModified',
  region: null,
  time: 2021-08-17T06:13:24.119Z,
  requestId: 'GZXPQZGF4FKQC3XY',
  extendedRequestId: '7fXRjHwQPkd8rD80mFhRj6s/y/gxhTp7Z4fsK6fI34TdvXl3J4giEEZgkAczYZzUm9WTGL21Mr0=',
  cfId: undefined,
  statusCode: 304,
  retryable: false,
  retryDelay: 6.728630404646108
}
```

Since this behaviour is consistent between AWS SDK v2 and v3, I don't think it'll be fixed upstream, and don't think it warrants a comment since it's evident that AWS SDK throws for 304?

inian: oh its the same as v2..yeah then I dont think this will be fixed..

PR #69, opened 2021-08-24
fix: install necessary packages for node-gyp, use multi-stage build

thebengeu:


Dockerfile
```diff
@@ -1,10 +1,21 @@
-FROM mhart/alpine-node:14
+FROM mhart/alpine-node:16
```
darora: Consider using the official `node:16-alpine3` images

thebengeu: Done.

inian: lets upgrade to 16 in a separate PR? I really want to upgrade to 16 too (native M1 node builds 🎉 ), but our [CI](https://github.com/supabase/storage-api/blob/master/.github/workflows/ci.yml) and [build](https://github.com/supabase/storage-api/blob/be6d3f5b1923594bda1dbad6ea21eae97ea51dad/.github/workflows/build.yml#L49) needs to be updated too. If not, our self-hosted version will be based on node16 and prod will be in node14 and there might be a mismatch there.

thebengeu: Switched back to 14.

inian (approved):

inian:
Good catch Beng! `fs-xattr` is a native module which requires node-gyp

PR #75, opened 2021-09-16
Support per-tenant file size limit if multitenant

thebengeu:


inian (approved):

inian:
Good work removing the postgrest dependency too @thebengeu 🔥

PR #77, opened 2021-10-06
fix extension function by getting the last item of a split

### LINKED ISSUE #12: Fix extension function
Right now the `extension` function [returns the second part](https://github.com/supabase/storage-api/blob/514899eb2d0f1823abc95d497f2e801978cc49ff/src/test/db/02-storage-schema.sql#L74) of the filename after splitting by `.`. 

Instead the last part should be returned so that the extension for the file `cat.meme.jpg` is returned as `jpg` and not as `meme`.

bariqhibat:
Signed-off-by: Bariq <bariqhibat@gmail.com>

Bug fix, feature, docs update, ...
Bug fix #12

We assume that the name of the file is `cat.jpg`, getting the extension is by getting the second index of the split by '.'.
However, files can have multiple '.', hence this PR introduces a more general implementation.

Please link any relevant issues here.

## What is the new behavior?

Reverse the string, split it by '.' and get the first item, reverse the string again. We'll get the last index of the split '.'.

Feel free to include screenshots if it includes visual changes.

## Additional context

### DIFF
--- a/src/test/db/02-storage-schema.sql
+++ b/src/test/db/02-storage-schema.sql
@@ -71,7 +71,7 @@ BEGIN
 	select string_to_array(name, '/') into _parts;
 	select _parts[array_length(_parts,1)] into _filename;
 	-- @todo return the last part instead of 2
-	return split_part(_filename, '.', 2);
+	return reverse(split_part(reverse(_filename), '.', 1));
 END
 $function$;
 


inian (approved):

inian:
Thanks @bariqhibat 🚀

PR #78, opened 2021-10-07
feat: setup husky for commit lint

### LINKED ISSUE #64: add a precommit hook to validate commit messages
it's annoying that commits dont get picked up by our semantic-commit-analyser plugin and they are not reflected properly in our changelog. our version number is also not bumped and we need to do fake commits like [this](https://github.com/supabase/storage-api/commit/87d45fdb0a3930bdf921bc5b54af1e8ca833ad03). 

We should have a pre commit hook (via [husky](https://www.npmjs.com/package/husky)?) to validate commit messages adhere to the [right format](https://github.com/semantic-release/commit-analyzer#default-rules-matching)

97amarnathk:

* add a precommit hook to validate commit messages (using husky)

Bug fix https://github.com/supabase/storage-api/issues/64

https://github.com/supabase/storage-api/issues/64

## What is the new behavior?
git precommit hook lints the commit message

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,0 +1,6 @@
+{
+  "extends": ["@commitlint/config-conventional"],
+  "rules": {
+      "type-enum": [2, "always", ["ci", "chore", "docs", "feat", "fix", "perf", "refactor", "revert", "style"]]
+  }
+}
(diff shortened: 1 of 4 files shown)


ferhatelmas:
we need CI enforcement otherwise not needed

PR #79, opened 2021-10-11
Prevent empty results when selecting folder(s) in one bucket that match an object name in another bucket

thebengeu:
Given a bucket "a"  and an object name "123/456", and a bucket "b" and an object/folder name "123", when selecting bucketname="a" and prefix="":

Before the fix:
If the latter doesn't exist, a row with only name="123" populated and the other columns being NULL is returned.
If the latter exists, name="123" is left joined with the row with object/folder name "123", and then filtered out by the `WHERE ... objects.bucket_id=bucketname` clause.

After the fix:
The row with object/folder name "123" isn't left joined, so a row with only name="123" populated and the other columns being NULL is returned correctly.

### DIFF
--- a/migrations/tenant/0008-fix-search-function.sql
+++ b/migrations/tenant/0008-fix-search-function.sql
@@ -1,5 +1,3 @@
-alter table storage.objects add column path_tokens text[] generated always as (string_to_array("name", '/')) stored;
-
 CREATE OR REPLACE FUNCTION storage.search(prefix text, bucketname text, limits int DEFAULT 100, levels int DEFAULT 1, offsets int DEFAULT 0)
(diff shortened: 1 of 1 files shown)


migrations/tenant/0008-fix-search-function.sql
```diff
@@ -0,0 +1,29 @@
+alter table storage.objects add column path_tokens text[] generated always as (string_to_array("name", '/')) stored;
```
inian: this column already exists

thebengeu: Oops, had copied it over from 0002-pathtoken-column.sql. Removed.

PR #97, opened 2022-02-08
Do not cache tenant config on migration error

thebengeu:


inian (approved):

src/utils/tenant.ts
```diff
@@ -14,9 +14,19 @@ const tenantConfigCache = new Map<string, TenantConfig>()
...
-  config: TenantConfig
+  config: TenantConfig,
+  logOnError = false
 ): Promise<void> {
-  await runMigrationsOnTenant(config.databaseUrl)
+  try {
+    await runMigrationsOnTenant(config.databaseUrl)
+  } catch (error) {
+    if (logOnError) {
+      console.error('Migration error:', error.message)
```
darora, line 24: this...gets rid of the stacktrace, no?

thebengeu: These are expected errors, such as `Migration error: timeout expired`, `Migration error: getaddrinfo ENOTFOUND db.*.supabase.co`. I haven't seen an instance of an unexpected error yet, I think the stack trace would just be noise?

PR #103, opened 2022-02-16
feat: add endpoint to get signed URLs

thebengeu:
@inian I'm using `prefixes` for consistency with the other batch API: https://github.com/supabase/storage-api/blob/fd072d448f3d860ad5f404406c7f09c025826206/src/routes/object/deleteObjects.ts#L31. However, we aren't using those strings as prefixes there, and aren't doing so here either. Do you recall why they're named `prefixes`, should we rename to say, `paths`, and be inconsistent with the delete objects endpoint?

### DIFF
--- a/src/routes/object/getSignedURLs.ts
+++ b/src/routes/object/getSignedURLs.ts
@@ -15,7 +15,7 @@ const getSignedURLsBodySchema = {
   type: 'object',
   properties: {
     expiresIn: { type: 'integer', minimum: 1, example: 60000 },
-    prefixes: {
+    paths: {
       type: 'array',
       items: { type: 'string' },
       minItems: 1,
@@ -23,14 +23,14 @@ const getSignedURLsBodySchema = {
       example: ['folder/cat.png', 'folder/morecats.png'],
     },
   },
-  required: ['expiresIn', 'prefixes'],
+  required: ['expiresIn', 'paths'],
 } as const
 const successResponseSchema = {
   type: 'array',
   items: {
     type: 'object',
     properties: {
-      prefix: {
+      path: {
         type: 'string',
         example: 'folder/cat.png',
       },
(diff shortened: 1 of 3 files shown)


src/routes/object/getSignedURLs.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      const objectResponse = await request.postgrest
+        .from<Obj>('objects')
+        .select('name')
+        .eq('bucket_id', bucketName)
+        .in('name', prefixes)
+
+      if (objectResponse.error) {
+        const { error, status } = objectResponse
+        request.log.error({ error }, 'error object')
```
darora: better error msg plz

src/routes/object/getSignedURLs.ts
```diff
@@ -0,0 +1,119 @@
...
+              'Non-existent or no access'
+            )
+          )
+      }
+
+      const jwtSecret = await getJwtSecret(request.tenantId)
+      const signedUrls = await Promise.all(
+        prefixes.map(async (prefix) => {
+          const urlToSign = `${bucketName}/${prefix}`
+          request.log.info(`going to sign ${urlToSign}`)
```
darora: Hmm this is going to get very spammy - let's take it out?

darora (approved):

src/routes/object/getSignedURLs.ts
thebengeu: This was for consistency with https://github.com/supabase/storage-api/blob/fd072d448f3d860ad5f404406c7f09c025826206/src/routes/object/getSignedURL.ts#L76, but yes it'd be even more spammy for batch calls, removed.

src/routes/object/getSignedURLs.ts
thebengeu: Done.

thebengeu:
Went ahead to rename `prefixes` to `paths`, unless anyone has objections.

inian (approved):

inian:
@thebengeu prefix is a bit of a s3 terminology here where everything is a prefix and a delimiter. we can change it to paths to be more explicit.

PR #123, opened 2022-03-16
feat: use PM2 in `Dockerfile`

thebengeu:
As a temporary mitigation for nodejs/node#39671.

### DIFF
--- a/ecosystem.config.js
+++ b/ecosystem.config.js
@@ -1,0 +1,11 @@
+module.exports = {
+  apps: [
+    {
+      env: {
+        NODE_ENV: 'production',
+      },
+      exp_backoff_restart_delay: 100,
+      script: 'dist/server.js',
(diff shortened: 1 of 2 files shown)


ecosystem.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  apps: [
+    {
+      env: {
+        NODE_ENV: 'production',
+      },
+      exp_backoff_restart_delay: 100,
```
darora: Given that we're adding this to recover from the SDK-triggered bug, wouldn't it make more sense to have a constant delay rather than the exp backoff?

darora (approved):

ecosystem.config.js
thebengeu: Yeah, that makes sense.

Issue #124, opened 2022-03-16
List endpoint to return folders first, followed by files


joshenlim:

alaister:
Solved in #127

Issue #88, opened 2022-01-18
Add support for search in bucket


joshenlim:
# Feature request

Relevant PR: https://github.com/supabase/supabase/pull/4961

Searching for storage in the dashboard at the moment is a client side search - we do a name string search against what was already fetched from the API

As we're looking to introduce infinite scrolling to the dashboard, this means that we won't (be attempting to) pull all of the folder contents when the user clicks on a folder, hence we need to shift the search logic to the API level

Searching should also support pagination as well (we can just stick with offset limit as per the list items API)

alaister:
Solved in #127

PR #127, opened 2022-03-21
feat: add ability to search files and folders by prefix

alaister:
Adds an optional `search` parameter to the search function, which can be used to search for files and folders within a given folder (`prefix`).

This new function is compatible (other than the order in which items are returned) with the previous version, so there shouldn't be any breaking changes.

Additionally, sorts by folders first, then files.

### DIFF
--- a/migrations/tenant/0009-search-files-search-function.sql
+++ b/migrations/tenant/0009-search-files-search-function.sql
@@ -1,0 +1,69 @@
+drop function storage.search;
+
+create or replace function storage.search (
+  prefix text,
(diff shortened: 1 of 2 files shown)


alaister:
Solves #88 and #124

inian (approved):
Wohoo nice one @alaister 🎉

joshenlim:
Awesome! Yup i'll give it a go locally first to test the functionality too 🚀

thebengeu (approved):

PR #136, opened 2022-04-12
Fix : Updating any object in storage is not updating update time

### LINKED ISSUE #67: Last modified timestamp will not update after update
# Bug report

## Describe the bug

A replacment of a file with upload (upsert: true) or update will not update the last modified timestamp. The file itself has been updated.

## To Reproduce

```
const avatarFile = event.target.files[0]
const { data, error } = await supabase
  .storage
  .from('avatars')
  .update('public/avatar1.png', avatarFile, {
    cacheControl: 3600,
    upsert: false
  })
```

## Expected behavior

A replacement of a file should update the timestamp.

## System information

supabase-js 1.22.0

rahul3v:
As a fix, It will now update the update time of an object, if any update occurs to an object in storage.
fix  #67

### DIFF
--- a/src/routes/object/createObject.ts
+++ b/src/routes/object/createObject.ts
@@ -224,6 +224,7 @@ export default async function routes(fastify: FastifyInstance) {
       const { error: updateError, status: updateStatus } = await request.superUserPostgrest
         .from<Obj>('objects')
         .update({
+          updated_at: new Date().toISOString(),
           metadata,
         })
         .match({ bucket_id: bucketName, name: objectName })
--- a/src/routes/object/createObject.ts
+++ b/src/routes/object/createObject.ts
@@ -224,7 +224,6 @@ export default async function routes(fastify: FastifyInstance) {
       const { error: updateError, status: updateStatus } = await request.superUserPostgrest
         .from<Obj>('objects')
         .update({
-          updated_at: new Date().toISOString(),
           metadata,
         })
         .match({ bucket_id: bucketName, name: objectName })
--- a/src/routes/object/updateObject.ts
+++ b/src/routes/object/updateObject.ts
@@ -162,6 +162,7 @@ export default async function routes(fastify: FastifyInstance) {
       const { error: updateError, status: updateStatus } = await request.postgrest
         .from<Obj>('objects')
         .update({
+          updated_at: new Date().toISOString(),
           metadata,
         })
         .match({ bucket_id: bucketName, name: objectName })
--- a/src/routes/object/updateObject.ts
+++ b/src/routes/object/updateObject.ts
@@ -162,7 +162,6 @@ export default async function routes(fastify: FastifyInstance) {
       const { error: updateError, status: updateStatus } = await request.postgrest
         .from<Obj>('objects')
         .update({
-          updated_at: new Date().toISOString(),
           metadata,
         })
         .match({ bucket_id: bucketName, name: objectName })
--- a/migrations/tenant/0010-add-trigger-to-auto-update-updated_at-column.sql
+++ b/migrations/tenant/0010-add-trigger-to-auto-update-updated_at-column.sql
@@ -1,0 +1,9 @@
+CREATE OR REPLACE FUNCTION update_updated_at_column() 
+RETURNS TRIGGER AS $$
+BEGIN
+    NEW.updated_at = now();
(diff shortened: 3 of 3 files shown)


src/routes/object/createObject.ts
```diff
@@ -224,6 +224,7 @@ export default async function routes(fastify: FastifyInstance) {
       const { error: updateError, status: updateStatus } = await request.superUserPostgrest
         .from<Obj>('objects')
         .update({
+          updated_at: new Date().toISOString(),
```
thebengeu: I believe that the default of `now()` should handle this on creation: https://github.com/supabase/storage-api/blob/12a12582febf54f5bdb36622344d2046666afa02/src/test/db/02-storage-schema.sql#L27?

src/routes/object/updateObject.ts
```diff
@@ -162,6 +162,7 @@ export default async function routes(fastify: FastifyInstance) {
       const { error: updateError, status: updateStatus } = await request.postgrest
         .from<Obj>('objects')
         .update({
+          updated_at: new Date().toISOString(),
```
thebengeu: A `BEFORE UPDATE` SQL trigger would be more robust, so that the column would be updated regardless of where we update from. Would you be ok with adding a migration under `migrations/tenant` to do that?

thebengeu (changes requested):
Thanks @rahul3v for your PR! I suggested an alternative approach.

src/routes/object/createObject.ts
rahul3v: I believe, we are calling an `upsert` cmd as well, that might be the case when user  just replaced the existing object, so it's batter to have that way, hope that make sense

src/routes/object/updateObject.ts
rahul3v: I guess, the SQL triggers are in preview for now, so I didn't check on that, 
And about "migration under `migrations/tenant` ",  Let me check with that if I can or not :)

src/routes/object/createObject.ts
thebengeu: You're right, I misread this as an `insert`.

src/routes/object/updateObject.ts
thebengeu: I meant something like this: https://www.revsys.com/tidbits/automatically-updating-a-timestamp-column-in-postgresql/.

rahul3v: @thebengeu , I didn't get the `migrations/tenant` part, might need your help, if you can do that, that would be great, (I didn`t understand how it's integrated, needed bit more diging),  or 
You can update it later, once the SQL trigger becomes production ready. what you think !!

rahul3v: I can write PSQL query

```sql
CREATE OR REPLACE FUNCTION update_modified_column() 
RETURNS TRIGGER AS $$
BEGIN
    NEW.updated_at = now();
    RETURN NEW; 
END;
$$ language 'plpgsql';

CREATE TRIGGER update_object_updated_time BEFORE UPDATE ON storage.objects FOR EACH ROW EXECUTE PROCEDURE  update_modified_column();
```

just check it once

thebengeu: Looks good to me, just a couple of nits:

- `plpgsql` doesn't have to be in quotes.
- Should be `updated_at` instead of `modified`.
- Maybe `update_objects_updated_at` instead of `update_object_updated_time`?

rahul3v:
```sql
CREATE OR REPLACE FUNCTION update_updated_at_column() 
RETURNS TRIGGER AS $$
BEGIN
    NEW.updated_at = now();
    RETURN NEW; 
END;
$$ language plpgsql;

CREATE TRIGGER update_objects_updated_at BEFORE UPDATE ON storage.objects FOR EACH ROW EXECUTE PROCEDURE update_updated_at_column();
```

Hope that's good, can you suggest any name for it, so that I can add as that name

what you think about 
 `auto-update-updated_at-column-with-trigger.sql`
or
 `add-trigger-to-auto-update-updated_at-column.sql`

thebengeu (approved):

src/routes/object/updateObject.ts
thebengeu: Yeah that's great, either works!

PR #137, opened 2022-04-14
fix: remove `under-pressure` `maxEventLoopUtilization` check

thebengeu:
In a recent incident, the `maxEventLoopUtilization` check led to 503's being returned for all requests even though the requests were being processed, albeit intermittently, in spurts of every half to one second. We should instead continue to process requests and rely on Auto Scaling. Without any checks, `under-pressure` is simply exposing a `/status` route, but it ain't broke.


alaister (approved):
I'm a fan of ain't broke! 👍

PR #139, opened 2022-04-16
Fix Dockerfile && Fix @aws-sdk versions

philanton:
Fixes:
- outdated format of Dockerfile
- outdated versions of @aws-sdk that don't accept GCP Storage

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ FROM node:14-alpine
 RUN npm install -g pm2
 WORKDIR /app
 COPY migrations migrations
-COPY ecosystem.config.js package.json .
+COPY ecosystem.config.js package.json ./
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -15,7 +15,7 @@ FROM node:14-alpine
 RUN npm install -g pm2
 WORKDIR /app
 COPY migrations migrations
-COPY ecosystem.config.js package.json .
+COPY ecosystem.config.js package.json ./
```
thebengeu, line 18: How is `.` outdated? There's no harm using `./`, but I believe using just `.` is common, e.g. there are various places in https://docs.docker.com/develop/develop-images/dockerfile_best-practices/ uses where just `.` is used to refer to the current working directory.

thebengeu (approved):

PR #147, opened 2022-06-15
fix: reduce object deletion batch size when emptying bucket from 500 to 300

thebengeu:
The current batch size can cause 414 URI too long errors when the PostgREST URL for deletion is more than 16K in length. 16K is the default for Kong/NGINX. Each object adds UUID length (36) + "%2C" (3) to the URL. 39*300=12K, well below 16K.


darora (approved):

alaister (approved):
Thanks Beng!

PR #148, opened 2022-06-15
feat: expose metrics endpoint

darora:


thebengeu (approved):

thebengeu:
Fix tests?

darora:
Having a hard time running them locally; keep getting `    error: password authentication failed for user "postgres"` for ~12 tests. Any suggestions? Using a vanilla `.env{,.test}`

darora:
I can see why it might be failing (the app object is built multiple times, which doesn't happen in normal usage), I just can't test fixes to it

darora:
Have a fix for the tests up.

The cleaner thing to do would be to just initialize the app once, though perhaps we're testing that things work as spec'ed across app initializations, so I've done the easier thing of using new registries each time.

thebengeu:
I don't see why we shouldn't initialise the app once in tests when that's what's done in production, but this works too, maybe we could clean it up next time.

PR #149, opened 2022-06-19
feat: support for a customizable request id header

darora:


src/utils/config.ts
```diff
@@ -43,25 +45,27 @@ export function getConfig(): StorageConfigType {
...
+    fileStoragePath: getOptionalConfigFromEnv('FILE_STORAGE_BACKEND_PATH'),
+    globalS3Bucket: getConfigFromEnv('GLOBAL_S3_BUCKET'),
+    globalS3Endpoint: getOptionalConfigFromEnv('GLOBAL_S3_ENDPOINT'),
+    isMultitenant: getOptionalConfigFromEnv('IS_MULTITENANT') === 'true',
+    jwtSecret: getOptionalIfMultitenantConfigFromEnv('PGRST_JWT_SECRET') || '',
+    multitenantDatabaseUrl: getOptionalConfigFromEnv('MULTITENANT_DATABASE_URL'),
+    postgrestURL: getOptionalIfMultitenantConfigFromEnv('POSTGREST_URL') || '',
+    postgrestURLSuffix: getOptionalConfigFromEnv('POSTGREST_URL_SUFFIX'),
+    region: getConfigFromEnv('REGION'),
+    requestIdHeader: getOptionalConfigFromEnv('REQUEST_ID_HEADER') || 'request-id',
```
thebengeu: Ditto.

src/utils/config.ts
```diff
@@ -43,25 +45,27 @@ export function getConfig(): StorageConfigType {
   dotenv.config()
 
   return {
+    adminApiKeys: getOptionalConfigFromEnv('ADMIN_API_KEYS') || '',
+    adminRequestIdHeader: getOptionalConfigFromEnv('ADMIN_REQUEST_ID_HEADER') || 'request-id',
```
thebengeu: We use `x-request-id` elsewhere, I know Fastify's default is `request-id` but should we use `x-request-id` as the default here too for consistency?

thebengeu (approved):

src/utils/config.ts
darora: I'm leaving it as fastify's default to avoid this being a breaking change for anyone else. It's easy enough for us to override it in our internal environment. Could just leave it undefined if preferred.

thebengeu: Good call. As for leaving it undefined, it might make it clear that we're using Fastify's default, but I'm fine either way.

PR #151, opened 2022-06-21
Support for optional S3 credentials

scottymack:

Feature

Unable to connect to non AWS S3 storage that requires the [Credentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/s3clientconfig.html#credentials) object to be included in the `S3ClientConfig`.

## What is the new behavior?
Allow passing in `access key ID` and `secret access key` as optional environment variables and then include them as the `Credentials` object in the `S3ClientConfig` if present.

## Additional context

### DIFF
--- a/src/backend/s3.ts
+++ b/src/backend/s3.ts
@@ -17,7 +17,7 @@ import { GenericStorageBackend, GetObjectHeaders } from './generic'
 export class S3Backend implements GenericStorageBackend {
   client: S3Client
 
-  constructor(region: string, endpoint?: string | undefined) {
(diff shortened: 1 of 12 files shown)


bernardomichel (approved):

thebengeu (approved):

scottymack:
Looks like we need approval for the workflow to run now @thebengeu

thebengeu:
Approved, but looks like there are still conflicts?

scottymack:
Looks like I forgot to run Prettier. Just did that

inian:
Can't this be done by setting the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY env vars @scottymack? I don't think this needs to be explicitly handled in the code.

Issue #125, opened 2022-03-20
Clean up after s3 errors


inian:
If S3 upload is not completed properly, we return the correct error code to the client. But the row in the storage.objects table is not cleaned up.

PR #154, opened 2022-07-05
fix: ensure that newly inserted rows are deleted when uploading fails

bnjmnt4n:

Bug fix

If an error occurs while uploading a new file, the associated row in `storage.objects` still exists even though the file might not exist on the storage backend (#125).

## What is the new behavior?

Any errors thrown when uploading a new file will be caught, and associated rows from `storage.objects` table will be removed.

### DIFF
--- a/src/routes/object/createObject.ts
+++ b/src/routes/object/createObject.ts
@@ -14,7 +14,7 @@ import { createDefaultSchema, createResponse } from '../../utils/generic-routes'
 import { S3Backend } from '../../backend/s3'
 import { FileBackend } from '../../backend/file'
 import { GenericStorageBackend } from '../../backend/generic'
-import { isS3Error } from '../../utils/errors'
+import { StorageBackendError } from '../../utils/errors'
 
 const { region, globalS3Bucket, globalS3Endpoint, storageBackendType } = getConfig()
 let storageBackend: GenericStorageBackend
@@ -179,7 +179,7 @@ export default async function routes(fastify: FastifyInstance) {
           // https://github.com/fastify/fastify-multipart/issues/196#issuecomment-782847791
           isTruncated = data.file.truncated
         } catch (err) {
-          return await handleUploadError(err)
+          return await handleUploadError(err as StorageBackendError)
         }
       } else {
         // just assume its a binary file
@@ -197,7 +197,7 @@ export default async function routes(fastify: FastifyInstance) {
           // @todo more secure to get this from the stream or from s3 in the next step
           isTruncated = Number(request.headers['content-length']) > fileSizeLimit
         } catch (err) {
-          return await handleUploadError(err)
+          return await handleUploadError(err as StorageBackendError)
         }
       }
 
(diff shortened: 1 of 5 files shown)


src/test/object.test.ts
```diff
@@ -270,6 +271,54 @@ describe('testing POST object via multipart upload', () => {
       })
     )
   })
+
+  test('should not add row to database if upload fails', async () => {
```
bnjmnt4n, line 288: Added a test by mocking the upload object function and always causing it to throw.

The test depends on the `list` endpoint to ensure that the row in `storage.objects` is cleaned up. The alternative would be to manually check the PostgREST server, but I thought that might be too much work to include here.

src/routes/object/createObject.ts
```diff
@@ -237,6 +248,47 @@ export default async function routes(fastify: FastifyInstance) {
...
+      async function handleUploadError(uploadError: unknown) {
+        if (uploadError) {
+          request.log.error({ error: uploadError }, 'upload error object')
+
+          let errorName: string
+          let errorStatusCode: number
+          let errorMessage: string | undefined
+
+          if (isS3Error(uploadError)) {
+            errorName = uploadError.name
```
inian: Lets normalise the errors before returning. We shouldn't leak that the backend is s3, file or backblaze, etc to the end user.

src/test/object.test.ts
```diff
@@ -270,6 +271,54 @@ describe('testing POST object via multipart upload', () => {
...
+    jest.spyOn(S3Backend.prototype, 'uploadObject').mockRejectedValue({
+      name: 'S3ServiceException',
+      message: 'Unknown error',
+      $fault: 'server',
+      $metadata: {
+        httpStatusCode: 500,
+      },
+    })
+
+    process.env.FILE_SIZE_LIMIT = '1'
```
inian, line 301: Do we require this? Since you are mocking an error response from s3 anyway

src/test/object.test.ts
inian: We have postgrest running anyway right, can you use postgrest-js to check the database so that we aren't relying on list endpoint's correctness?

inian (changes requested):

src/test/object.test.ts
bnjmnt4n: Updated to use `postgrest-js` to ensure that the row does not exist.

src/test/object.test.ts
bnjmnt4n: I'm not sure if it's strictly required, I'm following the other usages of it in the other upload tests in the file.

src/utils/errors.ts
```diff
@@ -0,0 +1,43 @@
+import { S3ServiceException } from '@aws-sdk/client-s3'
```
inian, line 1: Can we remove this file @bnjmnt4n?

src/test/object.test.ts
inian: That was added add tests for the case where the file size limits is exceeded. It is not required here.

src/utils/errors.ts
bnjmnt4n: What do you mean by remove? Shift into a different file or complete delete the code? Currently the code is being used to convert errors to a consistent format.

inian (approved):

PR #158, opened 2022-07-13
fix: createObject: return proper "413 Payload too large" status code

bnjmnt4n:

Bug fix.

HTTP 400 status code is returned even when the response JSON describes the error as status code 413.

## What is the new behavior?

Return HTTP 413 status code.

### DIFF
--- a/src/routes/object/createObject.ts
+++ b/src/routes/object/createObject.ts
@@ -204,7 +204,7 @@ export default async function routes(fastify: FastifyInstance) {
 
         // return an error response
         return response
-          .status(400)
+          .status(413)
           .send(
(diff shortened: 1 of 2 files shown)


inian:
@bnjmnt4n we always return the http status code as 400 for errors and we can return statusCode inside the response body like we do [here](https://github.com/supabase/storage-api/blob/512b4055c2dda15e36d8ad488ebe29a26638feda/src/routes/object/createObject.ts#L101)

bnjmnt4n:
Oh, I didn't realize that 400 is always returned for errors, even when there's a potentially more specific error code. Is there any reason for that?

PR #159, opened 2022-07-15
feat: bump Node.js version to v16

bnjmnt4n:

Feature.

Node.js v14 was being used.

## What is the new behavior?

Node.js v16 will be used.


darora (approved):
Would love to trigger a release with this

inian (approved):

PR #160, opened 2022-07-18
Deps: upgrade to Fastify v4

bnjmnt4n:

Dependencies upgrade.

Fastify v3 was being used.

## What is the new behavior?

Upgrade Fastify and associated plugins to v4, and other various fixes.

## Additional context

Other noteworthy change:
Previously, the examples were included as an `example` property in each schema object. However, this was invalid JSON Schema, and the appropriate usage to include examples was the `examples` property.


src/admin-app.ts
```diff
@@ -12,9 +12,13 @@ const build = (opts: FastifyServerOptions = {}, adminOpts: AdminOptions = {}): F
   app.register(tenantRoutes, { prefix: 'tenants' })
   app.register(metrics, {
     endpoint: '/metrics',
-    enableRouteMetrics: false,
-    blacklist: ['nodejs_version_info', 'process_start_time_seconds'],
```
bnjmnt4n, line 16: I removed this line, since there didn't seem to be such a configuration option in the new `fastify-metrics` configuration. I believe it's trying to remove some of the default metrics, however I don't think the original configuration worked before either, since it seemed to be for routes to blacklist metrics collection in, and not metrics to disable.

package.json
```diff
@@ -27,22 +27,23 @@
...
-    "fastify-cors": "^6.0.3",
-    "fastify-metrics": "^8.0.0",
-    "fastify-multipart": "^5.3.1",
-    "fastify-plugin": "^3.0.1",
-    "fastify-swagger": "^4.15.0",
+    "fastify": "^4.2.1",
+    "fastify-cors": "^6.1.0",
+    "fastify-metrics": "^9.0.0",
+    "fastify-plugin": "^4.0.0",
+    "fastify-swagger": "^5.2.0",
```
inian: this can be removed now?

inian (approved):

PR #161, opened 2022-07-19
chore: update postgrest to v9.0.0.20220531

inian:
matches the version deployed in prod and takes only 7 MB ram on m1 macs (as opposed to ~6 GB before)


darora (approved):

package.json
```diff
@@ -15,7 +15,7 @@
     "prettier:check": "prettier -c src/**",
     "eslint:check": "eslint 'src/**'",
     "infra:stop": "docker-compose --project-directory . -f src/test/db/docker-compose.yml down  --remove-orphans",
-    "infra:start": "docker-compose --project-directory . -f src/test/db/docker-compose.yml up -d && sleep 5 && npm run migration:run",
+    "infra:start": "docker-compose --project-directory . -f src/test/db/docker-compose.yml up -d && sleep 5 && npm run migration:run && docker-compose --project-directory . -f src/test/db/docker-compose.yml restart rest",
```
darora: why is the restart necessary with this upgrade?

inian: postgrest schema cache is not refreshed - not sure why. Working on giving @steve-chavez a reproducible example

bnjmnt4n: The current SQL setup in this repo doesn't seem to have this [piece of code](https://github.com/supabase/supabase/blob/6668265fece6c876b780fd40f06d9b8e48bc5431/docker/volumes/db/init/03-post-setup.sql#L2-L12) which notifies PostgREST of any DDL commands; maybe it needs to be included?

inian: I think thats it Benjamin. Can you see if adding that fixes it? We just didn't run into that problem before cos the postgrest container took so long to startup under emulation and the new one starts up pretty quickly surfacing error.

bnjmnt4n: @inian updated the PR; seems to works fine locally and on CI as well.

inian: Great thanks Benjamin!

PR #165, opened 2022-07-19
fix: re-enable per-route metrics for collection

darora:


inian (approved):

inian:
We will need to fix the tests

bnjmnt4n:
I think this is blocked on https://github.com/SkeLLLa/fastify-metrics/pull/54.

darora:
Not strictly necessary, we wanted to clean up the multiple initializations of the server in any case to more closely reproduce prod behaviour

PR #169, opened 2022-07-27
feat: store additional content metadata in storage backends

bnjmnt4n:

Feature.

When adding/updating objects, only the content type and cache-control values are stored as metadata in the object backend.

## What is the new behavior?

In addition to the content type and cache-control values being stored in the storage backed, the content encoding, content language and content disposition of the uploaded object are also stored in the storage backend. The `metadata` column of the `storage.objects` table is also updated to store these values.

## Additional context

This is supported by both S3 and Google Cloud Storage:
- S3: https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html
- GCS: https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata

This also partially addresses https://github.com/supabase/storage-api/issues/122 by allowing the content disposition to be set when uploading a storage object.

### DIFF
--- a/src/backend/file.ts
+++ b/src/backend/file.ts
@@ -35,8 +35,14 @@ export class FileBackend implements GenericStorageBackend {
     const file = path.resolve(this.filePath, `${bucketName}/${key}`)
     const body = await fs.readFile(file)
     const data = await fs.stat(file)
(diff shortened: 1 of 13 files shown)


fenos:
We are now storing this metadata

PR #170, opened 2022-07-27
feat(signed URLs): allow customizing response headers

bnjmnt4n:

Feature.

Users are unable to customize the response HTTP headers when creating signed URLs.

## What is the new behavior?

When creating signed URLs, users can customize the response content type and content disposition HTTP headers. 

## Additional context

Both S3 and Google Cloud Storage have this feature:
- S3: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/classes/getobjectcommand.html
  - allows customizing the following headers as well: content-language, expires, cache-control, content-encoding
- GCS:
  - https://cloud.google.com/storage/docs/xml-api/reference-headers#responsecontentdisposition
  - https://cloud.google.com/storage/docs/xml-api/reference-headers#responsecontenttype
  - only allows customizing the response content type and disposition

I've gone with allowing only the response content type and disposition to be customized to go along with the minimum common functionality between these two storage providers.

This addresses https://github.com/supabase/storage-api/issues/122 by allowing the content disposition to be set when requesting a signed URL.

Note: this PR is stacked on #169.

### DIFF
--- a/src/routes/object/getSignedObject.ts
+++ b/src/routes/object/getSignedObject.ts
@@ -67,7 +67,7 @@ export default async function routes(fastify: FastifyInstance) {
       try {
         const jwtSecret = await getJwtSecret(request.tenantId)
         const payload = await verifyJWT(token, jwtSecret)
-        const { url } = payload as SignedToken
+        const { url, contentDisposition, contentType } = payload as SignedToken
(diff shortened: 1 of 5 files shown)


fenos:
Thanks a lot @bnjmnt4n

PR #171, opened 2022-07-30
add HEAD Mehthod to getObjects route

buschco:

Add a `HEAD` Method to `getObject` route so clients can get headers without requesting whole object.

I get a `400` which is ok but a 405 would be better 😂 

## What is the new behavior?

A `HEAD` request just like the `GET` but without the body (see https://github.com/fastify/fastify/pull/2700/)

### DIFF
--- a/src/routes/object/getObject.ts
+++ b/src/routes/object/getObject.ts
@@ -102,6 +102,7 @@ export default async function routes(fastify: FastifyInstance) {
   fastify.get<getObjectRequestInterface>(
     '/authenticated/:bucketName/*',
     {
+      exposeHeadRoute: true,
       // @todo add success response schema here
       schema: {
         params: getObjectParamsSchema,
@@ -120,6 +121,7 @@ export default async function routes(fastify: FastifyInstance) {
   fastify.get<getObjectRequestInterface>(
     '/:bucketName/*',
     {
+      exposeHeadRoute: true,
       // @todo add success response schema here
       schema: {
         params: getObjectParamsSchema,


inian:
Hi @buschco, I am worried that people might do a HEAD request and assume that an object exists at that route when it doesn't. For example /bucket/memes/cat-doesnt-exist.jpg will return a 200 which is confusing. If we are adding a HEAD request for getObject, I would rather only return a 200 if the object exists and the user has permissions to download it.

buschco:
@inian 
is this in the behavior of HEAD request in Fastify (or in general) returning 200 even if the GET would result in 4xx?

I could not find evidence for such behavior in the Fastify PR, am I missing something?

inian:
Ah I just checked and it does look like Fastify returns the same status code as the GET endpoint rather than returning 200 always which is good. 

Can you help check if the [docs exported](https://github.com/supabase/storage-api/blob/9bf39cd2b5f897b55b60a4af7cadc8b5f53b33af/.github/workflows/docs.yml#L36) also have the HEAD method documented @buschco?

buschco:
Unfortunately the `HEAD` route is not in the generated `static/api.json` 😕 

We could raise an inssue in fastify/swagger or keep it a hidden feature.

inian:
Okay, I will merge the PR in. Can you open an issue in fastify/swagger and an issue in our repo so that we remember to add support this in our docs (or ensure this is fixed) after fastify adds support for it? Till then it can be a hidden feature ;)

inian (approved):

buschco:
I raised the issue -> https://github.com/fastify/fastify-swagger/issues/648

thank you 🙂

inian:
hmm the tests are failing for some reason. will merge once I debug why thats happening

PR #182, opened 2022-08-19
fix: add route description

inian:


src/routes/object/getPublicObject.ts
```diff
@@ -42,7 +42,7 @@ export default async function routes(fastify: FastifyInstance) {
       schema: {
         params: getPublicObjectParamsSchema,
         summary,
-        response: { '4xx': { $ref: 'errorSchema#' } },
+        response: { '4xx': { $ref: 'errorSchema#', description: 'Error response' } },
```
darora, line 45: what does this feed into? Is adding a generic 'error response' adding value to a 4xx code?

inian: [This page](https://supabase.com/docs/reference/storage/usage) looks better.

kiwicopple (approved):

PR #185, opened 2022-08-30
chore(logging): improve logging

fenos:

Improvement

- Not all the useful request infos are logged (ex, tenant ID)
- health check requests are logged
- postgrest URL scheme was hard coded to http

## What is the new behavior?

- We now log the tenant id on each request as well as non-sensitive headers
- We now exclude the `/status` endpoints from the requests logs it will reduce the amount of logs produced
- postgrest URL scheme is not an environment variable and can be switched to use HTTPS

### DIFF
--- a/src/monitoring/logger.ts
+++ b/src/monitoring/logger.ts
@@ -8,6 +8,25 @@ export const logger = pino({
       return { level: label }
     },
   },
+  redact: ['req.headers.authorization', 'req.headers.apikey', 'req.headers["api-key"]'],
+  serializers: {
+    res(reply) {
+      return {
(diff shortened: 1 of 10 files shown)


src/monitoring/logger.ts
```diff
@@ -8,6 +8,25 @@ export const logger = pino({
       return { level: label }
     },
   },
+  redact: ['req.headers.authorization', 'req.headers.apikey', 'req.headers["api-key"]'],
```
inian: are we sending api-key somewhere Fabrizio?

inian (approved):

PR #192, opened 2022-09-19
fix: use correct format for last modified date header

ikhsan:

bug fix

Browser (or other client) never cache the images from storage: https://github.com/supabase/storage-api/issues/193

## What is the new behavior?

When setting the `last-modified` header, use `toUTCString` to ensure the date format adheres to [the rfc2616 spec](https://datatracker.ietf.org/doc/html/rfc2616#section-3.3.1)

## Additional context

- I learned that all of these response headers are not yet tested (eg all tests are passing even after I removed all of these lines). This PR mocks the request headers and asserts the response headers
https://github.com/supabase/storage-api/blob/6e2ae7a521db22af37f8d612cdfab6ad00ca4c9f/src/routes/object/getObject.ts#L81-L85

- For running the tests multiple times, it seemed that we can't run them without restarting the whole container. Running jest twice failed for me because I think there is no resets between tests. I ended up running the tests one-by-one as I am only interested on the `GET` requests, so running them multiple times is fine as it doesn't change the db.

- The fix might be premature as I haven't actually confirmed the bug yet. Please check and respond to the [filed issue](https://github.com/supabase/storage-api/issues/193). Happy to close the PR if turns out that it's not an actual bug, happy to be corrected or improved if the bug is valid 👍

### DIFF
--- a/src/routes/object/getObject.ts
+++ b/src/routes/object/getObject.ts
@@ -82,7 +82,7 @@ async function requestHandler(
       .header('Cache-Control', data.metadata.cacheControl)
       .header('ETag', data.metadata.eTag)
       .header('Content-Length', data.metadata.contentLength)
-      .header('Last-Modified', data.metadata.lastModified)
+      .header('Last-Modified', data.metadata.lastModified?.toUTCString())
     if (data.metadata.contentRange) {
       response.header('Content-Range', data.metadata.contentRange)
     }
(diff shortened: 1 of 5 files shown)


ikhsan:
Test is failing due to missing `PGRST_JWT_SECRET` I think

Although, in the PR I don’t touch any env variables so I’m a bit lost here on how to debug this. Fyi, it passes on my local machine.

inian:
Hi @ikhsan, not sure if this completely fixes https://github.com/supabase/storage-api/issues/193 but agree that this is a bug and the fix seems to send the header in the correct format. Thank you!

Regarding the tests, the container needs to be restarted as a quick way to reset the internal schema to the right contents - probably there is a way to reset the storage schema to the initial state without doing a full container restart. 

Tests seem to be passing locally. I will fix this in a separate PR.

inian (approved):

PR #195, opened 2022-09-26
feat: force downloading file when providing filename

fenos:

feature

we can only rely on HTML tags to download a file

## What is the new behavior?

Adds a new query string parameter `filename` which if provided the file will be forced to be downloaded by the browser
by adding the content-disposition header

## Additional context

Closes https://github.com/supabase/storage-api/issues/122

### DIFF
--- a/src/routes/object/getObject.ts
+++ b/src/routes/object/getObject.ts
@@ -27,8 +27,17 @@ const getObjectParamsSchema = {
   },
   required: ['bucketName', '*'],
 } as const
+
+const getObjectQuerySchema = {
+  type: 'object',
+  properties: {
+    download: { type: 'string', examples: ['filename.jpg', null] },
+  },
+} as const
+
 interface getObjectRequestInterface extends AuthenticatedRangeRequest {
   Params: FromSchema<typeof getObjectParamsSchema>
+  Querystring: FromSchema<typeof getObjectQuerySchema>
 }
 
 async function requestHandler(
@@ -42,6 +51,7 @@ async function requestHandler(
   >
 ) {
   const { bucketName } = request.params
+  const { download } = request.query
   const objectName = request.params['*']
 
   if (!isValidKey(objectName) || !isValidKey(bucketName)) {
@@ -86,6 +96,20 @@ async function requestHandler(
     if (data.metadata.contentRange) {
       response.header('Content-Range', data.metadata.contentRange)
     }
+
+    if (typeof download !== 'undefined') {
+      if (download === '') {
+        response.header('Content-Disposition', 'attachment;')
+      } else {
+        const encodedFileName = encodeURIComponent(download)
+
+        response.header(
+          'Content-Disposition',
+          `attachment; filename=${encodedFileName}; filename*=UTF-8''${encodedFileName};`
+        )
+      }
+    }
+
     return response.send(data.body)
   } catch (err: any) {
     if (err.$metadata?.httpStatusCode === 304) {
--- a/src/routes/object/getPublicObject.ts
+++ b/src/routes/object/getPublicObject.ts
@@ -25,11 +25,20 @@ const getPublicObjectParamsSchema = {
   },
   required: ['bucketName', '*'],
 } as const
+
+const getObjectQuerySchema = {
+  type: 'object',
+  properties: {
+    download: { type: 'string', examples: ['filename.jpg', null] },
+  },
+} as const
+
 interface getObjectRequestInterface {
   Params: FromSchema<typeof getPublicObjectParamsSchema>
   Headers: {
     range?: string
   }
+  Querystring: FromSchema<typeof getObjectQuerySchema>
 }
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
@@ -49,6 +58,7 @@ export default async function routes(fastify: FastifyInstance) {
     async (request, response) => {
       const { bucketName } = request.params
       const objectName = request.params['*']
+      const { download } = request.query
 
       const { error, status } = await request.superUserPostgrest
         .from<Bucket>('buckets')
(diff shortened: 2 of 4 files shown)


src/routes/object/getObject.ts
```diff
@@ -27,8 +27,17 @@ const getObjectParamsSchema = {
   },
   required: ['bucketName', '*'],
 } as const
+
+const getObjectQuerySchema = {
+  type: 'object',
+  properties: {
+    filename: { type: 'string', example: 'filename.json' },
```
inian: What about supporting this via `response-content-disposition` and users can do `object/auth/cat.png?response-content-disposition=attachment`. It is more obvious than the `filename` parameter since `response-content-disposition` makes it explicit that it is to trigger the file as a download. It also matches the [S3 format](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax). 

If users want to specify a different filename when downloading, which [turns out to be an issue](https://engineering.resolvergroup.com/2022/02/aws-s3-utf-8-content-disposition/#The%20Problem%20with%20UTF-8%20in%20the%20%E2%80%98response-content-disposition%E2%80%99%20Parameter) when we support unicode characters in the filename, they can also specify the filename using the same parameter like `object/auth/cat.png?response-content-disposition=attachment; filename=cat.png`

We will also need to add this parameter to our client libs

inian (changes requested):

src/routes/object/getObject.ts
fenos: this approach makes sense, my initial idea was based on the fact that `Content-Disposition` could only be:

- attachment
- inline (default)

however, as you are suggesting the file name could be Unicode based, so probably we could support this by default on the backend.

For example, an alternative more dev-friendly approach could be:

adding a `?download=1` query string parameter with a value of `1` would force the file to be downloaded with the original name.

Where else if the value is specified with any other strings, ex: `?download=file.jpg` or `?download=고양이.jpg` we would download the file with the given name.

I believe your proposed approach gives more freedom to clients to customize the header, whereas the second approach takes a more opinionated approach on how the header should be behaving simplifying the use.

What do you think?

inian: Ah that's interesting. Agree that we should prioritize DX than maintaining compatibility (the method I proposed is not completely compatible with S3's XML based API anyway). 

What about this? `?download` forces a download and users can optionally specify a different file name via `?download=newfilename`

fenos: I like that!

New implementation changed, we can now download the file with:

- `?download` results in the original file name
- `?download=name.jpg` specify a name

it also supports UTF-8 chars, ex: `?download=고양이.jpg`

fenos: Cross linking PR for `storage-js` https://github.com/supabase/storage-js/pull/112

JimH44:
Thanks, developers, for this wonderful site. I'm wondering if the feature described here has broken recently?
I tried this query string in my browser:
https://github.com/silnrsi/unicode-resources/tree/main/ucd-spreadsheet/ucd.xlsx?filename=ucd.xlsx
to download the file "ucd.xlsx" from https://github.com/silnrsi/unicode-resources/tree/main/ucd-spreadsheet
but I got the message:
(Sorry about that, but we can’t show files that are this big right now.)
When I try
https://github.com/silnrsi/unicode-resources/tree/main/ucd-spreadsheet?filename=ucd.xlsx
I see the files in that folder. If I click on ucd.xlsx I see the message about showing files.
I can right-click the file and Save As, but I have a script that notices when this file is updated, and I would like it to be able to download the updated file itself, and not just ask me to download it :-)
Thanks for thinking about this, Jim

JimH44:
Or perhaps I didn't understand how to use the query string parameter "filename" properly.

JimH44:
@inian and @fenos , I like the look of what you did to let us create a link that would download a file from github rather than showing it. I'm wondering if that feature has broken recently. Would you have time to look at my comment to this issue, at https://github.com/supabase/storage/pull/195#issuecomment-2089397033
Thanks for thinking about this, Jim

khera:
> Or perhaps I didn't understand how to use the query string parameter "filename" properly.

Anyone coming here in the future: the correct parameter to add is `download=FILENAME` where FILENAME can be blank or a name you want it to download. It is not called `filename` now.

JimH44:
Thanks @khera , I added _?download=ucd.csv_ to the link in the README file for this project, but still got the message 
`[View raw](https://github.com/silnrsi/unicode-resources/raw/refs/heads/main/ucd-spreadsheet/ucd.csv)
(Sorry about that, but we can’t show files that are this big right now.)`
The URL that I tried is
`https://github.com/silnrsi/unicode-resources/blob/main/ucd-spreadsheet/ucd.csv?download=ucd.csv`
What should I have used to get the file to download without any attempt at previewing it?
(I'm using Chrome on macos.)
@inian and @fenos , would you have time to look at this also, and show me the link to suggest in the README file,  to download the file instead of trying to preview it?
Thanks

roblatour:
I added ?download=filename.ext to my filename and its just opening in the browser on the github page that shows the raw format.

Here is the line in my markdown file (the file is pointed to via a relative link)
[Aqua](RulerDefinition_Aqua.ar4w?download=RulerDefinition_Aqua.ar4w) 

Any suggestions on how this should be recoded in my markdown file so that the file downloads?

khera:
@JimH44 and @roblatour this issue is about Supabase storage URLs, not GitHub wiki files or comment URLs.

roblatour:
sorry - my bad @khera. I had just googled for this problem and found this thread which I clearly jumped onto too quickly.  I should not have.  my apologies.

cies:
Am I right that this does not work for the S3 compatibility layer of Supabase Storage. With S3's presigned url you need to make the query param part of the signature. Also S3 uses the "response-content-disposition" query param to set the header (where Supabase Storage uses "download").

PR #197, opened 2022-09-30
feat: return object id on object operations

fenos:

feature

currently is not possible to create relationships from the client side to an object because the id is never returned

## What is the new behavior?

object id is returned on the following operations:

- `createObject`
- `copyObject`
- `updateObject`

### DIFF
--- a/src/routes/object/copyObject.ts
+++ b/src/routes/object/copyObject.ts
@@ -29,9 +29,13 @@ const copyRequestBodySchema = {
 const successResponseSchema = {
   type: 'object',
   properties: {
+    Id: {
+      type: 'string',
+      examples: ['2eb16359-ecd4-4070-8eb1-8408baa42493'],
+    },
     Key: { type: 'string', examples: ['folder/destination.png'] },
   },
-  required: ['Key'],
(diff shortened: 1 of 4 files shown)


src/routes/object/copyObject.ts
```diff
@@ -29,9 +29,13 @@ const copyRequestBodySchema = {
...
   type: 'object',
   properties: {
+    Id: {
+      type: 'string',
+      examples: ['2eb16359-ecd4-4070-8eb1-8408baa42493'],
+    },
     Key: { type: 'string', examples: ['folder/destination.png'] },
   },
-  required: ['Key'],
+  required: ['Key', 'Id'],
```
inian, line 38: Let's change it to `id`. `Key` is the only place we use capital case - we should change that to `key` in the next breaking release

inian (changes requested):

groud:
Hello!

We are using supabase on our project and we would need the ID returned somehow, or at least a way to get it using an API endpoint. I've just found this PR would likely solve our issue and I was wondering if it was planned to merge it (once rebased), or if another solution is planned to be provided soon ?

PR #199, opened 2022-10-03
feat: image resizing with imgproxy backend

fenos:

Feature

Currently is not possible to do image transformations on the fly

## What is the new behavior?

With this PR `storage-api` will be able to offload image processing to imgproxy and return the transformed image.

2 New endpoints have been introduced:

- `/storage/v1/render/public/<bucket>/<key>`
- `/storage/v1/render/authenticated/<bucket>/<key>`

Currently supported processing options:

- `width`
- `height`
- `resize`

### DIFF
--- a/src/renderer/imgproxy.ts
+++ b/src/renderer/imgproxy.ts
@@ -1,0 +1,75 @@
+import { GenericStorageBackend } from '../backend/generic'
+import axios, { Axios } from 'axios'
+import { getConfig } from '../utils/config'
+
+interface TransformOptions {
+  width?: number
+  height?: number
+  resize?: 'fill' | 'fit' | 'fill-down' | 'force' | 'auto'
+}
+
+interface ImgProxyOptions {
+  url: string
+}
+
+const { imgLimits } = getConfig()
+
+const LIMITS = {
+  height: {
+    min: imgLimits.size.min,
+    max: imgLimits.size.max,
+  },
+  width: {
+    min: imgLimits.size.min,
+    max: imgLimits.size.max,
+  },
+}
+
+export class Imgproxy {
+  private client: Axios
+
+  constructor(
+    private readonly backend: GenericStorageBackend,
+    private readonly options: ImgProxyOptions
+  ) {
+    this.client = axios.create({
+      baseURL: options.url,
+      timeout: 8000,
+    })
+  }
+
+  getClient() {
+    return this.client
+  }
+
+  async transform(bucket: string, key: string, options: TransformOptions) {
+    const privateURL = await this.backend.privateAssetUrl(bucket, key)
+    const urlTransformation = this.applyURLTransformation(options)
+
+    const url = ['/public', ...urlTransformation, 'plain', privateURL]
+
+    return this.getClient().get(url.join('/'), {
+      responseType: 'stream',
+    })
+  }
+
+  protected applyURLTransformation(options: TransformOptions) {
+    const segments = []
+
+    if (options.height) {
+      segments.push(`height:${clamp(options.height, LIMITS.height.min, LIMITS.height.max)}`)
+    }
+
+    if (options.width) {
+      segments.push(`width:${clamp(options.width, LIMITS.width.min, LIMITS.width.max)}`)
+    }
+
+    if (options.width || options.height) {
+      segments.push(`resizing_type:${options.resize || 'fill'}`)
+    }
+
+    return segments
+  }
+}
+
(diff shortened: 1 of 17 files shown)


src/renderer/imgproxy.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+interface ImgProxyOptions {
+  url: string
+}
+
+const LIMITS = {
+  height: {
+    max: 5000,
+    min: 20,
```
inian: min limits can probably be 1?

src/renderer/imgproxy.ts
```diff
@@ -0,0 +1,72 @@
+import { GenericStorageBackend } from '../backend/generic'
+import axios, { Axios } from 'axios'
+
+interface TransformOptions {
+  width?: number
+  height?: number
+  resize?: 'fill' | 'fit' | 'fill-down' | 'force' | 'auto'
```
inian, line 8: @alaister you had an idea here on how to map these to css properties right?

src/renderer/imgproxy.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+interface ImgProxyOptions {
+  url: string
+}
+
+const LIMITS = {
+  height: {
+    max: 5000,
```
inian: Lets make these configurable anyway, we might want to offer different limits to different users

src/routes/render/renderAuthenticatedImage.ts
```diff
@@ -0,0 +1,111 @@
...
+      if (objectResponse.error) {
+        const { status, error } = objectResponse
+        request.log.error({ error }, 'error object')
+        return response.status(400).send(transformPostgrestError(error, status))
+      }
+
+      const s3Key = `${request.tenantId}/${bucketName}/${objectName}`
+
+      try {
+        const imageResponse = await imageRenderer.transform(globalS3Bucket, s3Key, request.query)
```
inian, line 83: is there a way to stream the response as a stream [here](https://github.com/supabase/storage-api/pull/199/files#diff-db92043d530d0a7fd1b0b47bf588452fef1c11d5eba52614e9768f9d7d556e59R49) and send that stream directly to the client instead of doing an await?

src/routes/render/renderAuthenticatedImage.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        response
+          .status(imageResponse.status)
+          .header('Accept-Ranges', 'bytes')
+          .header('Content-Type', normalizeContentType(imageResponse.headers['content-type']))
+          .header('Cache-Control', imageResponse.headers['cache-control'])
+          .header('Content-Length', imageResponse.headers['content-length'])
+          .header('ETag', imageResponse.headers['etag'])
```
inian, line 91: Send the last-modified header too like [here](https://github.com/supabase/storage-api/blob/a14d627002c0fbc4bb408757a01f68a4880f0eb3/src/routes/object/getObject.ts#L95)?

src/routes/render/renderPublicImage.ts
```diff
@@ -0,0 +1,108 @@
...
+      try {
+        const imageResponse = await imageRenderer.transform(globalS3Bucket, s3Key, request.query)
+
+        response
+          .status(imageResponse.status)
+          .header('Accept-Ranges', 'bytes')
+          .header('Content-Type', normalizeContentType(imageResponse.headers['content-type']))
+          .header('Cache-Control', imageResponse.headers['cache-control'])
+          .header('Content-Length', imageResponse.headers['content-length'])
+          .header('ETag', imageResponse.headers['etag'])
```
inian, line 88: Last modified header

inian (approved):

src/renderer/imgproxy.ts
fenos: ok, for the MVP we'll have these values as envs

src/routes/render/renderAuthenticatedImage.ts
fenos: yes good catch

src/routes/render/renderAuthenticatedImage.ts
fenos: we don't have this info as we don't store it currently. 
It could be a good thing to store the value in the metadata in another PR

src/routes/render/renderPublicImage.ts
fenos: we don't have this info as we don't store it currently. 
It could be a good thing to store the value in the metadata in another PR

PR #204, opened 2022-10-12
Image resizing improvements

fenos:

Fix & Improvements

## What is the new behavior?

- fix: S3 signed URL needs to be encoded before passing it to imgproxy
- reduce signed URL time to 5mins
- add X-Transformation header with transformation options

### DIFF
--- a/src/routes/render/renderAuthenticatedImage.ts
+++ b/src/routes/render/renderAuthenticatedImage.ts
@@ -80,7 +80,11 @@ export default async function routes(fastify: FastifyInstance) {
       const s3Key = `${request.tenantId}/${bucketName}/${objectName}`
 
       try {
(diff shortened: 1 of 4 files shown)


src/routes/render/renderAuthenticatedImage.ts
```diff
@@ -80,7 +80,12 @@ export default async function routes(fastify: FastifyInstance) {
       const s3Key = `${request.tenantId}/${bucketName}/${objectName}`
 
       try {
-        const imageResponse = await imageRenderer.transform(globalS3Bucket, s3Key, request.query)
+        const { image, urlTransformation } = await imageRenderer.transform(
+          globalS3Bucket,
+          s3Key,
+          request.query
+        )
+        const imageResponse = await image
```
inian: probably cleaner to do this await inside the transform function itself?

inian (approved):

PR #205, opened 2022-10-12
Custom HEAD request for object info

fenos:

feature

Current HEAD implementation would download the file from S3

## What is the new behavior?

A custom HEAD route  that will return the file headers and check RLS upon request


src/routes/object/getObjectInfo.ts
```diff
@@ -0,0 +1,147 @@
...
+    })
+    .single()
+
+  if (objectResponse.error) {
+    const { status, error } = objectResponse
+    request.log.error({ error }, 'error object')
+    return response.status(400).send(transformPostgrestError(error, status))
+  }
+
+  // send the object from s3
```
inian: nit: change comment

src/routes/object/getObjectInfo.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  try {
+    const data = await storageBackend.headObject(globalS3Bucket, s3Key)
+
+    response
+      .status(data.httpStatusCode ?? 200)
+      .header('Content-Type', normalizeContentType(data.mimetype))
+      .header('Cache-Control', data.cacheControl)
+      .header('Content-Length', data.size)
+      .header('ETag', data.eTag?.replace(new RegExp('"', 'g'), ''))
```
inian: why is this regex required? in [other places](https://github.com/supabase/storage-api/blob/165197e03f48758e92728ca497a95b57f69ad522/src/routes/object/getPublicObject.ts#L89) looks like we just pass it through

inian (approved):
Nice, with this PR the HEAD route will show up in our docs also hopefully. Previously it didn't cos fastify didn't export the swagger for the auto generated HEAD route

src/routes/object/getObjectInfo.ts
fenos: for some reason, the Etag that is coming from S3 is quoted "".
It felt a bit weird, but happy to return it as is if that's accetable

PR #215, opened 2022-11-03
Feat/webhooks

fenos:

Feature

None

## What is the new behavior?

The storage API will now be able to send webhooks to a specific WEBHOOK_URL using a queue-based mechanism.
We have a retry mechanism and exponential back-off in place, thanks to pg-boss

the events we currently send are:

- ObjectCreated:Post
- ObjectCreated:Put
- ObjectCreated:Move
- ObjectRemoved:Delete
- ObjectRemoved:Move
- ObjectUpdated:Metadata

## Additional context

A step forward for automatic cache invalidation


src/queue/events/base-event.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  static async sendWebhook<T extends BaseEvent<any>>(this: StaticThis<T>, payload: T['payload']) {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const { Webhook } = require('./webhook')
+
+    await Webhook.send({
+      eventName: this.constructor.name,
+      payload: payload,
+      project: payload.project,
```
inian: lets keep it consistent and use `tenantId` here?

src/storage/object.ts
```diff
@@ -295,14 +304,34 @@ export class ObjectStorage {
...
     await this.backend.copyObject(globalS3Bucket, s3SourceKey, s3DestinationKey)
     await this.backend.deleteObject(globalS3Bucket, s3SourceKey)
+
+    const movedObject = await this.db.findObject(
+      this.bucketId,
+      destinationObjectName,
+      'bucket_id, metadata'
+    )
+
+    await Promise.all([
```
inian, line 331: I dont think we should send delete and create events for move. This is more of an implementation detail cos s3 doesnt support move api directly

inian: [s3 object events](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html) also have different events for copy, upsert, create etc

src/storage/object.ts
```diff
@@ -93,6 +188,16 @@ export class ObjectStorage {
       if (data.length > 0) {
         results = results.concat(data)
 
+        await Promise.all(
```
inian: change to Promise.allSettled instead since one webhook call stop the other calls?

inian: What is the expected behaviour when the webhook sending fails but the object is successfully deleted?

src/storage/object.ts
```diff
@@ -18,6 +26,86 @@ export class ObjectStorage {
...
+      const objectMetadata = await uploader.upload(request, {
+        key: s3Key,
+      })
+
+      await this.db
+        .asSuperUser()
+        .updateObjectMetadata(this.bucketId, options.objectName, objectMetadata)
+
+      await ObjectCreated.sendWebhook({
+        project: this.db.project(),
```
inian: probably a good idea to add a version field to these webhook payloads

inian: or as a header https://webhooks.fyi/ops-experience/versioning

src/queue/events/base-event.ts
fenos: project it is an object:

```
 {
   host: '{project_id}.supabase.co',
   ref: '{project_id}',
}
```

I can rename it to `tenant`

src/storage/object.ts
fenos: Interesting point,
Yes it looks like they have different names, but they are namespaced with `ObjectCreated::{Operation}`
We could do something like that too

Are you thinking to have a specific event for Move too?

inian (approved):

PR #218, opened 2022-11-14
Feat/use environment variables with default values

### LINKED ISSUE #217: Make the port and host environment variables
# Feature request
There's currently issue with services such as Fly.io as they try to connect with an IPv6 address and there's currently no way to bind
to `::` as the address `0.0.0.0` is hard coded

## Is your feature request related to a problem? Please describe.
When hosted on fly.io, fastify refuses connections

## Describe the solution you'd like
Introduce HOST and PORT as environment variable with default values 0.0.0.0 and 5000 respectively

## Describe alternatives you've considered
A custom docker image with this project built from source on a modified branch. Not done at the time of writing this

sduduzog:

feature

Server binds to hard coded address 0.0.0.0:5000

Closes #217

## What is the new behavior?

Uses the variables HOST and PORT for the address to bind to when provided. If not set, defaults to 0.0.0.0:5000

See below, logs from the modified storage-api on fly.io
image

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -10,6 +10,9 @@ import { logger } from './monitoring'
 
 const exposeDocs = true
 
+const port = process.env.PORT || 5000;
+const host = process.env.HOST || '0.0.0.0'
+
 ;(async () => {
   const { isMultitenant, requestIdHeader, adminRequestIdHeader } = getConfig()
   if (isMultitenant) {
@@ -23,7 +26,7 @@ const exposeDocs = true
     })
 
     try {
-      await adminApp.listen({ port: 5001, host: '0.0.0.0' })
+      await adminApp.listen({ port, host })
     } catch (err) {
       adminApp.log.error(err)
       process.exit(1)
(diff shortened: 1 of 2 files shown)


src/server.ts
```diff
@@ -23,7 +23,7 @@ const exposeDocs = true
     })
 
     try {
-      await adminApp.listen({ port: 5001, host: '0.0.0.0' })
+      await adminApp.listen({ port, host })
```
fenos: The port for the admin app should be different from the app port,
we can have a different env for this ex: `ADMIN_PORT`

sduduzog: Very true. Good catch actually. I should add something like ADMIN_PORT with the default as PORT+1

fenos: you can simply default it to `5001` for backward compatibility

```
Number(getOptionalConfigFromEnv('ADMIN_PORT')) || 5001
```

fenos:
@sduduzog looks great! 
There are 2 lint issues would you be able to run prettier on these files pls, will merge right after

fenos (approved):

fenos:
Actually no worries, I'll fix the lints
Thanks a lot for the contribution

PR #233, opened 2022-12-02
chore: add ghcr image mirror

sweatybridge:

ci

## What is the new behavior?

publish images to ghcr for faster CLI start time on GHA

## Additional context

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -1,0 +1,33 @@
+name: Mirror Image
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Image tag"
+        required: true
+        type: string
+
+jobs:
+  mirror:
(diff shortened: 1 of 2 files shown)


.github/workflows/mirror.yml
```diff
@@ -0,0 +1,33 @@
+name: Mirror Image
```
inian, line 1: do we need a separate action for this?

inian (approved):

laktek (approved):

.github/workflows/mirror.yml
sweatybridge: this is helpful for doing a one time mirror for older images, for eg. the current version used by cli `v0.21.4`. Using this job, I can mirror it over from docker hub to ghcr without rebuilding.

PR #237, opened 2022-12-06
fix(migrate): change the connectionTimeoutMillis to 10000 from 10_000

key7men:

Bug fix

Database connect timeout is not a valid number


fenos:
Hello @key7men thanks for the contribution, in JS using an underscore makes it a valid number.
It is the same as 1000.

```
10_000 === 10000 // true
```

key7men:
> 
So it can also be like this, learn a new point of knowledge. Thank U.

PR #239, opened 2022-12-07
Feat/rate limiting

fenos:

feature

## What is the new behavior?

Implementing in memory/redis based rate limiting, to the render routes

environment variables added:

```
ENABLE_RATE_LIMITER=true

RATE_LIMITER_DRIVER=redis # or memory
RATE_LIMITER_REDIS_URL=6379

RATE_LIMITER_RENDER_PATH_MAX_REQ_SEC=4 #max number of requests per second before being limited 
```

to enable the limiter set to `true` the `ENABLE_RATE_LIMITER` env


src/http/routes/render/rate-limiter.ts
```diff
@@ -0,0 +1,40 @@
...
+export const rateLimiter = fp((fastify: FastifyInstance, ops: any, done: () => void) => {
+  fastify.register(fastifyRateLimit, {
+    global: true,
+    max: rateLimiterRenderPathMaxReqSec * 4,
+    timeWindow: 4 * 1000, //4s
+    continueExceeding: true,
+    nameSpace: 'image-transformation-ratelimit-',
+    redis: rateLimiterDriver === 'redis' ? new Redis(rateLimiterRedisUrl || '') : undefined,
+    keyGenerator: function (request) {
+      const ip = request.headers['x-real-ip'] || request.headers['x-client-ip'] || request.ip
```
inian, line 19: should the rate limiter be per (tenant, ip) instead of just ip if its running in multi tenant mode? that way we can have different rate limits per project (tier)

fenos: Make sense!

PR #240, opened 2022-12-08
Feature/aliyun oss

key7men:

New Feature

Unsupport Aliyun(the world's third largest cloud service provider) OSS (Object Storage Service)

## What is the new behavior?

When users use self-hosting, they have an additional option for object storage service.


.env.sample
```diff
@@ -3,6 +3,10 @@ SERVICE_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIjoic2VydmljZV9yb2xlIiw
 TENANT_ID=bjhaohmqunupljrqypxz
 REGION=region-of-your-s3-bucket
 GLOBAL_S3_BUCKET=name-of-your-s3-bucket
+GLOBAL_OSS_BUCKET=
```
fenos: You don't need this variable, we always use `GLOBAL_S3_BUCKET` maybe we can rename it to `GLOBAL_BUCKET` to make it more generic in a future PR

src/database/migrate.ts
```diff
@@ -33,7 +33,7 @@ export async function runMigrationsOnTenant(databaseUrl: string): Promise<void>
 async function connectAndMigrate(databaseUrl: string | undefined, migrationsDirectory: string) {
   const dbConfig = {
     connectionString: databaseUrl,
-    connectionTimeoutMillis: 10_000,
+    connectionTimeoutMillis: 10000,
```
fenos, line 36: no need to change this.
In js using _ makes it a valid number, just easier to read

src/storage/backend/index.ts
```diff
@@ -1,21 +1,31 @@
...
+  globalOSSEndpoint,
+  ossAccessKey,
+  ossAccessSecret,
+  storageBackendType,
+} = getConfig()
 
 export function createStorageBackend() {
   let storageBackend: StorageBackendAdapter
 
   if (storageBackendType === 'file') {
```
fenos, line 23: probably is time to change this if statement with a switch statement over `storageBackendType` and throw an error if the adapter name is not supported.

Defaulting to OSS is a breaking change

PR #243, opened 2022-12-09
chore: removes trailing space from migration file

sweatybridge:

style

identified by a CLI user https://github.com/supabase/cli/pull/648

## What is the new behavior?

removes trailing space

## Additional context


inian:
Since the migration tool for tenants stores the hash of the migration, modifying old migration files is not a good idea

PR #244, opened 2022-12-09
feat(object): Allow for presigning upload url

### LINKED ISSUE #81: Signed urls for upload
# Feature request

## Is your feature request related to a problem? Please describe.

At [Labelflow](https://github.com/Labelflow/labelflow), we developed a tool to upload images on our Supabase storage, based on one nextJs API route. The goal is for us to abstract the storage method from the client-side by querying a generic upload route to upload any file and to ease the permission management. Indeed, in the server-side function, one service role Supabase client is manipulated to actually make the upload. We use next-auth to secure the route (and to manage authentication in the app in general).

Client-side upload looks like that:
```javascript
await fetch("https://labelflow.ai/api/upload/[key-in-supabase]", {
                  method: "PUT",
                  body: file,
                });
```

Server-side API route looks more or less like that (I don't show the permission management part):
```javascript
import { createClient } from "@supabase/supabase-js";
import nextConnect from "next-connect";

const apiRoute = nextConnect({});
const client = createClient(
  process?.env?.SUPABASE_API_URL as string,
  process?.env?.SUPABASE_API_KEY as string
);
const bucket = "labelflow-images";

apiRoute.put(async (req, res) => {
  const key = (req.query.id as string[]).join("/");
  const { file } = req;
  const { error } = await client.storage.from(bucket).upload(key, file.buffer, {
    contentType: file.mimetype,
    upsert: false,
    cacheControl: "public, max-age=31536000, immutable",
  });
  if (error) return res .status(404);
  return res.status(200);
});

export default apiRoute;
```

**The problem is that we face a serious limitation in terms of upload size** since we use Vercel for deployment which [doesn't allow serverless functions to handle requests that are more than 5Mb](https://vercel.com/support/articles/how-to-bypass-vercel-5mb-body-size-limit-serverless-functions). Since we send over images in the upload request from the client to the server, we're likely to reach that limit quite often.

## Describe the solution you'd like

As we don't want to manipulate Supabase clients client-side, we think that the ideal solution would be to allow us to upload directly to Supabase, using an **upload signed URL**.  The above upload route could now take only a key as an input and return a signed URL to make the upload to.

Client-side upload would now be in two steps:
```javascript
// Get Supabase signed Url
const { signedUrl } = await (await fetch("https://labelflow.ai/api/upload/[key-in-supabase]", {
                  method: "GET",
                })).json();

// Upload the file
await fetch(signedUrl, {
                  method: "PUT",
                  body: file,
                });
```

And our API route would look like that, more or less:
```javascript
import { createClient } from "@supabase/supabase-js";
import nextConnect from "next-connect";

const apiRoute = nextConnect({});
const client = createClient(
  process?.env?.SUPABASE_API_URL as string,
  process?.env?.SUPABASE_API_KEY as string
);
const bucket = "labelflow-images";

apiRoute.get(async (req, res) => {
  const key = (req.query.id as string[]).join("/");
  const { signedURL } = await client.storage
    .from(bucket)
    .createUploadSignedUrl(key, 3600); // <= this is the missing feature

  if (signedURL) {
    res.setHeader("Content-Type", "application/json");
    return res.status(200).json({signedURL});
  }

  return res.status(404);
});

export default apiRoute;
```

## Describe alternatives you've considered

I described them in our [related issue](https://github.com/labelflow/labelflow/issues/576#issuecomment-972732048):
- Use the Supabase client, client-side. We'd need to give extra care about security, and the fact we don’t use Supabase auth is not helping. Some references to do that: https://supabase.io/docs/guides/auth#row-level-security & https://github.com/supabase/supabase/tree/master/examples/nextjs-ts-user-management#postgres-row-level-security
- Use another storage method (Google Cloud, AWS) that supports upload SignedUrl, in an end-to-end fashion instead of Supabase 
- Use another storage method (Google Cloud, AWS) that supports upload SignedUrl as an intermediate. E.g. upload on a GC bucket then download it from the API serverless function and finally upload on Supabase using the client

## Additional context

We're happy to work on developing this feature at Labelflow if you think this is the best option!

MagnusHJensen:

Introduces a new **feature** to sign an upload url, which allows for uploading straight to the bucket without credentials.

It's currently not possible to upload a file from the client side, without routing it through your own server, which is not the preferred way.

## What is the new behavior?

You can now presign a upload url with your service token, and give the url to any client which can then upload the file directly.

## Additional context
Closes #81 

_First PR for the Storage API_

### DIFF
--- a/src/storage/object.ts
+++ b/src/storage/object.ts
@@ -14,6 +14,7 @@ import {
   ObjectRemovedMove,
   ObjectUpdatedMetadata,
 } from '../queue'
+import { StorageBackendError } from './errors'
 
 export interface UploadObjectOptions {
   objectName: string
@@ -446,5 +447,40 @@ export class ObjectStorage {
         }
       })
     )
+  }
+
+  /**
+   * Generates a signed url for uploading an object
+   * @param objectName
+   * @param url
+   * @param expiresIn
+   * @param metadata
+   * @returns
+   */
+  async signUploadObjectUrl(
+    objectName: string,
+    url: string,
+    expiresIn: number,
+    metadata?: Record<string, string>
+  ) {
+    await this.db.findBucketById(this.bucketId) // Require bucket to be present
+
+    let found
+    try {
+      found = await this.findObject(objectName)
+    } catch (e) {
+      // Don't do anything, since it will throw an error if the object doesn't exist (but that is what we want.)
+    }
+
+    if (found) {
+      throw new StorageBackendError('Duplicate', 409, 'The resource already exists')
+    }
+
+    const urlParts = url.split('/')
+    const urlToSign = decodeURI(urlParts.splice(4).join('/'))
+    const jwtSecret = await getJwtSecret(this.db.tenantId)
+    const token = await signJWT({ url: urlToSign, ...metadata }, jwtSecret, expiresIn)
+
+    return `/object/upload/sign/${urlToSign}?token=${token}`
   }
 }
--- a/src/http/routes/object/getSignedUploadURL.ts
+++ b/src/http/routes/object/getSignedUploadURL.ts
@@ -1,0 +1,68 @@
+import { FastifyInstance } from 'fastify'
+import { FromSchema } from 'json-schema-to-ts'
+import { createDefaultSchema } from '../../generic-routes'
+import { AuthenticatedRequest } from '../../request'
+
+const getSignedUploadURLParamsSchema = {
+  type: 'object',
+  properties: {
+    bucketName: { type: 'string', examples: ['avatars'] },
+    '*': { type: 'string', examples: ['folder/cat.png'] },
+  },
+  required: ['bucketName', '*'],
+} as const
+const getSignedUploadURLBodySchema = {
+  type: 'object',
+  properties: {
+    expiresIn: { type: 'integer', minimum: 1, examples: [60000] },
+  },
+  required: ['expiresIn'],
+} as const
+
+const successResponseSchema = {
+  type: 'object',
+  properties: {
+    signedUploadURL: {
+      type: 'string',
+      examples: [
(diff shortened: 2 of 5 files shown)


fenos:
@MagnusHJensen great work with this PR!
This feature is indeed very useful.

fenos:
This week i'll spend sometime to check it out 👍

MagnusHJensen:
Sounds great @fenos 
Let me know of any findings or questions.

fenos:
Hello @MagnusHJensen sorry for the delay, I was away during the Christmas holiday.

I had time to look over the PR and generally looks good. 
However one important aspect that we need to get right is to understand who has the permission to create the upload URL.

src/storage/object.ts
```diff
@@ -447,4 +448,39 @@ export class ObjectStorage {
...
+    expiresIn: number,
+    metadata?: Record<string, string>
+  ) {
+    await this.db.findBucketById(this.bucketId) // Require bucket to be present
+
+    let found
+    try {
+      found = await this.findObject(objectName)
+    } catch (e) {
+      // Don't do anything, since it will throw an error if the object doesn't exist (but that is what we want.)
```
fenos, line 478: This is not going to work reliably,
If we have certain RLS policies defined this might return 404 because we don't have access to the file not because it doesn't exist

We might want to run this query `asSuperUser` to make sure the file is not present at this location.

However, when doing this as superUser we don't have any protection on who can create these upload URL.

As a second step, after the file existence, we need to check if the user has the CREATE RLS permission in order to generate the upload url

MagnusHJensen:
Hi @fenos 
Feel free to take over this PR if you have the time and find the feature needed enough.
I don't think I will have the time in the near future, for finishing this. _If it still stands once I do have time, I will finish it up_

subvertallchris:
Sorry to be annoying with a +1 that doesn't contribute anything but... I hope this makes it in, it would be a fantastic feature.

fenos:
Closing in favour of #81

PR #246, opened 2022-12-19
fix: unsupported extended file attributes on macOS

burmecia:

This PR is to fix `The file system does not support extended attributes or has the feature disabled` error on macOS.

See:

https://github.com/supabase/supabase/issues/10745
https://github.com/supabase/supabase/discussions/9539

When trying to upload a file/image using to a locally hosted supabase instance, get the error:

`The file system does not support extended attributes or has the feature disabled`

## What is the new behavior?

There should be no error.

## Additional context

After this fix, user also need to enable `VirtioFS` in Docker desktop to be able to set file attribute.

image

### DIFF
--- a/src/storage/backend/file.ts
+++ b/src/storage/backend/file.ts
@@ -14,6 +14,18 @@ interface FileMetadata {
   contentType: string
 }
 
+// file metadata attribute keys on different platforms
+const METADATA_ATTR_KEYS = {
+  'darwin': {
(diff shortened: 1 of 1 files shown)


inian (approved):

inian:
Can we document the VirtioFS requirement here @burmecia https://supabase.com/docs/guides/resources/self-hosting/docker

burmecia:
Sure, document is added in https://github.com/supabase/supabase/pull/11096.

PR #254, opened 2022-12-29
set forcePathStyle when using S3 endpoint

mingfang:
fixes https://github.com/supabase/storage-api/issues/252

enhancement

does not work with localstack or minio

## What is the new behavior?

now works with localstack and minio

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -10,6 +10,7 @@ type StorageConfigType = {
   fileStoragePath?: string
   globalS3Bucket: string
   globalS3Endpoint?: string
+  globalS3ForcePathStyle?: boolean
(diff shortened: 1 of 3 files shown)


fenos:
can we make this a configuration option using an env variable?

maybe something like: `S3_FORCE_PATH_STYLE` - let's try it to keep it backwards compatible

mingfang:
@fenos 
I added a new configuration called `GLOBAL_S3_FORCE_PATH_STYLE`

fenos (approved):

mingfang:
@fenos please review

fenos:
Thanks for the contribution @mingfang

PR #272, opened 2023-01-26
Migrations for customise buckets behaviour

fenos:

Feature

## What is the new behavior?

Add new migrations for the following features:

- avif auto detection toggle
- custom max file size per bucket
- custom mime type per bucket


migrations/tenant/0011-add-automatic-avif-detection-flag.sql
```diff
@@ -0,0 +1 @@
+alter table storage.buckets add column avif_autodetection bool default false
```
inian: nit: add semicolon and newline

inian (approved):

PR #277, opened 2023-02-08
feat: allow to customise max file size and mime types at bucket level

fenos:

Feature

## What is the new behavior?

- Allow creating buckets specifying a `file size limit` and white-listing `mime-types` which will serve for validating the file during upload

- Allow to update these values via the bucket update endpoint 

- Validate max file size and mime type during upload

### DIFF
--- a/src/http/routes/bucket/updateBucket.ts
+++ b/src/http/routes/bucket/updateBucket.ts
@@ -7,8 +7,8 @@ const updateBucketBodySchema = {
   type: 'object',
   properties: {
     public: { type: 'boolean', examples: [false] },
-    maxFileSizeKb: { type: 'integer', minimum: 1, examples: [1000] },
-    allowedMimeTypes: { type: 'array', items: { type: 'string' } },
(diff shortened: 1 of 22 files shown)


src/http/routes/bucket/updateBucket.ts
```diff
@@ -44,9 +46,13 @@ export default async function routes(fastify: FastifyInstance) {
     async (request, response) => {
       const { bucketId } = request.params
 
-      const { public: isPublic } = request.body
+      const { public: isPublic, maxFileSizeKb, allowedMimeTypes } = request.body
```
inian: I think we need to assume all of these properties are optional now. If someone calls /update with just `public`, the existing values will get nuked right?

inian (approved):

src/http/routes/bucket/updateBucket.ts
fenos: these are optional, if not provided they will not change the underline value

PR #282, opened 2023-02-26
feat(object): Allow for presigning upload url

abbit:

Introduces a new feature to sign an upload url, which allows for uploading straight to the bucket without credentials.

It's currently not possible to upload a file from the client side, without routing it through your own server, which is not the preferred way.

## What is the new behavior?

You can now presign a upload url with your service token, and give the url to any client which can then upload the file directly.

## Additional context

Closes https://github.com/supabase/storage-api/issues/81
This PR fixes things from https://github.com/supabase/storage-api/pull/244. Thanks to @MagnusHJensen for prior work!

### DIFF
--- a/src/http/routes/object/getSignedUploadURL.ts
+++ b/src/http/routes/object/getSignedUploadURL.ts
@@ -1,0 +1,68 @@
+import { FastifyInstance } from 'fastify'
+import { FromSchema } from 'json-schema-to-ts'
+import { createDefaultSchema } from '../../generic-routes'
+import { AuthenticatedRequest } from '../../request'
+
+const getSignedUploadURLParamsSchema = {
+  type: 'object',
+  properties: {
+    bucketName: { type: 'string', examples: ['avatars'] },
+    '*': { type: 'string', examples: ['folder/cat.png'] },
+  },
+  required: ['bucketName', '*'],
+} as const
+const getSignedUploadURLBodySchema = {
+  type: 'object',
+  properties: {
+    expiresIn: { type: 'integer', minimum: 1, examples: [60000] },
+  },
+  required: ['expiresIn'],
+} as const
+
+const successResponseSchema = {
+  type: 'object',
+  properties: {
+    signedUploadURL: {
+      type: 'string',
+      examples: [
+        '/object/sign/upload/avatars/folder/cat.png?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1cmwiOiJhdmF0YXJzL2ZvbGRlci9jYXQucG5nIiwiaWF0IjoxNjE3NzI2MjczLCJleHAiOjE2MTc3MjcyNzN9.s7Gt8ME80iREVxPhH01ZNv8oUn4XtaWsmiQ5csiUHn4',
+      ],
+    },
+  },
+  required: ['signedUploadURL'],
+}
+interface getSignedURLRequestInterface extends AuthenticatedRequest {
+  Params: FromSchema<typeof getSignedUploadURLParamsSchema>
+  Body: FromSchema<typeof getSignedUploadURLBodySchema>
+}
+
+export default async function routes(fastify: FastifyInstance) {
+  const summary = 'Generate a presigned url to upload an object'
+
+  const schema = createDefaultSchema(successResponseSchema, {
+    body: getSignedUploadURLBodySchema,
+    params: getSignedUploadURLParamsSchema,
+    summary,
+    tags: ['object'],
+  })
+
+  fastify.post<getSignedURLRequestInterface>(
+    '/upload/sign/:bucketName/*',
+    {
+      schema,
+    },
+    async (request, response) => {
+      const { bucketName } = request.params
+      const objectName = request.params['*']
+      const { expiresIn } = request.body
+
+      const urlPath = request.url.split('?').shift()
+
+      const signedUploadURL = await request.storage
+        .from(bucketName)
+        .signUploadObjectUrl(objectName, urlPath as string, expiresIn)
+
+      return response.status(200).send({ signedUploadURL })
+    }
+  )
+}
--- a/src/http/routes/object/getSignedUploadURL.ts
+++ b/src/http/routes/object/getSignedUploadURL.ts
@@ -22,14 +22,14 @@ const getSignedUploadURLBodySchema = {
 const successResponseSchema = {
   type: 'object',
   properties: {
-    signedUploadURL: {
+    url: {
       type: 'string',
       examples: [
         '/object/sign/upload/avatars/folder/cat.png?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1cmwiOiJhdmF0YXJzL2ZvbGRlci9jYXQucG5nIiwiaWF0IjoxNjE3NzI2MjczLCJleHAiOjE2MTc3MjcyNzN9.s7Gt8ME80iREVxPhH01ZNv8oUn4XtaWsmiQ5csiUHn4',
       ],
     },
   },
-  required: ['signedUploadURL'],
+  required: ['url'],
 }
 interface getSignedURLRequestInterface extends AuthenticatedRequest {
   Params: FromSchema<typeof getSignedUploadURLParamsSchema>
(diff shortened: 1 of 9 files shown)


src/storage/object.ts
```diff
@@ -465,4 +470,42 @@ export class ObjectStorage {
...
+   * @param objectName
+   * @param url
+   * @param expiresIn seconds
+   * @param owner
+   */
+  async signUploadObjectUrl(objectName: string, url: string, expiresIn: number, owner?: string) {
+    // check as super user if the object already exists
+    let found
+    try {
+      found = await this.findObject(objectName, 'id', true)
```
fenos: it would be better to do:

```
await this.asSuperUser().findObject(objectName, 'id')
```

instead of passing the the param in the function

src/storage/object.ts
```diff
@@ -465,4 +470,42 @@ export class ObjectStorage {
...
+    } catch (e) {
+      // Don't do anything, since it will throw an error if the object doesn't exist (but that is what we want.)
+    }
+
+    if (found) {
+      throw new StorageBackendError('Duplicate', 409, 'The resource already exists')
+    }
+
+    // check if user has INSERT permissions
+    await this.db.createObject({
```
fenos: We would need to create a stored function that does something along these lines:


```
// canInsertObject

BEGIN

INSERT INTO objects ....

ROLLBACK
```

I have not tested the above, but the idea is to simply invoke an rpc function and run the insert to check the permissions and then roll it back if it's fine, else it will throw row-level error

fenos:
@abbit this PR looks great!!

Very close to be merged in, left some comments

src/http/routes/object/getSignedUploadURL.ts
```diff
@@ -0,0 +1,69 @@
...
+      const { expiresIn } = request.body
+      const owner = request.owner
+
+      const urlPath = request.url.split('?').shift()
+
+      const signedUploadURL = await request.storage
+        .from(bucketName)
+        .signUploadObjectUrl(objectName, urlPath as string, expiresIn, owner)
+
+      return response.status(200).send({ signedUploadURL })
```
fenos: how about changing the key of the response to `url`?

Example:

```json
{
  "url": "signedurl"
}
```

abbit:
@fenos Thanks for the review! I will try to fix the things you mentioned in the next couple days.

src/storage/object.ts
abbit: Do you mean that I have to introduce a new function `asSuperUser()` for `ObjectStorage` or should I simply use `this.db.asSuperUser()`?

abbit:
@fenos, fixes are done. Used a hack with exception to rollback insertion since Postgres doesn't allow ROLLBACK in stored functions and PostgREST doesn't support stored procedures. Please review when you have time. Thanks.

fenos:
@abbit this PR looks awesome! well done and thanks for the contribution!
I gave it a second pass and looks fantastic! 

However, I realized that the upload URLs we are signing are not single-use.  
If the file gets deleted, the user who used the signed upload URL can now re-use it again. 

We don't store the token in the database so we don't have a way to invalidate it after use. 

Since we have this limitation, we cannot allow generating upload URL's which are too long, they could become a security risk.

We would need to limit the upload URL to a short period something like 2 minutes.
The client should not able to control the expiry in this case.

The good news is that, 
the short expiry time is simply the amount of time the upload URL is available to be uploaded, if the upload last longer than 2 minutes it will still be going through just fine.

src/http/routes/object/getSignedUploadURL.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  fastify.post<getSignedURLRequestInterface>(
+    '/upload/sign/:bucketName/*',
+    {
+      schema,
+    },
+    async (request, response) => {
+      const { bucketName } = request.params
+      const objectName = request.params['*']
+      const { expiresIn } = request.body
```
fenos: As mentioned in the comment above, we should not accept the `expiresIn` value from the client. We can promote this value as an environment variable with the default of 60s 

That is probably what i would run in production

fenos (approved):

acupofjose:
Question! How does this work with Bucket policies? Are bucket policies ignored?

farchond:
> The good news is that the short expiry time is simply the amount of time the upload URL is available to process the upload. if the upload last longer than 2 minutes it will still be going through just fine.

I've been playing around with this recently and I've found this to not be the case. When I upload a large file that takes longer than 60s to finish, I get a 400 response. Like S3, the upload URL appears to need to be valid at the time of upload completion, not just upload commencement. With this in mind, signed upload URLs become almost useless since large files or poor network conditions would cause a file to be un-uploadable :/

If my assessment is correct, would you consider adding back the `expiresIn` parameter with the caveat that owners should be mindful of the security risk it incurs? Thanks so much for your time and attention, @fenos

fenos:
Hey @farchond the upload URL is valid for only 60s but the upload can take as long as it needs, it is strange that you get a 400 - are you trying to overwrite a file?

farchond:
@fenos Thanks for the quick reply! We're definitely not trying to overwrite any file.

We've tried with many different scenarios:
- Small/medium files with an upload starting at close to 60s, so it ends after 60s -> 400
- Small/medium files with a <60s upload -> success
- Large file uploads (~1 gig) started immediately which take about 2 minutes to complete -> 400

This behavior didn't surprise us as it directly matches S3's behavior with regards to signed upload urls. The only exception being that S3 gives a 403, not a 400. Since Supabase storage uses S3 under the hood this all seems to fit to me, unless I'm missing something!

fenos:
Hey @farchond no worries! Happy to help!

I'm going to investigate this. Just for curiosity what error are you getting in the 400 response?

farchond:
Here's the full response:

```
<NSHTTPURLResponse: 0x282c376e0> { URL: https://tyymaqwprqsupipyalpj.supabase.co/storage/v1/object/upload/sign/test-audio/acbdbc9b-edf4-4ea2-83ba-d814ff33a265/58753dc5a845aeb15c272633cf462702a3af56ad.wav?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1cmwiOiJ0ZXN0LWF1ZGlvL2FjYmRiYzliLWVkZjQtNGVhMi04M2JhLWQ4MTRmZjMzYTI2NS81ODc1M2RjNWE4NDVhZWIxNWMyNzI2MzNjZjQ2MjcwMmEzYWY1NmFkLndhdiIsImlhdCI6MTY4MTc1OTgwNywiZXhwIjoxNjgxNzU5ODY3fQ.fRyQM1yyTgLOJBUYd-oZrMrcx-JWsV3u-8rLOwRkkGk } { Status Code: 400, Headers {
    "Access-Control-Allow-Origin" =     (
        "*"
    );
    "Alt-Svc" =     (
        "h3=\":443\"; ma=86400, h3-29=\":443\"; ma=86400"
    );
    "Content-Length" =     (
        66
    );
    "Content-Type" =     (
        "application/json; charset=utf-8"
    );
    Date =     (
        "Mon, 17 Apr 2023 19:36:34 GMT"
    );
    Server =     (
        cloudflare
    );
    "Strict-Transport-Security" =     (
        "max-age=2592000; includeSubDomains"
    );
    Vary =     (
        "Accept-Encoding"
    );
    "cf-cache-status" =     (
        DYNAMIC
    );
    "cf-ray" =     (
        "7b9714af4dbb3350-EWR"
    );
    "sb-gateway-mode" =     (
        direct
    );
    "sb-gateway-version" =     (
        1
    );
} }
```

fenos:
Are you not getting any body with the 400 response?

farchond:
Ah you're right - 

```
 {"statusCode":"400","error":"Invalid JWT","message":"jwt expired"}
```

farchond:
Thank you so much @fenos! Let us know what you find.

fenos:
It is my pleasure @farchond 

I will keep you posted

fenos:
@farchond right, I've deep-dived into this and my theory was correct.

We have a WAF service buffering part of the upload before reaching the origin, after some internal discussion we decided to increase the signed upload URL to 2h - which will allow you to upload any file size

In the near future, we'll be giving the opportunity to set an expiration time in the request.

We'll be rolling the change out to our platform today / tomorrow across all regions

farchond:
Thank you so much @fenos! We really appreciate the timely replies and willingness to help us out and improve the platform in the meantime!

fenos:
@scottcorgan what version of supabase cli are you using? it might not have the latest changes of storage

fenos:
@farchond this is now deployed globally 🎉 - let me know if it works for you

fenos:
@scottcorgan try to upgrade it and see how it goes, not entirely sure about the supabase cli version - let me know if updating works for you - if it doesn't i'll reach out to the cli team

PR #284, opened 2023-03-02
feat: return object id on object operations

supermar1010:

feature

currently is not possible to create relationships from the client side to an object because the id is never returned

## What is the new behavior?

object id is returned on the following operations:

- `createObject`
- `copyObject`
- `updateObject`

## Additional context

Reimplemented this PR. https://github.com/supabase/storage-api/pull/197

One remark, when testing this locally I had problems with RLS. I couldn't insert anything because it always failed with a RLS `new row violates row-level security policy for table "objects"`. 

What a problem is, is the RLS. Because by removing the `returning: 'minimal'`. You need the select permission to insert a file right now. 

I don't know if that's a problem?

### DIFF
--- a/src/http/routes/object/copyObject.ts
+++ b/src/http/routes/object/copyObject.ts
@@ -15,10 +15,12 @@ const copyRequestBodySchema = {
 const successResponseSchema = {
   type: 'object',
   properties: {
+    id: { type: 'string', examples: ['2eb16359-ecd4-4070-8eb1-8408baa42493'] },
     Key: { type: 'string', examples: ['folder/destination.png'] },
   },
-  required: ['Key'],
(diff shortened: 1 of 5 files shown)


GaryAustin1:
What is the intent with storing the object id versus just path in your associated tables?  It is true you also have to save bucket name if the bucket is not fixed for your structure and save a bit of database space.  But the cost is you have to read the storage table directly to do any followup storage-api call. 

The object id has no operation it can be used for directly.  My fear in dealing with user issues, is that if this is implemented, it further gives users the impression they can just delete the storage.objects row and the associated s3 file data is deleted also.  Which is not the case.

fenos:
This is now being implemented

PR #288, opened 2023-03-20
Feat TUS integration with Knex

fenos:

Feature

## What is the new behavior?

- Replace Postgrest with Knex
- TUS protocol implementation


src/app.ts
```diff
@@ -55,10 +55,15 @@ const build = (opts: buildOpts = {}): FastifyInstance => {
   app.register(plugins.tenantId)
   app.register(plugins.logTenantId)
   app.register(plugins.logRequest({ excludeUrls: ['/status'] }))
+  app.register(routes.multiPart, { prefix: 'multi-part' })
   app.register(routes.bucket, { prefix: 'bucket' })
   app.register(routes.object, { prefix: 'object' })
   app.register(routes.render, { prefix: 'render/image' })
 
+  app.post('/dummy-webhook', (req, reply) => {
```
inian: remove?

package.json
```diff
@@ -32,9 +32,14 @@
     "@fastify/rate-limit": "^7.6.0",
     "@fastify/swagger": "^7.4.1",
     "@supabase/postgrest-js": "^0.36.0",
+    "@tus/file-store": "^1.0.0-beta.1",
+    "@tus/s3-store": "https://gitpkg.now.sh/supabase/tus-node-server/packages/s3-store/dist?build",
```
inian, line 35: what would be the flow here? code changes get approved and merged to the build branch?

src/database/connection.ts
```diff
@@ -0,0 +1,116 @@
+import pg from 'pg'
+import { ConnectionString } from 'connection-string'
+import LRU from 'lru-cache'
+
+import { StorageBackendError } from '../storage'
+import { Knex, knex } from 'knex'
+import { verifyJWT } from '../auth'
+import { JwtPayload } from 'jsonwebtoken'
+
+pg.types.setTypeParser(20, 'text', parseInt)
```
inian, line 8: what is this for?

src/database/connection.ts
```diff
@@ -0,0 +1,116 @@
...
+  role: string
+  jwt: JwtPayload
+  jwtRaw: string
+  jwtSecret: string
+}
+
+export const connections = {
+  leastUsed: undefined,
+  values: new LRU<string, Promise<Knex>>({
+    max: 200,
```
inian: 200 is a large number. That's [all of the connections](https://supabase.com/docs/guides/platform/performance#optimizing-the-number-of-connections) for a free project. I think we are better off setting a much smaller max number. Let's take this in as an env var, so that we can fine tune it based on the container size and number of containers running.

src/http/plugins/jwt.ts
```diff
@@ -12,8 +12,9 @@ declare module 'fastify' {
 export const jwt = fastifyPlugin(async (fastify) => {
   fastify.decorateRequest('jwt', '')
   fastify.addHook('preHandler', async (request, reply) => {
-    request.jwt = (request.headers.authorization as string).substring('Bearer '.length)
+    request.jwt = (request.headers.authorization || '').substring('Bearer '.length)
 
+    console.log(request.jwt)
```
inian: dont print jwt

src/http/routes/object/getSignedURLs.ts
```diff
@@ -32,7 +32,7 @@ const successResponseSchema = {
         error: ['string', 'null'],
         examples: ['Either the object does not exist or you do not have access to it'],
       },
-      path: {
```
inian, line 35: does signedURLs return version? seems to only return path

src/monitoring/logger.ts
```diff
@@ -14,6 +14,7 @@ export const logger = pino({
   },
   serializers: {
     res(reply) {
+      // console.log(reply)
```
inian, line 17: remove

src/server.ts
```diff
@@ -61,6 +61,11 @@ const exposeDocs = true
     console.log(`Server listening at ${address}`)
   })
 
+  // process.on('uncaughtException', (e) => {
```
inian: why are we removing this?

src/database/connection.ts
```diff
@@ -0,0 +1,116 @@
...
+        pool: {
+          min: 0,
+          max: isExternalPool ? 200 : poolSize,
+          idleTimeoutMillis: 5000,
+        },
+        connection: connectionString.toString(),
+        acquireConnectionTimeout: 1000,
+      })
+
+      await k.raw(`SELECT 1`)
```
inian: why do we need this Fabrizio?

src/database/connection.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  transaction(isolation?: Knex.IsolationLevels, instance?: Knex) {
+    return (instance || this.pool).transactionProvider({
+      isolationLevel: isolation,
+    })
+  }
+
+  async setScope(tnx: Knex) {
+    await tnx.raw(`
+      SET LOCAL ROLE ${this.options.role};
```
inian: there might be some folks also relying on [request.headers](https://postgrest.org/en/stable/api.html#accessing-request-headers-cookies-and-jwt-claims) being set. I remember some customer using cloudflare headers in their RLS policies!

src/database/client.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    const tenant = await getTenantConfig(options.tenantId)
+    url = tenant.databaseUrl
+    jwtSecretKey = tenant.jwtSecret
+  }
+
+  const verifiedJWT = await verifyJWT(jwt, jwtSecretKey)
+
+  if (!verifiedJWT) {
```
inian, line 66: verifyJWT throws in case of a decoding error instead of returning a null-ish object

package.json
fenos: Yes, currently using a forked version of tus-node-server on the build branch - will need to set up CI there

src/database/connection.ts
fenos: apparently postgres bigints columns are converted to strings by knex / js, this will convert them in numbers

fenos: Ref: https://github.com/knex/knex/issues/387#issuecomment-51554522

src/database/connection.ts
fenos: Absolutely! will need to revise the pooling size

src/http/plugins/jwt.ts
fenos: eheh debugging artifacts

src/server.ts
fenos: no, debugging artifacts

src/database/connection.ts
fenos: knex uses a lazy approach to establish a connection, by issuing a select 1 we are testing that the connection is healthy

src/database/client.ts
fenos: Unfortunately I need to make typescript happy  😄

src/database/connection.ts
fenos: Will include in the next minor release

src/database/connection.ts
inian: ah i see, add this comment to the code too?

src/config.ts
```diff
@@ -170,7 +186,14 @@ export function getConfig(): StorageConfigType {
     signedUploadUrlExpirationTime: parseInt(
       getOptionalConfigFromEnv('SIGNED_UPLOAD_URL_EXPIRATION_TIME') || '60'
     ),
+
+    tusPath: getOptionalConfigFromEnv('TUS_URL_PATH') || '/upload/resumable',
```
inian, line 190: why are we making this configurable?

src/database/connection.ts
```diff
@@ -0,0 +1,148 @@
...
+    return (instance || this.pool).transactionProvider({
+      isolationLevel: isolation,
+    })
+  }
+
+  async setScope(tnx: Knex) {
+    const headers = JSON.stringify(this.options.headers || {})
+    await tnx.raw(
+      `
+        SELECT
```
inian, line 133: @steve-chavez is this sufficient to mimic the config postgrest passes into the query?

src/http/routes/tus/handlers.ts
```diff
@@ -0,0 +1,248 @@
...
+  getFileIdFromRequest(rawRwq: http.IncomingMessage) {
+    return getFileIdFromRequest(rawRwq, this.options.path)
+  }
+
+  async send(rawReq: http.IncomingMessage, res: http.ServerResponse) {
+    const req = rawReq as MultiPartRequest
+
+    const id = this.getFileIdFromRequest(req)
+    if (id === false) {
+      throw ERRORS.FILE_NOT_FOUND
```
inian, line 61: Can we keep the message consistent with the 404 response from our normal upload endpoints.

src/http/routes/tus/handlers.ts
```diff
@@ -0,0 +1,248 @@
...
+    if (process.env.NODE_ENV === 'production') {
+      req.headers['x-forwarded-proto'] = 'https'
+    }
+
+    id = Buffer.from(id, 'utf-8').toString('base64url')
+    return super.generateUrl(req, id)
+  }
+}
+
+export class Options extends OptionsHandler {
```
inian, line 141: Options requests are handled at the api-gateway level and wouldnt reach storage-api

src/http/routes/tus/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    storage: Storage
+    owner?: string
+    tenantId: string
+  }
+}
+
+function createTusStore() {
+  if (storageBackendType === 's3') {
+    return new S3Store({
+      partSize: 6 * 1024 * 1024, // Each uploaded part will have ~8MB,
```
inian: how does the part size come out to 8MB?

src/storage/backend/s3.ts
```diff
@@ -148,6 +148,27 @@ export class S3Backend implements StorageBackendAdapter {
     }
   }
 
+  async updateObjectInfoMetadata(bucketName: string, key: string) {
```
inian: rename to setMetadataToComplete or something

src/storage/database/knex.ts
```diff
@@ -0,0 +1,567 @@
...
+        throw new DBError('resource_locked', 409, 'Resource is locked')
+      }
+
+      return true
+    })
+  }
+
+  async waitObjectLock(bucketId: string, objectName: string, version?: string) {
+    return this.runQuery('WaitObjectLock', async (knex) => {
+      const hash = hashStringToInt(`${bucketId}/${objectName}${version ? `/${version}` : ''}`)
```
inian, line 426: should the lock include the tenantID?

src/config.ts
fenos: This value is used to generate the upload link, since we run behind a reverse proxy our prefix will be different ex: `storage/v1/...`

src/http/routes/tus/index.ts
fenos: ops forgot to update the comment

src/storage/database/knex.ts
fenos: in theory we are already on the tenant db at this point

inian: oh yeah this is lock on their db, makes sense

inian (approved):

src/database/connection.ts
steve-chavez: Here's the full list:

- https://postgrest.org/en/stable/api.html#http-context
- https://postgrest.org/en/stable/api.html#accessing-request-path-and-method

So:

- `request.cookies` is missing, thought this is just a convenience. It's already included in the headers.
- `request.method` and `request.path` also missing, but not sure if there's a use case for this on storage policies.

---

Additionally, postgREST also enforces a `search_path` so it does `set_config('search_path',..`) which for Supabase is defined on the API settings(by default is `public, storage, graphql_public`).

Again, not sure how relevant is that for storage.

PR #293, opened 2023-03-30
feat: backwards compatible object versioning

fenos:

Feat

## What is the new behavior?

In preparation for a major refactor to the data layer of storage-api, this PR will make the future code backwards compatible


inian (approved):

hetaochen:
Can it be made optional? 
I use storage-api manage my files,  and mount a S3 bucket via FUSE in another container.

PR #301, opened 2023-04-08
chore: add some missing tags and summaries for docs

inian:
we need to come up with the json schema for the tus uploads though. This at least shows some description on https://supabase.github.io/storage-api


src/http/routes/render/renderPublicImage.ts
```diff
@@ -10,7 +10,6 @@ const renderPublicImageParamsSchema = {
   type: 'object',
   properties: {
     bucketName: { type: 'string', examples: ['avatars'] },
-    download: { type: 'string' },
```
inian, line 13: I don't think download needs to be in the params schema

fenos (approved):

PR #308, opened 2023-04-11
fix: mime-types and max-file-size

fenos:

Bug fix

## What is the new behavior?

- Allow null for `file_size_limit` and `allowed_mime_types`
- Add more strict validations on mime types

### DIFF
--- a/src/storage/storage.ts
+++ b/src/storage/storage.ts
@@ -86,10 +86,17 @@ export class Storage {
 
     const bucketData: Parameters<Database['createBucket']>[0] = data
 
-    if (data.fileSizeLimit) {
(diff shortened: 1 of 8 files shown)


src/storage/storage.ts
```diff
@@ -86,10 +86,17 @@ export class Storage {
 
     const bucketData: Parameters<Database['createBucket']>[0] = data
 
-    if (data.fileSizeLimit) {
+    if (typeof data.fileSizeLimit === 'number' || typeof data.fileSizeLimit === 'string') {
       bucketData.file_size_limit = await this.validateMaxSizeLimit(data.fileSizeLimit)
```
inian: should probably be named as parseMaxSizeLimit since it also does the string to number conversion or the parse and validate should be separated out

inian (approved):

PR #311, opened 2023-04-13
fix: use batches for queue

fenos:

Feature

## What is the new behavior?

Use batches when for webhook queue


src/queue/events/webhook.ts
```diff
@@ -3,8 +3,11 @@ import { Job, WorkOptions } from 'pg-boss'
 import axios from 'axios'
 import { getConfig } from '../../config'
 import { logger } from '../../monitoring'
+import rateLimit from 'axios-rate-limit'
 
-const { webhookURL, webhookApiKey, webhookQueuePullInterval } = getConfig()
+const { webhookURL, webhookApiKey, webhookQueuePullInterval, webhookQueueBatchSize } = getConfig()
+
+const httpClient = rateLimit(axios.create(), { maxRPS: 100 })
```
inian, line 10: we should probably make this an env var

src/queue/events/base-event.ts
```diff
@@ -46,6 +46,20 @@ export abstract class BaseEvent<T extends Omit<BasePayload, '$version'>> {
     return {}
   }
 
+  static ack(job: Job<BaseEvent<any>['payload']> | Job<BaseEvent<any>['payload']>[]) {
```
inian, line 49: complete is a better word for this?

inian (approved):

PR #313, opened 2023-04-17
fix: limit connection pool to 1 when using external pool

fenos:

Bug fix

## What is the new behavior?

Limit knex connection pool to 1 when using an external pool

### DIFF
--- a/src/storage/uploader.ts
+++ b/src/storage/uploader.ts
@@ -36,28 +36,26 @@ export interface UploadObjectOptions {
 export class Uploader {
   constructor(private readonly backend: StorageBackendAdapter, private readonly db: Database) {}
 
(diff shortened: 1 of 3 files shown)


src/storage/uploader.ts
```diff
@@ -36,28 +36,26 @@ export interface UploadObjectOptions {
 export class Uploader {
   constructor(private readonly backend: StorageBackendAdapter, private readonly db: Database) {}
 
-  canUpload(options: Pick<UploadObjectOptions, 'bucketId' | 'objectName' | 'isUpsert'>) {
-    return this.db.withTransaction(async (db) => {
```
fenos, line 40: We don't need this to be in a transaction, so i just unwrapped it

pcnc (approved):

PR #320, opened 2023-04-21
fix: release connection when can't switch to role

fenos:

When the pg user doesn't have permission to switch to a role, the transaction was never rolled back

Bug fix

## What is the new behavior?

Rollback the transaction

### DIFF
--- a/src/storage/database/knex.ts
+++ b/src/storage/database/knex.ts
@@ -46,16 +46,16 @@ export class StorageKnexDB implements Database {
           this.options.tnx
         )()
 
-        await this.connection.setScope(tnx)
-
-        tnx.on('query-error', (error) => {
-          throw DBError.fromDBError(error)
-        })
-
-        const opts = { ...this.options, tnx }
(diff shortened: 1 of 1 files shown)


inian (approved):

src/storage/database/knex.ts
```diff
@@ -467,12 +467,15 @@ export class StorageKnexDB implements Database {
...
-    } else if (differentScopes) {
-      await this.connection.setScope(tnx)
     }
 
     try {
+      if (needsNewTransaction) {
+        await this.connection.setScope(tnx)
+      } else if (differentScopes) {
+        await this.connection.setScope(tnx)
+      }
```
pcnc: Maybe it's more concise and/or readable if it's set like this?

```suggestion
      if (needsNewTransaction || differentScopes) {
        await this.connection.setScope(tnx)
      }
```

pcnc (approved):

src/storage/database/knex.ts
fenos: Way more readable, good catch

PR #336, opened 2023-05-24
fix: pass owner to test permission

fenos:

Bug fix

## What is the new behavior?

Passing the owner column on previously missing test permission checks will make sure that if you wrote RLS
against the owner column, they will pass

### DIFF
--- a/src/storage/object.ts
+++ b/src/storage/object.ts
@@ -269,6 +269,7 @@ export class ObjectStorage {
         db.updateObject(bucketId, sourceKey, {
           name: sourceKey,
           version: newVersion,
+          owner: owner,
         }),
       ])
     })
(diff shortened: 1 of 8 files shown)


src/storage/object.ts
```diff
@@ -352,10 +331,16 @@ export class ObjectStorage {
     await this.db.testPermission((db) => {
       return Promise.all([
         db.findObject(this.bucketId, sourceObjectName, 'id'),
-        db.createObject({
+        db.updateObject(this.bucketId, sourceObjectName, {
+          name: sourceObjectName,
+          version: '1',
+          owner,
+        }),
+        db.asSuperUser().createObject({
```
inian, line 341: why is this testpermission check required?

fenos: If the object already exists we don't want to start the operation of moving the file on S3 - by trying to create the object we can quickly verify if an object already exists at the destination.

Previously the `move` method was requiring `select + insert` RLS policy but now will require `select + update` so we still need to verify if the destination object is not yet created

inian (approved):

src/storage/object.ts
inian: probably better to have it as a separate check? or add a comment.. clubbing it in testpermission isnt obvious cos I was thinking about what permissions is this clause checking for (esp when it is being run as a superuser?

PR #340, opened 2023-06-13
feat: deprecate owner column in favour of owner_id

fenos:

feature

## What is the new behavior?

- Deprecate `owner` column in favor of `owner_id` 
- owner_id is now of type `text`
- Write to both columns if the owner is uuid, otherwise only on the `owner_id` column

### DIFF
--- a/migrations/tenant/0017-add_owner_id_column_deprecate_owner.sql
+++ b/migrations/tenant/0017-add_owner_id_column_deprecate_owner.sql
@@ -1,0 +1,8 @@
+alter table storage.objects add column if not exists owner_id text default null;
+alter table storage.buckets add column if not exists owner_id text default null;
+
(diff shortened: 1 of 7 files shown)


migrations/tenant/0017-add_owner_id_column_deprecate_owner.sql
```diff
@@ -0,0 +1,8 @@
+alter table storage.objects add column if not exists owner_id text default null;
+alter table storage.buckets add column if not exists owner_id text default null;
+
+comment on column storage.objects.owner is 'Field is deprecated, use owner_id instead';
+comment on column storage.buckets.owner is 'Field is deprecated, use owner_id instead';
+
+ALTER TABLE storage.buckets
+    DROP CONSTRAINT IF EXISTS buckets_owner_fkey;
```
inian: newline

inian (approved):

PR #371, opened 2023-09-29
feat: standardise logs

fenos:

Improvement

Currently, logs are a bit all over the place with different shapes.

## What is the new behavior?

This PR tries to standardize logs across the service and include more useful information such as:

- owner
- reqId
- headers:
  -  'content-type',
  - 'if-none-match',
  - 'if-modified-since',
  - 'upload-metadata',
  - 'upload-length',
  - 'tus-resumable',

Lifecycle events are now also logged and you can find the lifecycle of a specific asset by filtering with:

```
m.objectPath:"{ref}/{bucket}/{object}
```

additionally, you can filter by the following properties:

```
m.type:"event" m.event:~"ObjectCreated" m.project:""
```

All events will also include a `reqId` field which can be correlated to the request that was issued from.

```
m.type:"request" m.reqId:"reqId"
```

### DIFF
--- a/src/monitoring/logflare.ts
+++ b/src/monitoring/logflare.ts
@@ -1,5 +1,7 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 
+import { defaultPreparePayload } from 'pino-logflare'
+
 const dotenv = require('dotenv')
 const { createWriteStream: createLogFlareWriteStream } = require('pino-logflare')
 
@@ -20,5 +22,10 @@ export default function () {
   return createLogFlareWriteStream({
     apiKey: logflareApiKey,
     sourceToken: logflareSourceToken,
+    onPreparePayload: (payload: any, meta: any) => {
+      const item = defaultPreparePayload(payload, meta)
+      item.project = payload.project
+      return item
+    },
   })
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
         "pg-boss": "^8.1.1",
         "pg-listen": "^1.7.0",
         "pino": "^8.2.0",
-        "pino-logflare": "^0.3.12",
+        "pino-logflare": "^0.4.2",
         "postgres-migrations": "^5.3.0",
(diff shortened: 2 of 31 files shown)


src/monitoring/logflare.ts
```diff
@@ -20,5 +20,8 @@ export default function () {
   return createLogFlareWriteStream({
     apiKey: logflareApiKey,
     sourceToken: logflareSourceToken,
+    onPreparePayload: (payload: any) => {
+      return payload
```
Ziinc: Screenshot 2023-09-29 at 8 25 10 PM

would need to set the project key here 

https://github.com/Logflare/pino-logflare/blob/master/docs/API.md

fenos: @Ziinc the project key is already inferred from the log-middleware https://github.com/supabase/storage-api/pull/371/files#diff-7657d30e7d33d62772ae092233fbe7844f1601fbda2d5ebe6252b220988928abR7

do you think is still worth calling `defaultPreparePayload(payload, meta)`?

If yes, would this make sense:

```ts
onPreparePayload: (payload: any, meta: any) => {
      const item = defaultPreparePayload(payload, meta)
      item.project = payload.project
      return item
    },
```

Ziinc: Yea it is needed as the key needs to be a top level field for enabling clustering

src/monitoring/logflare.ts
```diff
@@ -20,5 +22,10 @@ export default function () {
   return createLogFlareWriteStream({
     apiKey: logflareApiKey,
     sourceToken: logflareSourceToken,
+    onPreparePayload: (payload: any, meta: any) => {
+      const item = defaultPreparePayload(payload, meta)
+      item.project = payload.project
```
inian, line 27: @supabase/logging I want to avoid `project` as the top level key. Can we cluster by `tenant` instead?

inian (approved):

PR #372, opened 2023-10-01
fix: increase queue concurrency

fenos:

Performance improvement

Queue jobs are fetched and processed one at the time

## What is the new behavior?

We increase the number of jobs fetched in a single call to reduce overall queries issued to the database as well as decrease the load.

We also run 5 concurrent jobs at any point instead of 1

### DIFF
--- a/src/queue/events/object-admin-delete.ts
+++ b/src/queue/events/object-admin-delete.ts
@@ -1,6 +1,6 @@
 import { BaseEvent, BasePayload } from './base-event'
 import { getConfig } from '../../config'
-import { Job } from 'pg-boss'
+import { Job, WorkOptions } from 'pg-boss'
 import { withOptionalVersion } from '../../storage/backend'
 import { logger } from '../../monitoring'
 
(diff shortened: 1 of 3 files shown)


src/queue/events/object-admin-delete.ts
```diff
@@ -15,6 +15,13 @@ const { globalS3Bucket } = getConfig()
 export class ObjectAdminDelete extends BaseEvent<ObjectDeleteEvent> {
   static queueName = 'object:admin:delete'
 
+  static getWorkerOptions(): WorkOptions {
+    return {
+      teamSize: 50,
```
inian: We need to make this configurable

inian:
we have another PR using the batchSize option of pgboss instead - https://github.com/supabase/storage-api/pull/311/files

Both seem like valid approaches. Batchsize seems like it will use lesser resources for the storage server due to the teamconcurrency still being 1 while achieving lower DB load too.

src/queue/events/object-admin-delete.ts
fenos: Done

inian (approved):

PR #379, opened 2023-10-11
fix: configure pgBoss from env

fenos:

Improvement

## What is the new behavior?

- Updated pg-boss from 8.1.1 -> 9.0.3
- queue configurations are now possible to edit via ENV
- Changed the following default configurations:
  -  deleteAfterDays: 7 -> 2
  - archiveCompletedAfterSeconds: 14_400 (4h) -> 7200 (2h)
  - retentionDays: 7 -> 2

- Handle pgBouncer server_login issue


package.json
```diff
@@ -55,7 +55,7 @@
     "knex": "^2.4.2",
     "md5-file": "^5.0.0",
     "pg": "^8.10.0",
-    "pg-boss": "^8.1.1",
+    "pg-boss": "^9.0.3",
```
inian, line 58: any breaking changes to be aware of here?

inian (approved):

package.json
fenos: Yes, but nothing that would affect us:
https://github.com/timgit/pg-boss/releases/tag/9.0.0

PR #388, opened 2023-10-23
performance / fix remove listeners on query and connection

fenos:

Bug fix

Listeners seems to be still registered even when the connection is closed

## What is the new behavior?

Remove all unnecessary listeners


src/database/connection.ts
```diff
@@ -88,6 +88,7 @@ export class TenantConnection {
     DbActivePool.inc({ tenant_id: options.tenantId, is_external: isExternalPool.toString() })
 
     knexPool.client.pool.on('createSuccess', () => {
+      console.log('successs')
```
kamilogorek:
```suggestion
```

kamilogorek (approved):

PR #402, opened 2023-12-04
feat: add gravity support for imgproxy 

egor-romanov:

feat: add gravity param support for imgproxy transformation feature

## Additional context

https://imgproxy.net/blog/perfect-cut-smart-image-resizing-with-imgproxy/

### DIFF
--- a/src/storage/renderer/image.ts
+++ b/src/storage/renderer/image.ts
@@ -325,37 +325,34 @@ export class ImageRenderer extends Renderer {
       segments.push(`format:${options.format}`)
     }
 
-    if (options.gravity) {
-      switch (options.gravity) {
-        case 'sm': {
-          segments.push(`gravity:${options.gravity}`)
-          break
-        }
-        case 'fp': {
-          const { x_offset: xOffset, y_offset: yOffset } = options
-          if (
-            xOffset === undefined ||
-            yOffset === undefined ||
-            !Number.isFinite(xOffset) ||
-            !Number.isFinite(yOffset) ||
-            xOffset < 0 ||
-            yOffset < 0 ||
-            xOffset > 1 ||
-            yOffset > 1
-          ) {
-            throw ERRORS.InvalidParameter('gravity', {
-              message: 'Focal point requires x and y coordinates within 0-1 range',
-            })
-          }
-          segments.push(`gravity:${options.gravity}:${xOffset}:${yOffset}`)
-          break
-        }
-        default: {
-          segments.push(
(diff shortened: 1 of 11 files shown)


src/storage/renderer/image.ts
```diff
@@ -170,6 +207,15 @@ export class ImageRenderer extends Renderer {
         case 'quality':
           all.quality = parseInt(value, 10)
           break
+        case 'gravity':
+          all.gravity = value
+          break
+        case 'x_offset':
+          all.x_offset = parseFloat(value)
```
fenos: i think this wouldn't work since currently, we are splitting on `:` to get the transformation type and value, however, gravity has the following format: `gravity:sm:100:100`

So value would just be `sm` in this case. We would need to change slightly the way we parse the string for the gravity

egor-romanov: not sure i got you, from what i understood we parse string of our own format, isn't it?

```
this.transformOptions = params.reduce((all, param) => {
const [name, value] = param.split(':') as [keyof TransformOptions, any]
```
And TransformOptions is our own interface where we specify `gravity`, `x_offset`, `y_offset`. And then it is applied be method above.

Or do we rely on imgproxy native format here?

fenos: Ah! You are right!
Ignore the comment above

fenos (approved):

egor-romanov:
hey @fenos , we also need this one right?

https://github.com/supabase/api-gateway/pull/352

src/storage/renderer/image.ts
```diff
@@ -115,6 +118,40 @@ export class ImageRenderer extends Renderer {
...
+      switch (options.gravity) {
+        case 'sm': {
+          segments.push(`gravity:${options.gravity}`)
+          break
+        }
+        case 'fp': {
+          if (
+            !(
+              options.x_offset &&
+              options.y_offset &&
```
itslenny: Is `0` a valid value for this? This would throw if x or y is `0` because `0` is "falsey" ... maybe replace it with `isNaN`

ferhatelmas: good point, addressed

ferhatelmas:
@claude review always

fenos (approved):

egor-romanov:
haha, didn't expect this email!

I believe there were also some changes required in the api and i had pr open too

ferhatelmas:
hi @egor-romanov good memory but it's not needed anymore. Due to changes in between, it works automatically

PR #407, opened 2024-01-02
feat: add jwks support

hf:
Adds [JWKS](https://datatracker.ietf.org/doc/html/rfc7517) support, similar to what [PostgREST](https://postgrest.org/en/stable/references/auth.html#asymmetric-keys) supports.

**Why is this necessary?**

This opens up avenues where foreign JWTs like those issued by Firebase Auth, Cognito, Clerk, Auth0 can be used directly with Storage, bypassing Supabase Auth or other symmetric use cases. 

It also helps with secret rotation.

**How it works?**

Set the `JWT_JWKS` env var to a JWKS value. In a multi-tenant setup the `jwt_jwks` column should encode the JWKS as a string, which is then parsed.

**Security Information**

This only affects storage calls that are made "in the name of the user." Signed URLs still continue to use the secret and ignore the existence of the JWKS.

JWK support is built-in to Node.js, so no third-party libraries are used.


src/auth/jwt.ts
```diff
@@ -26,25 +33,136 @@ export type SignedUploadToken = {
...
+  if (jwks && jwks.keys && jwks.keys.length) {
+    const hasRSA = jwks.keys.find((key) => key.kty === 'RSA')
+    const hasECC = jwks.keys.find((key) => key.kty === 'EC')
+    const hasED = jwks.keys.find(
+      (key) => key.kty === 'OKP' && ((key as any).crv === 'Ed25519' || (key as any).crv === 'Ed448')
+    )
+
+    algorithms = [
+      jwtAlgorithm as jwt.Algorithm,
+      ...(hasRSA ? JWT_RSA_ALGOS : []),
```
hf, line 120: Add HS algos here.

src/auth/jwt.ts
```diff
@@ -26,25 +33,136 @@ export type SignedUploadToken = {
...
+  jwks: { keys: { kid?: string; kty: string }[] } | null
+) {
+  if (!jwks || !jwks.keys) {
+    return secret
+  }
+
+  if (JWT_HS_ALGOS.indexOf(header.alg as jwt.Algorithm) > -1) {
+    // JWT is using HS, find the proper key
+
+    if (!header.kid && header.alg == jwtAlgorithm) {
```
hf: ===

src/test/jwt.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      type?: string
+      options?: any
+      alg: string
+      kid?: string
+      publicKey: any
+      privateKey: any
+    }[] = [
+      { type: 'rsa', options: { modulusLength: 2048 }, alg: 'RS256' },
+      { type: 'ec', options: { namedCurve: 'P-256' }, alg: 'ES256' },
+      //{ type: 'ed25519', options: null, alg: 'EdDSA' },
```
hf: Requires tests to be running on node 20+.

fenos:
You will need to add a migration under `migrations/multitenant` to add the `jwt_jwks` column

hf:
> You will need to add a migration under `migrations/multitenant` to add the `jwt_jwks` column

Oh yeah, forgot about that!

hf:
@fenos Can you check why the tests are failing -- I can't seem to figure out what the error is.

fenos (approved):

PR #409, opened 2024-01-04
Feat: Standalone Storage Service

fenos:

Standalone Storage Service

Base migrations are now part of the service.
Storage can now be run as a fully standalone service.

Fully decoupled from auth,realtime,postgrest

### Migrations

Added a new migration `0002-storage-schema.sql` 

To keep backward compatibility and not break self-hosted instances the migration tool is now able to "refresh" the `migrations` table if it detects that the newly created migration is not present in the `migrations` records.

This refreshing mechanism only takes place for projects that have already run the migrations at least once.

New installations and new projects will start fresh and don't need to refresh

### Granular Roles

Introduced 4 new environment variables for new installations

- `DB_ANON_ROLE` default `anon`
- `DB_SERVICE_ROLE` default `service_role`
- `DB_AUTHENTICATED_ROLE` default `authenticated`
- `DB_INSTALL_ROLES` default `true`
- `DB_SUPER_USER` default `supabase_storage_admin`
- `DB_SEARCH_PATH` default ''

If these roles are already present in the database we can simply pass `DB_INSTALL_ROLES=false` and the migration will not install these roles


inian:
I still see base migrations has a dependency on a [extensions schema](https://github.com/supabase/storage/blob/2cef99de14891b3824d799b9c8ee5a3fa68d8d41/migrations/tenant/base/0001-supabase-schema.sql#L10) and [supabase_admin](https://github.com/supabase/storage/blob/2cef99de14891b3824d799b9c8ee5a3fa68d8d41/migrations/tenant/base/0001-supabase-schema.sql#L34). Do we really need supabase_admin role here? Shouldn't supabase_storage_admin role be sufficient?
Do we need realtime publication [here](https://github.com/supabase/storage/blob/2cef99de14891b3824d799b9c8ee5a3fa68d8d41/migrations/tenant/base/0001-supabase-schema.sql#L2)?

Any other schema, extension or role thats not required for the standalone functioning of storage?

inian (approved):

PR #421, opened 2024-01-16
Remove postgres extensions dependencies

fenos:

Feature

Currently, the storage schema relies on the `ossd-uuid` extension specifically on the `uuid_generate_v4()` function.

## What is the new behavior?

Storage drops support for PG 12 - and is now only compatible with > PG 13, which allows us to use the built-in function `gen_random_uuid` and remove dependencies on other extensions.

pgcrypto extension was also removed since there was no usage in the code-base

### DIFF
--- a/migrations/tenant/0019-alter-default-value-objects-id.sql
+++ b/migrations/tenant/0019-alter-default-value-objects-id.sql
@@ -1,0 +1,1 @@
+ALTER TABLE storage.objects ALTER COLUMN id SET DEFAULT gen_random_uuid();
--- a/migrations/tenant/0002-storage-schema.sql
+++ b/migrations/tenant/0002-storage-schema.sql
@@ -6,10 +6,6 @@ BEGIN
     END IF;
 END$$;
 
-
-create extension if not exists "uuid-ossp" WITH schema storage;
-create extension if not exists pgcrypto WITH schema storage;
-
 DO $$
 DECLARE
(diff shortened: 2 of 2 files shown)


migrations/tenant/0019-alter-default-value-objects-id.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE storage.objects ALTER COLUMN id SET DEFAULT gen_random_uuid();
```
inian, line 1: should be safe to run on large tables?

fenos: Changing the default value of a column in PostgreSQL typically does not require a lock that would prevent concurrent accesses to the table. When you alter a column to set a default value, PostgreSQL does not need to rewrite the table or the existing rows; instead, it just changes the metadata of the table to specify the new default for future inserts.

However, it's important to note a few things:

- Brief Access Exclusive Lock: The ALTER TABLE command acquires a brief ACCESS EXCLUSIVE lock on the specific table, which blocks all other accesses (reads and writes) to the table until the command is complete. But because changing the default value is usually a quick operation, the lock is typically very brief and may not be noticeable.

- No Impact on Existing Rows: Changing the default value of a column does not affect existing rows in the table. The new default value is only used for new rows that are inserted into the table without a specific value for that column.

inian (approved):

PR #422, opened 2024-01-17
Fix: use per request cache on tus metadata

fenos:

Bug fix

The metadata cache is stored in Map in memory. 
When running multiple instances creates issues as it never gets invalidated

## What is the new behavior?

Introducing a RequestCacheMetadata which is scoped to a single request.
This reduces the calls to fetch the metadata to 1 call per request instead of 8 per request.


fenos:
Will wait for:

- https://github.com/tus/tus-node-server/pull/558
- https://github.com/tus/tus-node-server/pull/552

to remove the ported s3Store and use the S3Store from the `@tus/s3-store` package.

Additionally, we can also remove the `Metadata.parse` call present on the `namingFunction` since it will be now be part of the @tus/server api 🎉

src/storage/tus/als-memory-kv.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export class AlsMemoryKV implements KvStore<MetadataValue> {
+  static localStorage = new AsyncLocalStorage<Map<string, MetadataValue>>()
+
+  async delete(value: string): Promise<void> {
+    AlsMemoryKV.localStorage.getStore()?.delete(value)
+  }
+
+  async get(value: string): Promise<MetadataValue | undefined> {
+    return AlsMemoryKV.localStorage.getStore()?.get(value)
```
inian, line 13: als?

inian (approved):

PR #428, opened 2024-01-25
fix: db install roles default env to false

fenos:

Bug fix

DB_INSTALL_ROLES default to true

## What is the new behavior?

DB_INSTALL_ROLES default to false


sweatybridge (approved):
thx!

PR #429, opened 2024-01-26
fix: add error handler on client

fenos:
Add pg client on error handler


src/http/routes/tus/lifecycle.ts
```diff
@@ -43,14 +39,19 @@ export async function onIncomingRequest(
     })
   })
 
+  if (rawReq.method === 'OPTIONS') {
```
inian, line 42: is this change meant to be part of this PR?

inian (approved):

PR #430, opened 2024-01-29
pgboss on error event handler

fenos:

Bug fix

Add on error callback to prevent unhandled exception


fenos:
Closing in favour #431

PR #431, opened 2024-01-29
Multitenant migrations execution

fenos:

Feature

Currently, migrations on tenants are run on the first request from when the server starts.
With this approach, there are a few different drawbacks:

- Expensive migrations (ex adding an index) cannot be run as this impacts the response time
- Build up connections to the database when concurrent requests come in

## What is the new behavior?

This PR introduces 3 migrations strategy modes:

- on_request
- full_fleet
- progressive

Migrations are now tracked in the multi-tenant db with `migrations_version` and `migrations_status`.
In all modes, migrations will be run only if a tenant is not on the latest version.

### On Request

**on_request mode** will run the migrations to a tenant database during a request. Migrations will block the request lifecycle until completed.

In case the migration fails, the request will also fail with a 500 error.

### Full Fleet

This mode offers a background scheduler to run migrations across all tenants asynchronously.

On Server Start, the instances race to acquire an advisory lock.
The instance that acquires the lock is responsible for selecting all tenants that need a migration to run.

We emit `RunMigrationsOnTenant` job for each tenant in a batch, we use a singleton key to make sure only 1 job for this specific tenant can exist at any one time.

Each Storage instance can consume the events and run the migrations on each single tenant database.
Migrations are retried automatically 3 times.

Failures don't impact the request lifecycle and the migrations will run on all tenants that are not on the latest version.

### Progressive

This mode is a hybrid version of the 2 above. It works similarly to the `full_fleet` mode since migrations are run with an async job, but the tenants are selected during a request.

During a tenant request, we determine if the tenant needs a migration to run. 
If it does we register the tenantId in a list in memory.

A background worker will pop a batch of 200 tenants from this list and schedule `RunMigrationsOnTenant` jobs every 1m.

Deduping between instances is done via the `singletonKey` and early checks. 

In case of failure, migrations are retried 3 times, after which the status becomes `FAILED_STALE`.
When using progressive migrations and the staus is `FAILED_STALE` we will continue running the migrations during the request lifecycle **without blocking the request** for this tenant, with the hope that it recovers from unexpected prolonged issues to the underline database.

## Additional context

With this change, we will see the following benefits:

- Increased performance on req/res time on server restart
- Fewer database connections to tenant databases
- Possibility to run long-running async migrations


Dockerfile
```diff
@@ -1,24 +1,36 @@
-FROM node:18-alpine
+# Base stage for shared environment setup
```
fenos, line 1: Improved Dockerfile to use node_modules cache on each step

.env.sample
```diff
@@ -24,20 +24,13 @@ RESPONSE_S_MAXAGE=0
...
-
 #######################################
 # Multi Tenancy
 # uncomment MULTI_TENANT=true to enable multi tenancy
-# the [Single Tenant] configurations will be ignored
 #######################################
 # MULTI_TENANT=true
 DATABASE_MULTITENANT_URL=postgresql://postgres:postgres@127.0.0.1:5433/postgres
-REQUEST_X_FORWARDED_HOST_REGEXP=
+REQUEST_X_FORWARDED_HOST_REGEXP=^([a-z]{20}).local.(?:com|dev)$
```
fenos, line 33: default value for multitenant x-forwarded-host: `{tenant-id}.local.dev` to make it easy to run Storage out of the box as multitenant

.env.sample
```diff
@@ -85,7 +78,7 @@ STORAGE_BACKEND=s3
 #######################################
 # S3 Backend
 #######################################
-STORAGE_S3_BUCKET=name-of-your-s3-bucket
+STORAGE_S3_BUCKET=supa-storage-bucket
```
fenos, line 81: default value for bucket name which is the same as the bucket we create on the docker-compose.
This makes it easy to simply rename the `.env.sample` to `.env` without any additional changes

README.md
```diff
@@ -23,9 +23,10 @@ A scalable, light-weight object storage service.
 
 - Copy `.env.sample` to `.env` file.
 - Copy `.env.test.sample` to `.env.test`.
-- Change `GLOBAL_S3_BUCKET` and `REGION` to the name and region of a S3 bucket.
```
fenos, line 26: Now simply rename `.env.sample` to `.env` and that's it, all default values will work out of the box

src/config.ts
```diff
@@ -258,8 +261,9 @@ export function getConfig(options?: { reload?: boolean }): StorageConfigType {
     logflareEnabled: getOptionalConfigFromEnv('LOGFLARE_ENABLED') === 'true',
     logflareApiKey: getOptionalConfigFromEnv('LOGFLARE_API_KEY'),
     logflareSourceToken: getOptionalConfigFromEnv('LOGFLARE_SOURCE_TOKEN'),
-    defaultMetricsEnabled:
-      getOptionalConfigFromEnv('DEFAULT_METRICS_ENABLED', 'ENABLE_DEFAULT_METRICS') === 'true',
+    defaultMetricsEnabled: !(
```
fenos, line 274: default metrics are now on by default

src/database/migrate.ts
```diff
@@ -346,7 +396,7 @@ async function refreshMigrationPosition(
   if (shouldUpdateMigrations) {
     await client.query(`BEGIN`)
     try {
-      await client.query(`DELETE FROM ${migrationTableName}`)
+      await client.query(`DELETE FROM ${migrationTableName} WHERE id is not NULL`)
```
fenos, line 472: running deletes without where statements seems to be blocked by default on staging / prod with a pg extension

src/http/routes/tenant/migrations.ts
```diff
@@ -0,0 +1,18 @@
...
+const { pgQueueEnable } = getConfig()
+
+export default async function routes(fastify: FastifyInstance) {
+  fastify.register(apiKey)
+
+  fastify.get('/', async (req, reply) => {
+    if (!pgQueueEnable) {
+      return reply.code(400).send({ message: 'Queue is not enabled' })
+    }
+    const queueSize = await Queue.getInstance().getQueueSize(RunMigrationsOnTenants.getQueueName())
```
fenos: admin route to see how many remaining migrations are pending at: `GET /admin/migrations`

payload:

```
{
  "remaining": 10
}
```

src/monitoring/logger.ts
```diff
@@ -60,17 +60,25 @@ interface ErrorLog {
...
+}
+
 export const logSchema = {
+  info: (logger: BaseLogger, message: string, log: InfoLog) => logger.info(log, message),
   request: (logger: BaseLogger, message: string, log: RequestLog) => {
-    logger.info(log, message)
+    const is4xxResponse = statusOfType(log.res.statusCode, 400)
+    const is5xxResponse = statusOfType(log.res.statusCode, 500)
+
+    const logLevel = is4xxResponse ? 'warn' : is5xxResponse ? 'error' : 'info'
```
fenos, line 80: Thanks to @egor-romanov to have pointed out that request logs were always info.
This changes the log level:

- anything 4xx is a warning
- anything 5xx is an error

src/queue/events/base-event.ts
```diff
@@ -42,20 +57,36 @@ export abstract class BaseEvent<T extends Omit<BasePayload, '$version'>> {
...
+  static getQueueOptions<T extends BaseEvent<any>>(payload: T['payload']): SendOptions | undefined {
     return undefined
   }
 
-  static getWorkerOptions(): WorkOptions {
+  static getWorkerOptions(): WorkOptions | BatchWorkOptions {
     return {}
   }
 
+  static batchSend<T extends BaseEvent<any>[]>(messages: T) {
```
fenos, line 84: Implemented batch insert for inserting multiple jobs in one insert statement

src/queue/events/run-migrations.ts
```diff
@@ -0,0 +1,56 @@
...
+import { getTenantConfig, updateTenantMigrationVersion } from '../../database/tenant'
+import { Job, SendOptions, WorkOptions } from 'pg-boss'
+import { lastMigrationName, runMigrationsOnTenant } from '../../database/migrate'
+import { logger, logSchema } from '../../monitoring'
+
+interface RunMigrationsPayload extends BasePayload {
+  tenantId: string
+}
+
+export class RunMigrationsOnTenants extends BaseEvent<RunMigrationsPayload> {
```
fenos, line 16: Job that handles tenant migrations

src/server.ts
```diff
@@ -28,6 +33,20 @@ const exposeDocs = true
   if (isMultitenant) {
     await runMultitenantMigrations()
     await listenForTenantUpdate(PubSub)
+
+    runMigrationsOnAllTenants()
```
fenos: on server start up, all storage api instances will race to acquire the lock.
The one who acquired the lock will create the jobs for running the migrations, the other instances who failed will simply return and do nothing.

src/database/migrations/migrate.ts
```diff
@@ -32,45 +38,151 @@ const backportMigrations = [
...
+/**
+ * Runs migrations for all tenants
+ * only one instance at the time is allowed to run
+ */
+export async function runMigrationsOnAllTenants(signal: AbortSignal) {
+  if (!pgQueueEnable) {
+    return
+  }
+  const result = await multitenantKnex.raw(`SELECT pg_try_advisory_lock(?);`, [
+    '-8575985245963000605',
```
inian, line 86: whats does this number represent?

inian (approved):

src/database/migrations/migrate.ts
fenos: the lock id

PR #453, opened 2024-04-15
Fix s3 file driver xattr

fenos:

Bug fix

## What is the new behavior?

- removed @types/ajv since they are shipped with the lib
- disable bodyParser for PutObject requests (improves performance)
- fix xattr metadata for s3 multi-part upload
- fix equality check for copying objects across buckets


sweatybridge (approved):
not sure if my +1 counts

inian (approved):

PR #460, opened 2024-04-23
feat: allow upload signed url to overwrite files

### LINKED ISSUE #380: Signed upload URL for existing files in blob storage
### Discussed in https://github.com/orgs/supabase/discussions/14538

<div type='discussions-op-text'>

<sup>Originally posted by **IzStriker** May 23, 2023</sup>
Hi.

I noticed it isn't possible to create a signed upload URL for existing objects in blob storage. Yet it seems that the storage-js library has the functionality to upsert files via a signed upload URL. Was there a specific reason the decision was made for this restriction, technical or otherwise? If not I would be more than happy to contribute to resolve this.

Generating signed upload URL for existing object error: https://github.com/supabase/storage-api/blob/81e9b7b6ed762f2f21bce9f39dd3b4cf7ce38a7a/src/storage/object.ts#LL534C3-L537C1

Upseritng file via supabase storage js client: https://github.com/supabase/storage-js/blob/b2e3486222126620cc65a5a1c1f967d25b4fb099/src/packages/StorageFileApi.ts#LL83C3-L85C8</div>

fenos:

feature

Currently is not possible to use signed upload URL to overwrite files

## What is the new behavior?

Allows providing x-upsert header for signed upload URL and overwrite files

Closes #380

### DIFF
--- a/src/storage/object.ts
+++ b/src/storage/object.ts
@@ -572,8 +572,15 @@ export class ObjectStorage {
    * @param url
    * @param expiresIn seconds
    * @param owner
+   * @param options
    */
-  async signUploadObjectUrl(objectName: string, url: string, expiresIn: number, owner?: string) {
+  async signUploadObjectUrl(
+    objectName: string,
+    url: string,
+    expiresIn: number,
+    owner?: string,
+    options?: { upsert?: boolean }
+  ) {
     // check as super user if the object already exists
     const found = await this.asSuperUser().findObject(objectName, 'id', {
       dontErrorOnEmpty: true,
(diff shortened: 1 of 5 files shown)


src/storage/object.ts
```diff
@@ -584,19 +591,21 @@ export class ObjectStorage {
...
-        bucket_id: this.bucketId,
-        name: objectName,
-        owner,
-        metadata: {},
-      })
+    await this.uploader.canUpload({
+      bucketId: this.bucketId,
+      objectName,
+      owner,
+      isUpsert: options?.upsert || false,
```
inian:
```suggestion
      isUpsert: options?.upsert ?? false,
```
in this case doesn't matter cos default value is false, but in general good to use ?? for checking for existence of a property

inian (approved):

softmarshmallow:
How do I create signed url for signed upload for existing files?
createSignedUploadUrl does not have a upsert option and will reject if called for existing objects.

PR #467, opened 2024-05-10
feat: operation based logging and RLS

fenos:

Feature

## What is the new behavior?

This PR introduces a couple of interesting approaches for operation-based RLS and logging semantics.

### Operation based RLS

Every Storage request will now include a Postgres local config called `storage.operation`. This variable will be very useful to write RLS policies that only allows certain operations.

Previously, when Granting a `SELECT` permission to specific objects, they were always available in either `List` endpoint or `Get` endpoint. However, sometime you'd like to allow the user to view the object but not show it in the `list` endpoint.

You can now achieve this by doing:

```
 CREATE POLICY "Allow anonymous download files but not listing" 
    ON storage.objects 
    AS PERMISSIVE FOR SELECT
    TO anon 
    USING (
      (bucket_id::text = 'my-bucket') AND
      (SELECT(current_setting('storage.operation', true) != 'object.list') 
    )
```

### Loggin

All logs will now include the operation and the resources associated to that operation.
This will allow to filter logs by a specific resource and see the whole lifecycle.

For example:

```
# Show all lifecycle of a resource
m.resources:@>"/bucket-1/file.jpg"

# Only uploads operations for this specific object
m.resources:@>"/bucket-1/file.jpg" m.operation:~'object.upload'
```


src/http/routes/object/copyObject.ts
```diff
@@ -37,15 +37,16 @@ export default async function routes(fastify: FastifyInstance) {
     '/copy',
     {
       schema,
+      config: {
+        operation: { type: 'object.copy' },
+        resources: (req: FastifyRequest<copyRequestInterface>) => {
+          const { sourceKey, destinationKey, bucketId, destinationBucket } = req.body
+          return [`${bucketId}/${sourceKey}`, `${destinationBucket || bucketId}/${destinationKey}`]
```
inian, line 45: did you confirm if array was the best way to log this? does logflare support searching an array?

fenos: Yes is possible to filter by arrays, @Ziinc told me he would implement regex matching too

Ziinc: In this case wouldn't it be more explicit and clear if it was two string fields though? One for source and one for destination.
If it was a variable length array then chucking it all in an array would make sense for searching, but it isn't so for this

fenos: Hey @Ziinc the `resources` fields might be different for different operation, in this case the resources for this operations are the source and destination.

The reason i'd like to use a single field for this is because if I want to query any operation for a specific resource I can simply query the array and I will get all the logs for this specific resource.

Ex, it has been copied, uploaded, delete, moved, etc...

Ziinc: I see gotcha 👍👍

inian (approved):

PR #468, opened 2024-05-14
feat: allow passing authorization via query string

fenos:

Feature

## What is the new behavior?

This PR allows passing the `?authorization` param as a query string param. Specifically, this behaviour is only enabled on the GET requests for fetching the asset: `/authenticated/:bucketName/*`

## Additional context

This functionality allows to embedding of authenticated images directly in HTML tags


fenos:
Closing for now as this is a non very secure practice to pass the token in the querystring

PR #494, opened 2024-05-28
feat: otel tracing

fenos:

Feature

## What is the new behavior?

This PR introduces optional OTEL traces


migrations/multitenant/0010-delete-cache-cache-notifier-on-delete.sql
```diff
@@ -0,0 +1,14 @@
+CREATE FUNCTION tenants_delete_notify_trigger ()
+    RETURNS TRIGGER
+AS $$
+BEGIN
+    PERFORM
+        pg_notify('tenants_update', '"' || OLD.id || '"');
```
inian, line 6: what is this used for?

monitoring/otel/config/otel-collector-config.yml
```diff
@@ -0,0 +1,183 @@
...
+                    },
+                },
+                {
+                  # trace.mode = premium
+                  name: trace-mode-default,
+                  type: string_attribute,
+                  string_attribute:
+                    {
+                      key: trace.mode,
+                      values: [ premium ],
```
inian: premium is a weird term for this. enhanced?

inian (approved):

migrations/multitenant/0010-delete-cache-cache-notifier-on-delete.sql
fenos: When a DELETE request to tenants is issued, we currently don't clear the cache across all storage-api services. 
Only on the instance that received the DELETE request.

By using this trigger on delete all containers will have a consistent view of the tenants cache

monitoring/otel/config/otel-collector-config.yml
fenos: Sure, can change to `enhanced` or `always_trace`

monitoring/otel/config/otel-collector-config.yml
```diff
@@ -0,0 +1,183 @@
...
+                    },
+                },
+                {
+                  name: success-status-codes,
+                  type: numeric_attribute,
+                  numeric_attribute: { key: http.status_code, min_value: 200, max_value: 399 }
+                },
+                {
+                  name: basic-sampling,
```
inian: should the name be changed here too?

PR #503, opened 2024-06-19
fix: skip Mime type check for .emptyFolderPlaceholder

Meyazhagan:

Bug fix

when MIME type is enabled, Folders can't be created.

## What is the new behavior?

for .emptyFolderPlaceholder files, will skip MIME type check

## Additional context

closes: https://github.com/supabase/supabase/issues/26359


fenos:
Thanks a lot for the PR.
I thought it was quicker to make the tidy-up changes for this to merge promptly.

I have added the constraint that the emptyFolder must be 0kb, and a specific function to determine if it is an emptyFolder

PR #508, opened 2024-06-25
Chore/improvements

fenos:

Bug fix

## What is the new behavior?

- Automatically push mirrors after release
- Refactored and tidied up SignatureV4 verify method
- Webhook event is now using a connection pool
- always bind the metrics route on the admin app


inian (approved):
Assuming the refactor commit was done in an automated way, other ones lgtm

PR #509, opened 2024-06-26
Allow creating folders even if the mime type validation is set

fenos:

Bug fix

When the mime-type is set, creating empty folders fails

## What is the new behavior?

Empty folders can still be created

Closes #503 this PR also restricts `emptyFolders` to 0kb

### DIFF
--- a/src/storage/limits.ts
+++ b/src/storage/limits.ts
@@ -115,3 +115,7 @@ export function parseFileSizeToBytes(valueWithUnit: string) {
 export function isUuid(value: string) {
   return /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(value)
 }
+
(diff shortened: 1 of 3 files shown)


pcnc (approved):

src/storage/limits.ts
```diff
@@ -115,3 +115,7 @@ export function parseFileSizeToBytes(valueWithUnit: string) {
 export function isUuid(value: string) {
   return /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(value)
 }
+
+export function isEmptyFolder(object: string) {
+  return object.endsWith('.emptyFolderPlaceholder')
```
pcnc, line 120: I reckon this gets removed whenever an object is uploaded in the folder

fenos: not always, is it a client side logic.
However, this PR allow creating this special file even if mime-type are defined

PR #520, opened 2024-07-14
feat: prevent passing in `role` claim in presigned URL JWTs

hf:
Just adds defense-in-depth that Storage can't be used as a signing oracle (an all-knowing entity you can tell to "sign" things for you) that would allow bypassing PostgREST RLS.

It achieves this by disallowing any `role` metadata to ever be supplied to the `signObjectUrl` method.


staaldraad (approved):
LGTM 🔒

PR #525, opened 2024-07-23
feat: remove crypto-js in favour of crypto native dependency

fenos:

Performance

Currently, storage uses crypto-js which run on the same JS thread and potentially block the event loop for longer than necessary.

## What is the new behavior?

Removed crypto-js in favour of the native crypto module which will be much more efficient

## Additional context


staaldraad (approved):

src/internal/auth/crypto.ts
```diff
@@ -1,21 +1,53 @@
...
 
 /**
- * Decrypts a text with the configured encryption key via ENCRYPTION_KEY env
- * @param ciphertext
- */
+ * Generate CryptoJs.AES key from passphrase
+ * https://github.com/brix/crypto-js/issues/468
+ * */
+function convertPassphraseToAesKeyBuffer(key: string, salt: Buffer): Buffer {
+  const password = Buffer.concat([Buffer.from(key, 'binary'), salt])
```
ferhatelmas: If key has a special character (unlikely but possible), isn't better to use default `utf8` here?

ferhatelmas: updated and added tests

PR #542, opened 2024-09-05
feat: handle authentication on top level routes

fenos:

Feature

Currently, we use `/public` and `/authenticated` endpoints to determine the access control

## What is the new behavior?

This PR modifies the route without the `/public` or `/authenticated` endpoint, which allows authentication to happen at the route level.

This fixes many issues on downloading objects or fetching info without knowing their privacy control


inian (approved):

criskell:
This PR doesn't seem to be working in production @fenos 
Without prefix:
image
With prefix:
image
Maybe related with this issue in storage-js:
https://github.com/supabase/storage-js/issues/209

Issue #133, opened 2022-04-08
Storage upload file bug


jet10000:

thebengeu:
Thanks for the bug report @jet10000! We'll take a look.

rahul3v:
@thebengeu @jet10000 
As per now for both `objectName` and  `bucketName` ,  supabase only allow s3 safe characters as per AWS guideline
```
 // only allow s3 safe characters and characters which require special handling for now
 // https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
```

alaister:
S3 supports UTF-8 characters in filenames. However, at the moment, we are very strict with which filenames we allow.
I think this is a valid use case to add support for different languages in filenames.

One option is to update the `isValidKey` function https://github.com/supabase/storage-api/blob/9480891af024396c58045578d16c91778aae67d2/src/utils/index.ts#L76-L80 to allow everything aside from the characters outlined in https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-guidelines-avoid-characters

PiddePannkauga:
I would love support for å,ä,ö.

Love from Sweden

javierfern03:
I would love support for  ´  I can't use words with accents in my images, which is very common in Spanish.

turulix:
Also ÖÄÜ in german

li4man0v:
MacOS generates screenshot names that don't match the pattern in `isValidKey`. For example, "Screenshot 2024-01-24 at 12.25.39 AM".

jingsam:
Ops! This issue seems easy to fix but have last for 2 year's. Unbelievable!

ThaddeusJiang:
hi everyone

My solution is base64 encoding when uploading file, demo: https://github.com/ThaddeusJiang/supabase-helpers/blob/main/backup_storage_buckets.ts#L71-L73

gitnik:
This is a big pretty big oversight. At the very least it would be nice if this was documented somewhere

stefan-girlich:
You can upload files using any name (e.g. base64, thx @ThaddeusJiang ), store the original file name in a different table, and then use it when generating the public link:

```ts
const {
  data: { publicUrl },
} = supabase.storage
  .from('my_bucket')
  .getPublicUrl(internalFileName, { download: originalFileName })

Source: https://supabase.com/docs/guides/storage/serving/downloads#downloading
```

jingsam:
[...]

PR #555, opened 2024-09-24
Bugfix/#133 special characters in filename

major-winter:

Bug fix

This fix addresses the issue described in [this bug item](https://github.com/supabase/storage/issues/133)

## What is the new behavior?

Allowing special characters to be used as object key


major-winter:
It seems I messed up the PR by including changes of others. Does anyone can help me resolve this or do I need to cancel this PR and create a new one instead?

PR #578, opened 2024-10-31
feat: connection pool

fenos:

Feature

## What is the new behavior?

This PR changes how connections to the database are made in multi-tenant.
Prior to this PR, Storage was using 1 connection per request and tearing down once not needed.

We now have an internal pool per tenant per instance which keeps the connections with an idle_timeout settings for clean up.


itslenny:
Using #679 instead

Issue #543, opened 2024-09-07
Storage API behind proxy returns invalid TUS resumable upload URL


Obeyed:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Using the default [docker supabase example](https://github.com/supabase/supabase/tree/master/docker); try to perform a resumable upload with [Storage API v1.10.0 or newer](https://github.com/supabase/storage/releases/tag/v1.10.0). This will return an invalid URL, for example: `http://localhost:8000/storage/v1:8000/upload/resumable/xxxxx`, where the port is defined twice.

## Expected behavior

A valid URL for the resumable upload.

## Additional context

What is the expected way to configure the `kong.yml` file for the storage service?

PR #585, opened 2024-11-14
fix: check for proxied host before appending port

Obeyed:

Bug fix

Fixes https://github.com/supabase/storage/issues/543

## What is the new behavior?

Avoids invalid URLs such as `https://gateway.example.com/storage/v1:8000/upload/resumable/bWxxx` where the `":8000"` is inserted incorrectly.

## Additional context

https://github.com/supabase/storage/pull/545 had partially fixed this

### DIFF
--- a/src/http/routes/tus/lifecycle.ts
+++ b/src/http/routes/tus/lifecycle.ts
@@ -106,7 +106,7 @@ export function generateUrl(
     const port = req.headers['x-forwarded-port']
 
     if (typeof port === 'string' && port && !['443', '80'].includes(port)) {
-      if (!host.includes(':')) {
+      if (!host.includes(':') && !host.includes('/')) {
         host += `:${req.headers['x-forwarded-port']}`
       } else {
         host = host.replace(/:\d+$/, `:${req.headers['x-forwarded-port']}`)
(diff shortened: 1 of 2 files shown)


fenos:
Hey @Obeyed usually the `host` header doesn't include any uri any specific reason why this change is needed?

Obeyed:
You're right / that doesn't make much sense. It actually seems to be some left over "fix" on my end (seemingly no longer relevant). I don't remember where I got it from, but from some old github issue here in the repo. Anyway, thanks for pointing me in the right direction. Closing this.



**_Just for future reference, hopefully this helps others. These are the changes I made to make it all work as expected:_**

_I added the `KONG_PORT_MAPS=443:8000` environment variable in the docker-compose file for the kong service to help Kong set the right values for `X-Forwarded-Port` header, (see https://docs.konghq.com/gateway/latest/reference/configuration/#port_maps). Without the port_maps the forwarded port is set to the kong service's docker container port instead, which is not right when kong is behind a proxy._

_I prepended the path `/storage/v1` to the `TUS_URL_PATH` env, ie. `TUS_URL_PATH: /storage/v1/upload/resumable`, and
I updated the `kong.yml` config to **not** append the path `/storage/v1` to the forwarded header on the storage service - it was like this (I don't remember why):_

```
    plugins:
      - name: cors
      - name: request-transformer
        config:
          add:
            headers:
              - "Forwarded: host=$(headers.host)/storage/v1;proto=https"
```

PR #605, opened 2024-12-31
fix: allow UTF-8 encoded object names

mlatief:

Bug fix 

Object names are restricted and Unicode non-ASCII characters are not allowed. See #133.

## What is the new behavior?

Allow any sequence of Unicode characters with UTF-8 encoding.

## Additional context


TimurBas:
Would like to see this being merged @fenos. 
Can't upload files that has Danish letters æ, ø, å.

PR #610, opened 2025-01-10
fix: updated fastify to version 5

ajitsinghkaler:
Fastify 4 is EOL at 2025-06-30. So updated fastify 4 to 5  to get to the latest version of fastify also to support version 5 node 20 is needed so updated node version to 20 too.

<br class="Apple-interchange-newline">

Currently the repo uses fastify 4. Which also supports node 14. 

## What is the new behavior?
The repo will be updated to fastify 5. Updated node version to 20.


fenos:
Thanks a lot for this PR
Since it is a major version I need a bit more time to test all the bits and pieces that might be affected

ajitsinghkaler:
@fenos is the review completed.

PR #613, opened 2025-01-12
feat: Add `mtime` as optional ETag algorithm in file backend alongside default `md5`

### LINKED ISSUE #612: File backend is very slow for Range requests to large files
# Bug report

## Describe the bug

In the local environment’s Storage (i.e., file backend), small Range requests to large files (e.g., a 1GB video) take several seconds, sometimes exceeding 10 seconds, to begin.

The delay is caused by the MD5 hash computation of the entire file for ETag generation.

In use cases that rely heavily on Range requests (e.g., [Self-host Maps with Protomaps and Supabase Storage](https://supabase.com/blog/self-host-maps-storage-protomaps) on your blog), the performance drops to unusable levels.

## To Reproduce

1. Set `STORAGE_BACKEND=file` in the `.env` file.
2. Upload a large file, such as a 1GB file, to the local Supabase Storage.
3. Generate a signed URL for the file.
4. Send a Range request using the following command:

    ```
    curl -X GET -I http://localhost:5000/object/sign/mybucket/large.bin?token=<token> -H 'Range: bytes=0-16384'
    ```

5. Observe a long delay before the response begins (e.g., a few seconds or more than 10 seconds).

## Expected behavior

The total file size should not significantly affect the response time for Range requests.

## System information

- OS: macOS
- Browser (if applies) Chrome and Safari
- Version of Node.js: 22.12.0

ciscorn:

Fix #612 

#612

In the file backend, ETags are generated by calculating the MD5 hash of the entire file contents, which can take several seconds or even more than 10 seconds for larger files. This becomes particularly problematic when utilizing Range requests.

## What is the new behavior?

The ETag in the file backend is now generated using only a file's modification time (`mtime`) and `size`. I believe this is sufficient as it follows the same [approach used by Nginx](https://github.com/nginx/nginx/blob/57d54fd922e7ecbebb78598d13adc9df1a4b69c0/src/http/ngx_http_core_module.c#L1701C23-L1706) (and moreover, the file backend is primarily used in local development).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -274,7 +274,7 @@ export function getConfig(options?: { reload?: boolean }): StorageConfigType {
       'STORAGE_FILE_BACKEND_PATH',
       'FILE_STORAGE_BACKEND_PATH'
     ),
-    storageFileEtagAlgorithm: getOptionalConfigFromEnv('STORAGE_FILE_ETAG_ALGORITHM') || 'mtime',
+    storageFileEtagAlgorithm: getOptionalConfigFromEnv('STORAGE_FILE_ETAG_ALGORITHM') || 'md5',
 
     // Storage - S3
     storageS3MaxSockets: parseInt(
--- a/src/config.ts
+++ b/src/config.ts
@@ -20,6 +20,7 @@ type StorageConfigType = {
   uploadFileSizeLimit: number
   uploadFileSizeLimitStandard?: number
   storageFilePath?: string
+  storageFileEtagAlgorithm: 'mtime' | 'md5'
   storageS3MaxSockets: number
   storageS3Bucket: string
   storageS3Endpoint?: string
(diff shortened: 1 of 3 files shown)


fenos:
Hi @ciscorn Awesome optimisation!
Could you please keep this backwards compatible and use an environment variable to control the checksum algorithm please

ciscorn:
@fenos Thanks!
I’ve modified the implementation to respect the environment variable. While I think “mtime” is a reasonable default, I’ve set “md5” as the default to maintain backward compatibility. What do you think?

src/config.ts
```diff
@@ -273,6 +274,7 @@ export function getConfig(options?: { reload?: boolean }): StorageConfigType {
       'STORAGE_FILE_BACKEND_PATH',
       'FILE_STORAGE_BACKEND_PATH'
     ),
+    storageFileEtagAlgorithm: getOptionalConfigFromEnv('STORAGE_FILE_ETAG_ALGORITHM') || 'mtime',
```
fenos: This should default to md5 for backwards compatibility

fenos (changes requested):

fenos:
Awesome job @ciscorn 
Just one little change requested, to keep the md5 as the default algorithm to not break existing projects.

In a major version we could consider switching to this as default

src/config.ts
ciscorn: Changed! thanks.

fenos (approved):

fenos:
Awesome job @ciscorn 🎉

Issue #267, opened 2022-03-29
Storage: header if-modified-since being ignored


ffabss:

fenos:
This should now be fixed

PR #623, opened 2025-01-22
fix: added cache headers to file backend

ajitsinghkaler:
Earlier cache headers were only working with S3 backend added them to file backend too.

On S3 backend respond with correctly if-none-match and if-modified-since headers added.

## What is the new behavior?
Added these changes to file backend too.

## Additional context


ajitsinghkaler:
Fixes #267

PR #626, opened 2025-01-24
feat: search objects v2

fenos:

Feature 

Super slow search functionality when there are millions of objects

## What is the new behavior?

Fast, reliable and performant search functionality across hundreds of millions of objects


migrations/tenant/0028-objects-level-index.sql
```diff
@@ -0,0 +1,3 @@
+-- postgres-migrations disable-transaction
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "objects_bucket_id_level_idx"
```
za-arthur: Do you not create an UNIQUE index here because you want to keep existing UNIQUE index?

migrations/tenant/0029-object-bucket-name-sorting.sql
```diff
@@ -0,0 +1,2 @@
+-- postgres-migrations disable-transaction
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS idx_name_bucket_unique on storage.objects (name COLLATE "C", bucket_id);
```
za-arthur: For the purpose of migration I think the index can be created without `COLLATE "C"`, because the `LIKE` operator is not used.
But it shouldn't affect anything in terms of performance or functionality, so it is up to you.

za-arthur (approved):
Thank you for the PR. In overall all looks good to me. I just left couple of comments below.

migrations/tenant/0029-object-bucket-name-sorting.sql
fenos: I think it matter that we use COLLATE "C" in this index, since we order the results by name COLLATE "C"

za-arthur (approved):
Thank you for the changes. LGTM.

PR #644, opened 2025-03-17
fix: improve migrations of prefixes

fenos:

Bug fix

- Make sure only 1 jobs for migration is ever run
- Batch migration is now using an algorithm to increase / decrease delays and batch size depending on performance

#### Batch Size algorithm
- Starting batch size 10k
- Have batch size increase by 5k every iteration
- Measure execution time of each batch if it is over 3s reduce back to 10k
- If 10k is still slow decrease batch size -1k for each iteration until 5k

#### Delay algorithm:
- Start with 1s delay
- Increase the delay to 1s for every iteration
- When reaches 30s reset back to 1s

### DIFF
--- a/migrations/tenant/0029-create-prefixes.sql
+++ b/migrations/tenant/0029-create-prefixes.sql
@@ -3,13 +3,19 @@
 -- We run this with 50k batch size to avoid long running transaction
 DO $$
     DECLARE
-        batch_size INTEGER := 50000;
+        batch_size INTEGER := 10000;
         total_scanned INTEGER := 0;
         row_returned INTEGER := 0;
         last_name TEXT COLLATE "C" := NULL;
         last_bucket_id TEXT COLLATE "C" := NULL;
+        delay INTEGER := 1;
+        start_time TIMESTAMPTZ;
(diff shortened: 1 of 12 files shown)


inian (approved):

migrations/tenant/0029-create-prefixes.sql
```diff
@@ -48,6 +61,24 @@ DO $$
                 EXIT;
             ELSE
                 COMMIT;
+                PERFORM pg_sleep(delay);
+                -- Increase delay by 1 second for each iteration until 30
+                -- then reset it back to 1
+                SELECT CASE WHEN delay >= 30 THEN 1 ELSE delay + 1 END INTO delay;
```
olirice: what's the rationale for increasing the delay by 1 second per iteration?
it looks like that'd lead to waves of different levels of iops

fenos: for projects with a low number of objects, ~100k it will finish relatively quicker with an increasing delay. as opposed to taking a lot more time ex with a fixed 30s per iteration, min ~15mins

PR #645, opened 2025-03-24
fix: optimize prefix triggers (#supabase34257)

catcoon33:

Bug fix, this fixes a "ghost" directory issue

checkout this issue:
[#supabase34257](https://github.com/supabase/supabase/issues/34257)

https://github.com/user-attachments/assets/3be5d5e7-04da-4d68-9ebd-273358a03766

When renaming a folder (by updating an object's path), the old prefix remains in the `storage.prefixes` table even after all objects have been moved out of it. This creates a "ghost" directory that:
- Still appears in directory listings
- Has no actual objects inside it
- Creates confusion for users
- Pollutes the storage hierarchy

## What is the new behavior?

- Added a cleanup of old prefixes when renaming objects

## Additional context

https://github.com/user-attachments/assets/986b4a2a-d707-4ff5-9490-377716444625

### DIFF
--- a/migrations/tenant/0026-objects-prefixes.sql
+++ b/migrations/tenant/0026-objects-prefixes.sql
@@ -129,6 +129,38 @@ BEGIN
 END;
 $func$ LANGUAGE plpgsql VOLATILE;
 
+CREATE OR REPLACE FUNCTION "storage"."objects_update_prefix_trigger"()
+    RETURNS trigger
+AS $func$
+DECLARE
+    old_prefixes TEXT[];
+BEGIN
+    -- Ensure this is an update operation and the name has changed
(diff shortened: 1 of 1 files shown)


fenos:
Hey @catcoon33. Thanks a lot for this; it is an elegant solution to the problem!
I'll include this fix in another PR because we cannot change existing migrations.

I offer a massive thanks for this!

fenos:
For reference, i've improved this further, and optimised the deletes with a CTE 
https://github.com/supabase/storage/pull/644/files#diff-e757d9e5b3619862081b16a587fc4974659d078314f5faf9992c7307aad021bfR103-R138

PR #649, opened 2025-04-01
feat: Pre-Signed URL Signatures with Storage JWT Secret

itslenny:

Feature

URLs can only be signed using the tenant jwt secret associated with the project 

## What is the new behavior?

- Use a new tenant specific "url signing jwk" to sign urls
- Generate url signing jwk for all existing tenants using the worker queue
- Create url signing jwk for new tenants 
- Allow user-submitted jwks (including asymmetric) to be used for self-signing of URLs
- Legacy JWT Secret is still used as fallback validation to ensure previously created long lived URLs are still valid


src/internal/auth/jwt.ts
```diff
@@ -154,19 +157,38 @@ export function verifyJWT<T>(
...
   })
 }
+
+/**
+ * Generate a new random HS256 JWK that can be used for signing JWTs
+ */
+export function generateHS256JWK(): JwksConfigKeyOct {
+  // Generate a 256-bit (32-byte) random key
+  // Convert the secret key to Base64URL encoding (JWK standard)
+  const k = crypto.randomBytes(32).toString('base64url')
```
hf: 32 should be enough, but generally speaking it should be 64 bytes -- you decide if the storage size is worth the tradeoff

src/http/routes/admin/jwks.ts
```diff
@@ -0,0 +1,146 @@
...
+    '/:tenantId/jwks',
+    { schema: addSchema },
+    async (request, reply) => {
+      const {
+        params: { tenantId },
+        body: { jwk, kind },
+      } = request
+
+      if (kind.includes('_')) {
+        return reply.status(400).send('Kind cannot contain underscore characters')
```
fenos: let's return json as errors to keep it consistent

fenos: let's extract the validation in it's own function

src/http/routes/admin/tenants.ts
```diff
@@ -364,6 +352,7 @@ export default async function routes(fastify: FastifyInstance) {
     }
 
     await multitenantKnex('tenants').insert(tenantInfo).onConflict('id').merge()
+    await generateUrlSigningJwk(tenantId)
```
fenos: should this be in the same transaction as the create tenant?

src/internal/database/url-signing-jwk-generator.ts
```diff
@@ -0,0 +1,76 @@
+import { logger, logSchema } from '../monitoring'
```
fenos: i would move this file in `internal/auth/generators/jwk-generator.ts`

src/internal/database/url-signing-jwk-generator.ts
```diff
@@ -0,0 +1,76 @@
...
+  running: boolean
+  sent: number
+}
+
+export class UrlSigningJwkGenerator {
+  private static signal: AbortSignal
+  private static isRunning: boolean = false
+  private static countSent: number = 0
+
+  static Init(signal: AbortSignal) {
```
fenos: this method it's not doing initialisation, maybe we can rename it to something like:

```
setAbortSignal()
```

src/start/server.ts
```diff
@@ -74,6 +75,9 @@ async function main() {
     startAsyncMigrations(shutdownSignal.nextGroup.signal)
   }
 
+  // Attach abort signal to jwk generation runner
+  UrlSigningJwkGenerator.Init(shutdownSignal.nextGroup.signal)
```
fenos: I don't think this is necessary because if we invoke the method that creates jobs in an HTTP handler, if the server closes, the abort signal will also be aborted

src/internal/database/url-signing-jwk-generator.ts
fenos: this shouldn't be needed, it's better to pass the signal in the method that can be aborted

src/storage/events/jwks-create-signing-secret.ts
```diff
@@ -0,0 +1,60 @@
...
+      retryDelay: 5,
+      priority: 10,
+    }
+  }
+
+  static async handle(job: Job<JwksCreateSigningSecretPayload>) {
+    const { tenantId } = job.data
+
+    try {
+      const { kid } = await generateUrlSigningJwk(tenantId)
```
fenos: this method doesn't seem to be idempotent, what happens if this job is called twice?

src/storage/object.ts
```diff
@@ -721,10 +721,10 @@ export class ObjectStorage {
       isUpsert: options?.upsert ?? false,
     })
 
-    const { secret: jwtSecret } = await getJwtSecret(this.db.tenantId)
+    const { urlSigningKey } = await getJwtSecret(this.db.tenantId)
```
fenos, line 724: is the `urlSigningKey` always available? even for projects that do not have the jwks generated?
If not, shall we keep this backwards compatible

src/internal/database/tenant.ts
```diff
@@ -205,13 +206,22 @@ export async function getServiceKey(tenantId: string): Promise<string> {
  */
 export async function getJwtSecret(
   tenantId: string
-): Promise<{ secret: string; jwks: TenantConfig['jwks'] | null }> {
```
fenos, line 215: The `tenant.ts` file is getting a bit out hands 😄 
We might need to consider to start refactoring this. We can do it either in this PR (at least for the new part) or after we merge It.

The part to refactor are: 

- S3 keys manager 
- JWKS generator manager


We probably need to use a simple pattern with 2 classes:

```ts
interface JWKSManagerStore {
  create()
  delete()
  ....etc
}
class JWKSManager {
  constructor(storage: JWKSManagerStore) {}
}

class JWKSManagerStoreKnex {
   constructor(knex: Knex)
}
```
 
a similar pattern for S3Credentials:

```ts
interface S3CredentialStore {
   create()
   delete()
   ...etc
}

class S3CredentialManager {
   constructor(store: S3CredentialStore) 
}
```

fenos (changes requested):
Overall looks great!
Just left some questions / small improvements to consider improving

src/storage/object.ts
itslenny: Yes, it is always available. It is always what should be used for signing URLs. This keeps the logic for which key in one place. If there is no JWK for url signing it will be the old jwt secret string, if there is an active jwk it will be that jwk.

src/internal/database/tenant.ts
itslenny: Sounds good. I was planning to fix that little cache bug I found in the S3 Credentials and add test coverage for it as there is none currently. I think it makes sense to do this refactor as part of that PR.

src/storage/events/jwks-create-signing-secret.ts
itslenny: Good call. Updated.

src/http/routes/admin/tenants.ts
itslenny: Makes sense. Updated here and in POST.

src/internal/auth/jwt.ts
itslenny: Updated to 64. I don't think the storage size is a concern. Better security is more better.

src/http/routes/admin/jwks.ts
```diff
@@ -0,0 +1,175 @@
...
+    if (running) {
+      return reply
+        .status(400)
+        .send(`Generate missing jwks is already running, and has sent ${sent} items so far`)
+    }
+
+    // TODO: get this abort signal somewhere or remove entirely
+    const { signal } = new AbortController()
+
+    UrlSigningJwkGenerator.generateUrlSigningJwksOnAllTenants(signal)
```
itslenny: @fenos I'm now passing the abort signal through when calling. Where can I get an appropriate signal to use for this and/or do we even need an abort on this?

src/storage/events/jwks-create-signing-secret.ts
```diff
@@ -0,0 +1,60 @@
...
+import { generateUrlSigningJwk } from '@internal/database'
+
+interface JwksCreateSigningSecretPayload extends BasePayload {
+  tenantId: string
+}
+
+export class JwksCreateSigningSecret extends BaseEvent<JwksCreateSigningSecretPayload> {
+  static queueName = 'tenants-jwks-create'
+
+  static getWorkerOptions(): WorkOptions {
```
itslenny, line 14: @fenos are these settings ok for this worker/queue? I copied these values (both `getWorkerOptions` and `getQueueOptions`) from `RunMigrationsOnTenants` let me know if you think they should be adjusted at all

src/http/routes/admin/jwks.ts
itslenny: Switched to req.signals.disconnect.signal

src/internal/database/tenant.ts
```diff
@@ -250,11 +262,169 @@ export async function listenForTenantUpdate(pubSub: PubSubAdapter): Promise<void
...
+  signal: AbortSignal,
+  batchSize = 200
+): AsyncGenerator<string[]> {
+  let lastCursor = 0
+
+  while (!signal.aborted) {
+    const data = await multitenantKnex('tenants')
+      .select('id', 'cursor_id')
+      .where('cursor_id', '>', lastCursor)
+      .whereNotExists(function () {
```
fenos: You don't need `whereNotExists` you can simply chain the conditions on the main query:

```ts
.andWhere((b) => {
 b.andWhere('tenants_jwks.kind', '!=' JWK_KIND_STORAGE_URL_SIGNING)
 b.andWhere('tenants_jwks.active', '!=', true)
})


```

src/http/routes/admin/jwks.ts
```diff
@@ -0,0 +1,172 @@
...
+
+interface JwksUpdateRequestInterface extends RequestGenericInterface {
+  Body: FromSchema<typeof updateSchema.body>
+  Params: {
+    tenantId: string
+    kid: string
+  }
+}
+
+function validateAddJwkRequest(
```
fenos: is cleaner if you just return an object with the error message and use res.status(400).reply() in the HTTP handler instead of passing the resp, done callback

fenos: ah, you are using this function as a `preHandler` of fastify!
Interesting, however, the validation logic cannot be re-used or tested easily

 Just calling the function directly in the handler could be easier to reason about:
 
 ```ts
app.post((res,rep) => {
   const { isValid, error } = validateJWKSRequest(req.body)
   if (!isValid) return res.reply(400).send(error)
})
```

alternatively if you still want to use a preHandler that logic above can be in the pre-handler instead

File: src/internal/database/tenant.ts [hunk 14]
```diff
@@ -250,11 +262,169 @@ export async function listenForTenantUpdate(pubSub: PubSubAdapter): Promise<void
...
+ * @param trx optional transaction to add the jwk within
+ */
+export async function generateUrlSigningJwk(
+  tenantId: string,
+  trx?: Knex.Transaction
+): Promise<{ kid: string }> {
+  const query = trx || multitenantKnex
+
+  return await query.transaction(async (transaction) => {
+    const activeJwk = await transaction('tenants_jwks')
```
fenos: as discussed, we need a unique index instead of find for update to guarantee uniqueness and idempotency

src/internal/database/tenant.ts
itslenny: I don't think this will work in this case. The goal of this query is to find `tenants` that do not have an active url signing jwk. This is using a subquery to achieve that check. The only alternative that would work would be to do a left join and check for null, but the sub-query seemed cleaner to me.

fenos (approved):

PR #650, opened 2025-04-02
feat: Purge cache for specific objects

fenos:

Feature

Currently is not possible to purge the cache programmatically

## What is the new behavior?

This PR introduces a new endpoint: `DELETE /cdn/{bucket}/{object}` that allows purging the cache for a specific object

## Additional context

The actual implementation for purging the cache sits outside of storage, this is because we don't want to couple the service 
to a specific purge mechanism.

Storage invokes an external HTTP endpoint, which does the purging to keep it agnostic


itslenny (approved):
Looks good to me. Maybe resolve those eslint warnings.

edit: I saw the warnings in "files changed" tab, but didn't realize this wasn't actually part of the change. Removing `(e: any)` if it's not going to be used should resolve both warnings.

PR #654, opened 2025-04-06
fix: update IP address literals detection routines

darora:
The simpler regexps used earlier supported one of numerous IPv6 literal formats. The updated implementation uses a more comprehensive library.


thebengeu (approved):

darora:
Failure is for coveralls upload - guessing it's an outage

darora:
Oops - mixed up the build errors between PRs, this one does actually have a test failure, will revert.

PR #656, opened 2025-04-07
feat: freeze migration name

fenos:

Feature

## What is the new behavior?

Allows to freeze storage to a certain migration even if new migrations are added

### DIFF
--- a/src/http/plugins/db.ts
+++ b/src/http/plugins/db.ts
@@ -25,7 +25,7 @@ declare module 'fastify' {
   }
 }
 
-const { dbMigrationStrategy, isMultitenant } = getConfig()
+const { dbMigrationStrategy, isMultitenant, dbMigrationFreezeAt } = getConfig()
 
 export const db = fastifyPlugin(
   async function db(fastify) {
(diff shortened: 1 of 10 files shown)


src/http/plugins/db.ts
```diff
@@ -209,6 +213,7 @@ export const migrations = fastifyPlugin(
         const tenant = await getTenantConfig(request.tenantId)
         const migrationsUpToDate = await areMigrationsUpToDate(request.tenantId)
 
+        console.log('MIGRATIONS UP TO DATE', migrationsUpToDate)
```
itslenny: Is this an log intentional or left over debugging?

src/internal/database/connection.ts
```diff
@@ -148,7 +148,7 @@ export class TenantConnection {
 
   async dispose() {
```
itslenny: [nit] Remove async if no await

itslenny (approved):

src/http/plugins/db.ts
fenos: left over 👍

PR #657, opened 2025-04-08
Use drop trigger if exists to support pg12

jingsam:

Bug fix

`create or replace trigger` only available up from PG14. When I connect to PG12, migration fails as https://github.com/supabase/storage/blob/37be2a349bb95be654da7c750e1dd1cf5812f55b/migrations/tenant/0026-objects-prefixes.sql#L148-L169

https://github.com/supabase/storage/blob/37be2a349bb95be654da7c750e1dd1cf5812f55b/migrations/tenant/0035-add-insert-trigger-prefixes.sql#L2-L8

contains unsupported syntax.

## What is the new behavior?

Use `drop trigger if exists / create trigger` instead of `create or replace trigger` to support pg12.

## Additional context

### DIFF
--- a/migrations/tenant/0026-objects-prefixes.sql
+++ b/migrations/tenant/0026-objects-prefixes.sql
@@ -146,24 +146,28 @@ END;
 $func$ LANGUAGE plpgsql VOLATILE;
 
 -- "storage"."prefixes"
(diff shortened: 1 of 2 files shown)


ferhatelmas:
As a guidance, we aim to support 13+ so will close this.

Is there any particular reason for not upgrading?

PR #668, opened 2025-04-10
fix: tenant s3 credentials fixes and refactor

itslenny:

Bug fix

Cache does not invalidate on database update/delete and s3 credential count always returns 0

## What is the new behavior?

Cache is invalidated on database update/delete, credential count returns correct value, credentials limited to 50, cleaner architecture, improved test coverage

## Changes

- migration to use correct cache key in pg_notify cache invalidation
- fix s3 credentials count, was always returning 0 (and allowing creation of > 50 credentials)
- check for valid uuid on delete credential endpoint (used to throw 500 on insert)
- refactor s3 credentials into manager pattern similar to jwks
- add test coverage
- updated add jwk endpoint to use 201 to match existing s3 credentials
- some easy eslint fixes that caught my attention


fenos (approved):

src/internal/database/s3-credentials-manager/manager.ts
```diff
@@ -0,0 +1,126 @@
+import crypto from 'node:crypto'
```
fenos: Can we move this to `storage/protocols/s3/credentials-manager.ts`

PR #672, opened 2025-04-21
fix: orphan object client detector

fenos:

Bug fix

## What is the new behavior?

Client-side tool to interact with the object scanner to list and/or detect orphan objects.

## Additional context

This tool has 2 commands: `list` and `delete`
In both operations, detected orphans' references are saved to disk.

Is recommended to run the list command before delete

### DIFF
--- a/src/scripts/orphan-client.ts
+++ b/src/scripts/orphan-client.ts
@@ -1,0 +1,185 @@
+import axios from 'axios'
+import { NdJsonTransform } from '@internal/streams/ndjson'
+import fs from 'fs'
+import path from 'path'
+
+const ADMIN_URL = process.env.ADMIN_URL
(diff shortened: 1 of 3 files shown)


src/scripts/orphan-client.ts
```diff
@@ -0,0 +1,175 @@
+import axios from 'axios'
+import { NdJsonTransform } from '@internal/streams/ndjson'
+import fs from 'fs'
+import path from 'path'
+
+const ADMIN_URL = ''
+const ADMIN_API_KEY = ''
```
itslenny: Can we pull these values from `process.env` instead? I'd just like to eliminate the chance of accidentally pushing a commit with secrets. Using node 20+ we can just use the `--env-file` flag so we don't need any 3rd party libraries to support this.

itslenny (approved):

PR #673, opened 2025-04-21
fix: support orioledb migrations by removing concurrent index creation

itslenny:

Bug fix

Unable to run database migrations on OrioleDB because it does not support concurrent index creation

## What is the new behavior?

Runs index creation non-concurrently when running on OrioleDB

## Additional context

- Detect OrioleDB and strip "concurrently" from migrations
- Add support for starting infrastructure using OrioleDB
- Run migrations against OrioleDB in GitHub actions to ensure we maintain support
  - Example of failure: https://github.com/supabase/storage/actions/runs/14577324884/job/40886209203
- Fix eslint warnings
- Remove version form docker compose files per this warning: `the attribute 'version' is obsolete, it will be ignored, please remove it to avoid potential confusion`


darora (approved):

src/internal/database/migrations/migration-transformer/disable-concurrent-index-transformer.ts
```diff
@@ -0,0 +1,25 @@
+import { Migration } from 'postgres-migrations/dist/types'
```
fenos: i would rename the folder `transformers` since we are already in the `migrations` folder

PR #678, opened 2025-04-23
fix: switch from jsonwebtoken to jose for jwt signing/verification

itslenny:

Bug fix / enhancement

We currently use jsonwebtoken for signing JWTs which is fully synchronous even when using the callback pattern

## What is the new behavior?

Switch to [jose](https://github.com/panva/jose) for jwt signing which uses libuv to sign jwts asynchronously. Also, adds support for OKP (ed25519/Ed448) jwks.


src/internal/auth/jwt.ts
```diff
@@ -145,39 +139,32 @@ export function verifyJWT<T>(
...
 
 /**
- * Generate a new random HS256 JWK that can be used for signing JWTs
+ * Generate a new random HS512 JWK that can be used for signing JWTs
  */
-export function generateHS256JWK(): JwksConfigKeyOCT {
-  // Generate a 64-byte random key, convert the secret key to Base64URL encoding (JWK standard)
-  const k = crypto.randomBytes(64).toString('base64url')
-  return { kty: 'oct', alg: 'HS256', k }
+export async function generateHS512JWK(): Promise<JwksConfigKeyOCT> {
```
fenos, line 167: is this a breaking change?

itslenny: No, it shouldn't be. This function is only used for generating the "signing jwk" for new tenants (and when we backfill). Existing customers using HS256 won't be modified (unless we roll their secret). Since it is a jwk the algorithm is encoded inside so it'll just work.

Previously we were using a 64 byte (512 bit) key with a HS256 jwk. Jose has an async generate secret function which takes an algo as input. To get it to produce a 64 byte key it has to be set to HS512.

fenos (approved):

PR #679, opened 2025-04-24
feat: rebalancing connection pooling

fenos:

Feature

Currently, storage opens 1 connection to the database for every request and tears it down once it ends

## What is the new behavior?

- Dedicated connection poolers for each tenant
- The pools are torn down after 5 seconds of inactivity
- starts with small pools and allows expanding the connection with additional pools to handle spiky traffic
- rebalances the pool when container autoscale
- Optional JWT caching

## Additional context


src/internal/cluster/discovery.ts
```diff
@@ -0,0 +1,53 @@
...
+      logger.info('[Cluster] Initial cluster size', {
+        type: 'cluster',
+        size: Cluster.size,
+      })
+
+      Cluster.watcher = setInterval(() => {
+        cluster
+          .getClusterSize()
+          .then((size) => {
+            if (size === Cluster.size) {
```
itslenny: This should be `!==`

src/internal/cluster/discovery.ts
```diff
@@ -0,0 +1,53 @@
...
+    clusterEvent.on(event, listener)
+  }
+
+  static async init(abortSignal: AbortSignal) {
+    if (process.env.CLUSTER_DISCOVERY === 'ECS') {
+      const cluster = new ClusterDiscoveryECS()
+      Cluster.size = await cluster.getClusterSize()
+
+      logger.info('[Cluster] Initial cluster size', {
+        type: 'cluster',
```
itslenny: Should `type` be ECS here? I feel like a cluster of type cluster doesn't tell us anything

src/start/server.ts
```diff
@@ -77,6 +78,19 @@ async function main() {
     startAsyncMigrations(shutdownSignal.nextGroup.signal)
   }
 
+  // Cluster information
+  await Cluster.init(shutdownSignal.nextGroup.signal)
+
+  Cluster.on('change', (data) => {
+    logger.info('[Cluster] Cluster size changed', {
+      type: 'cluster',
```
itslenny, line 86: same here... I think this should probably be `type: 'ECS'` and maybe get it from `data.clusterType` or something

src/internal/cluster/discovery.ts
fenos: the `type` field in the logs is the component involved, I agree we can add ECS in the metadata, ex: type: `cluster`, clusterType: `ECS` open for improvements

src/internal/database/tenant.ts
```diff
@@ -241,8 +251,43 @@ const TENANTS_UPDATE_CHANNEL = 'tenants_update'
...
-  await pubSub.subscribe(TENANTS_UPDATE_CHANNEL, (cacheKey) => {
+  await pubSub.subscribe(TENANTS_UPDATE_CHANNEL, async (cacheKey) => {
+    const oldConfig = tenantConfigCache.get(cacheKey)
     tenantConfigCache.delete(cacheKey)
+
+    if (!oldConfig) {
+      return
+    }
+
+    try {
```
itslenny: Might be nice to move this into a `onTenantConfigChange()` (or similar) function.

itslenny (approved):

src/internal/auth/jwt.ts
```diff
@@ -111,6 +113,54 @@ function getJWTAlgorithms(jwks: JwksConfig | null) {
   return algorithms
 }
 
+const jwtCache = new LRUCache<string, { token: string; payload: JWTPayload }>({
+  maxSize: 1204 * 1024 * 50, // 50MB
```
cemalkilic: Looks like a typo?

```suggestion
  maxSize: 1024 * 1024 * 50, // 50MB
```

PR #682, opened 2025-04-30
chore: update tests so they exit gracefully and compute more accurate coverage

itslenny:
- Ensure all tests release all handles
- Remove `--forceExit` flag since tests exit gracefully
- Update test order. Jwt test coverage wasn't getting reported correctly
- Remove unused test code (counting against our coverage %)
- Improve x-forward-host tests to ensure values are propagated to database layer
- Reduce size of S3 test payload to avoid timeouts
- Disallow / remove all floating promises
- Fix all ES Lint warnings in tests


src/http/routes/tus/index.ts
```diff
@@ -359,7 +359,7 @@ const publicRoutes = fastifyPlugin(
           },
         },
         (req, res) => {
-          options.tusServer.handle(req.raw, res.raw)
+          return options.tusServer.handle(req.raw, res.raw)
```
fenos: I'm not certain we need to return here, since the handle function uses the raw, req/resp

fenos (approved):

src/http/routes/tus/index.ts
fenos: From the documentation of Fastify, I discovered we might need to use `reply.hijack()` in these methods

https://fastify.dev/docs/latest/Reference/Reply/#hijack

itslenny: The return is because I banned "floating promises" in eslint, and `tusSever.handle()` returns a promise.

I tested this a bit, and you're right if we return `handle()`  this without `hijack()` it throws a `Reply was already sent` error, but the file still uploads successfully. Adding hijack resolves that error, BUT if there is an error inside of tus it never responds and just leaves the connection to timeout.

If we want to use `hijack()` we have to do something like this to get error responses

```typescript
res.hijack()
options.tusServer.handle(req.raw, res.raw).catch((e) => {
  res.raw.statusCode = 500
  res.raw.end(JSON.stringify(ERRORS.InternalError(e, 'TUS resumable upload error')))
})
```

The other option (what I wound up doing) that seems to work is just awaiting `handle()` but not returning it. So, the return winds up being `Promise<void>` with this the response is only sent once and there is a 500 if there is an error inside TUS.

To me this feels safer than hijack + catch, but I'm open to either.

PR #683, opened 2025-05-01
chore: improve folder structure

fenos:

chore

## What is the new behavior?

- moved `@internal/database/jwks-manager` -> `@internal/auth/jwks`
- moved `@internal/auth/generators/jwks-generator.ts` -> `@internal/auth/jwks/generator.ts`
- renamed `@internal/storage/protocols/s3/credentials-manager` -> `@internal/storage/protocols/s3/credentials`
- Added signelton function to create jwksManager and s3CredentialManager


src/internal/auth/jwks/instance.ts
```diff
@@ -0,0 +1,21 @@
+import { Knex } from 'knex'
+import { JWKSManagerStoreKnex } from './store-knex'
+import { JWKSManager } from './manager'
+
+let defaultJWKSManager: JWKSManager | undefined
+
+export function createDefaultJWKSManager(db: Knex): JWKSManager {
```
itslenny, line 7: Should this exit early if (somehow) `defaultJWKSManager` is already set?

I think maybe having separate `create` and `get` could be problematic. It's accessed via `create` in `tenant.ts`, but with `get` everywhere else. In the future, if someone copies the `create` from `tenant.ts` without realizing the difference it could cause issues.

Maybe we should just have `get` and make it create if not exists. Otherwise, maybe `create` should throw if it's called a second time to ensure we only create the default manager exactly once.

fenos: sounds good

fenos: The only issue of having only a getter is that we need to pass the multitenantKnex everywhere 😢 
maybe we can throw as you suggested

PR #685, opened 2025-05-03
fix: add bucket name length constraint

TylerHillery:

Bug fix, restricting bucket name length to a max of 63 characters to adhere to [S3 Bucket naming rules
](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html)

No restrictions on the length of bucket name

## What is the new behavior?

Added Postgres buckets name length check constraint and updated `isValidBucketName` function. Also added test in `bucket.test.ts` to verify `400` is being returned when issuing a `POST` request with name longer than 63 characters.

## Additional context

I updated the `fromDBerror` switch statement to include the new pgError.code for the buckets name length check constraint. The `ERRORS.InvalidBucketName` error requires the bucket name I opted to parse it from the `pgError.detail` Example `pgError.detail`:

`'Failing row contains (asdfasssssssssssssssssssssssssssssssssssssssssssssssssssssssssss, asdfasssssssssssssssssssssssssssssssssssssssssssssssssssssssssss, null, 2025-05-03 21:04:15.907656+00, 2025-05-03 21:04:15.907656+00, f, f, null, null, null).'`

### DIFF
--- a/src/storage/database/knex.ts
+++ b/src/storage/database/knex.ts
@@ -901,6 +901,13 @@ export class DBError extends StorageBackendError implements RenderableError {
           query,
           code: pgError.code,
         })
+      case '23514': {
+        const bucketName = pgError.detail ? pgError.detail.split(' ')[4]?.slice(0, -1) : 'unknown'
+        return ERRORS.InvalidBucketName(bucketName, pgError).withMetadata({
+          query,
+          code: pgError.code,
+        })
+      }
       default:
         return ERRORS.DatabaseError(pgError.message, pgError).withMetadata({
           query,
--- a/src/storage/limits.ts
+++ b/src/storage/limits.ts
@@ -59,9 +59,12 @@ export function isValidKey(key: string): boolean {
 export function isValidBucketName(bucketName: string): boolean {
   // only allow s3 safe characters and characters which require special handling for now
   // https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
+  // https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
   // excluding / for bucketName
   return (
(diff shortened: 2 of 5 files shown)


src/storage/database/knex.ts
```diff
@@ -901,6 +901,13 @@ export class DBError extends StorageBackendError implements RenderableError {
           query,
           code: pgError.code,
         })
+      case '23514': {
```
itslenny: Wouldn't this trigger for any database constraint? So, there is no way to be sure this is the bucket name error? Maybe this should be a more generic "violates constraint" error (or just let it fall through to DatabaseError) unless we can differentiate this one somehow.

src/storage/limits.ts
```diff
@@ -59,9 +59,12 @@ export function isValidKey(key: string): boolean {
 export function isValidBucketName(bucketName: string): boolean {
   // only allow s3 safe characters and characters which require special handling for now
   // https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
+  // https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
```
itslenny, line 64: Maybe add a tiny note here that we just take the length (and slash) restriction from bucket and the rest of the rules are based on S3 object validation.

itslenny (changes requested):

src/storage/database/knex.ts
TylerHillery: Good catch, I thought it was the error code was unique per check constraint but it was not. There is no way to distinguish any difference between check constraints so I think it makes most sense to let it fall through to DatabaseError

TylerHillery:
Updated PR to change the max bucket name length to 100 to stay backwards compatible

TylerHillery:
We can't add postgres constraint if there is data already in the table that violates the constraint. Putting this PR back in draft while I think of a way we can implement this while maintaining compatibility with buckets that already violate the constraint.

fenos (approved):

itslenny (approved):

PR #687, opened 2025-05-06
fix: support S3 pagination with path prefix and url encoding enabled

### LINKED ISSUE #686: S3 Pagination Broken for Large Folders in Supabase Storage (rclone & AWS CLI fail with NextMarker/ContinuationToken errors)
# Bug report

## Describe the bug

When listing a folder with a large number of subfolders in Supabase Storage using S3-compatible tools (rclone, AWS CLI), the listing fails with a pagination error. This prevents access to files in that folder.

- With rclone, the error is:
  ```
  s3 protocol error: received listing v1 with IsTruncated set, no NextMarker and no Contents
  ```
- With AWS CLI, the error is:
  ```
  Error during pagination: The same next token was received twice: {'ContinuationToken': '<token>'}
  ```

Other folders with fewer subfolders list fine. The issue appears to be related to the number of subfolders and the S3 pagination implementation.

## To Reproduce

1. Use a bucket/folder in Supabase Storage with a large number of subfolders (in my case, 9224 subfolders in `property-files`).
2. Attempt to list the folder using rclone:
   ```
   rclone lsd <redacted>:/storage/global/property-files
   ```
   Output:
   ```
   2025/05/06 15:10:33 ERROR : : error listing: s3 protocol error: received listing v1 with IsTruncated set, no NextMarker and no Contents
   2025/05/06 15:10:33 Failed to lsd with 2 errors: last error was: s3 protocol error: received listing v1 with IsTruncated set, no NextMarker and no Contents
   ```
3. Or attempt to list with AWS CLI:
   ```
   aws s3 ls s3://storage/global/property-files/ --profile <redacted> --endpoint-url <redacted>/storage/v1/s3 --page-size 10
   ```
   Output:
   ```
   PRE 0015298896902832bdd0a0d8dda50fbd7e9952fa1d9daaf43f4cd2fdb2bfe4ea/
   ... (more results)
   Error during pagination: The same next token was received twice: {'ContinuationToken': '<token>'}
   ```

4. Other folders (e.g., `profile-thumbnails`) list fine, even with subfolders.

## Expected behavior

I expect to be able to list all subfolders and files in a bucket/folder, regardless of the number of subfolders, using S3-compatible tools. Pagination should work as per the S3 API specification.

## Screenshots

N/A (see error messages above).

## System information

- OS: macOS (Darwin/24.4.0, arm64)
- rclone version: v1.67.0
- AWS CLI version: 2.27.8 (Python/3.13.3)
- Node.js version: v20.13.1
- Number of subfolders in affected folder: 9224
- Supabase S3 endpoint: <redacted>/storage/v1/s3

## Additional context

- The issue is blocking access to files in the affected folder.
- Other folders with fewer subfolders do not exhibit this issue.
- The problem occurs with both rclone and AWS CLI, suggesting a backend pagination bug.
- Supabase S3 compatibility docs: [https://supabase.com/docs/guides/storage/s3/compatibility#implemented-endpoints](https://supabase.com/docs/guides/storage/s3/compatibility#implemented-endpoints)
- [Discord discussion](https://discord.com/channels/839993398554656828/1369303534041825320)

itslenny:

Bug fix

Pagination fails when listing directories using S3 protocol with url encoding enabled because the continuation token contents was being url encoded.

## What is the new behavior?

Don't url encode the continuation token contents.

## Additional context

By default aws cli and rclone enable url encoding so this fixes pagination on both. This only impacts listing nested objects (objects with a `/` delimiter in their name)

Fixes #686

### DIFF
--- a/src/storage/object.ts
+++ b/src/storage/object.ts
@@ -550,7 +550,7 @@ export class ObjectStorage {
     const delimiter = options?.delimiter
 
     const cursor = options?.cursor ? decodeContinuationToken(options?.cursor) : undefined
-    const searchResult = await this.db.listObjectsV2(this.bucketId, {
+    let searchResult = await this.db.listObjectsV2(this.bucketId, {
       prefix: options?.prefix,
       delimiter: options?.delimiter,
       maxKeys: limit + 1,
@@ -558,7 +558,6 @@ export class ObjectStorage {
       startAfter: cursor || options?.startAfter,
     })
 
-    let results = searchResult
     let prevPrefix = ''
 
     if (delimiter) {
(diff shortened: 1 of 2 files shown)


jasperverbeet:
Wow! Thanks for opening this PR so quickly!

src/storage/object.ts
```diff
@@ -575,32 +574,36 @@ export class ObjectStorage {
...
+    if (searchResult.length > limit) {
+      searchResult = searchResult.slice(0, limit)
       isTruncated = true
     }
 
-    const folders = results.filter((obj) => obj.id === null)
-    const objects = results.filter((obj) => obj.id !== null)
+    const folders: Obj[] = []
+    const objects: Obj[] = []
+    searchResult.forEach((obj) => {
```
fenos, line 597: I'm wondering if it's better to not have a double loop to URL encode, and instead simply urlDecode the last item name, when encodingType === 'url'

itslenny: This single loop replaces the two filters (loops) that used to be here. We have to bucket them into folders/objects either way. This just uses that opportunity to do the url encoding.

fenos (approved):

src/storage/object.ts
fenos: Make sense! Thanks for clarifying

PR #696, opened 2025-05-22
feat: pgboss v10

fenos:

Feature

## What is the new behavior?

PgBoss update to V10

## Additional context


src/internal/database/pool.ts
```diff
@@ -217,6 +217,11 @@ class TenantPool implements PoolStrategy {
   }
 
   protected async drainPool(pool: Knex) {
+    if (!pool.client.pool) {
```
itslenny: Maybe this should use `pool?.client?.pool` the `if (pool)` below implies that pool could be undefined which would throw an error.

src/internal/queue/queue.ts
```diff
@@ -189,78 +197,146 @@ export abstract class Queue {
...
-          }
-          const res = await event.handle(job)
-
-          QueueJobCompleted.inc({
-            name: event.getQueueName(),
-          })
+    const queueName = event.getQueueName()
+    const deadLetterName = event.deadLetterQueueName()
+
+    const concurrentTaskCount = event.getWorkerOptions().batchSize || maxConcurrentTasks
```
itslenny: Do we want to allow batchSize to override maxConcurrentTasks?

If not maybe this should use `Math.min()` to constrain batch size <= maxConcurrentTasks

src/internal/queue/queue.ts
```diff
@@ -189,78 +197,146 @@ export abstract class Queue {
...
-                : await Queue.getInstance().getJobById(job.id)
+  protected static pollQueue(
+    event: SubclassOfBaseClass,
+    queueOpts: {
+      concurrentTaskCount: number
+      onMessage?: (job: Job) => void
+      signal?: AbortSignal
+    }
+  ) {
+    const pollingInterval = event.getWorkerOptions().pollingIntervalSeconds || 5 * 1000
```
itslenny: I think this should be `(event.getWorkerOptions().pollingIntervalSeconds || 5) * 1000` 

Without the parenthesis `pollingIntervalSeconds` isn't multiplied by 1000 it's just used as-is if it's truthy

src/start/worker.ts
```diff
@@ -36,6 +36,9 @@ main()
  */
 export async function main() {
   const { requestTraceHeader, adminPort, host } = getConfig()
+  mergeConfig({
+    pgQueueEnableWorkers: true,
+  })
```
itslenny: Doesn't this bypass config settings? Was this for debugging and left in by accident or intentional?

src/internal/database/pool.ts
fenos: Fixed thanks!

src/internal/queue/queue.ts
fenos: Fair point,
I've adjusted this by having:

- batchSize: controls how many records it pulls from the DB (default: concurrent-tasks + (concurrent-tasks % 20))
- concurrentTaskCount: controls how many parallel tasks are running (default: 50 from env)

src/internal/queue/queue.ts
fenos: Fixed thanks!

src/start/worker.ts
fenos: good catch, it was a debugging left-over

PR #707, opened 2025-06-09
feat: add pagination to getAllBuckets

TylerHillery:

Feature to add pagination to the GET buckets endpoint by introducing optional query parameters `limit`, `offset`, `sortColumn` and `sortOrder`

Currently, we do not allow pagination for GET buckets endpoint.

## What is the new behavior?

Now we allow optional query parameters to be passed in to allow pagination.

### DIFF
--- a/src/http/routes/bucket/getAllBuckets.ts
+++ b/src/http/routes/bucket/getAllBuckets.ts
@@ -28,9 +28,10 @@ const requestQuerySchema = {
   type: 'object',
   properties: {
     limit: { type: 'integer', minimum: 1, examples: [10] },
-    offset: { type: 'integer', minimum: 0, examples: [0] },
+    offset: { type: 'integer', minimum: 1, examples: [1] },
     sortColumn: { type: 'string', enum: ['id', 'name', 'created_at', 'updated_at'] },
     sortOrder: { type: 'string', enum: ['asc', 'desc'] },
+    search: { type: 'string', examples: ["my-bucket"] }
   },
 } as const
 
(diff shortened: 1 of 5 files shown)


src/http/routes/bucket/getAllBuckets.ts
```diff
@@ -28,9 +28,10 @@ const requestQuerySchema = {
   type: 'object',
   properties: {
     limit: { type: 'integer', minimum: 1, examples: [10] },
-    offset: { type: 'integer', minimum: 0, examples: [0] },
+    offset: { type: 'integer', minimum: 1, examples: [1] },
```
itslenny: I think offset min being 0 is fine. It's "useless" but a lot of time there will be code that sets a counter to 0 and increments it by limit per iteration and we shouldn't error if they pass in 0.

Also, I just checked listObjects and that is the behavior there so that would be consistent.

src/storage/database/knex.ts
```diff
@@ -288,15 +288,19 @@ export class StorageKnexDB implements Database {
     const data = await this.runQuery('ListBuckets', (knex) => {
       const query = knex.from<Bucket>('buckets').select(columns.split(','))
 
-      if (options?.sortColumn) {
+      if (options?.search !== undefined) {
+        query.where('name', 'like', `${options.search}%`)
```
itslenny: I think this should use `ilike` and probably `%${options.search}%` to search anywhere in the string without case sensitivity.

The current dashboard search behavior searches anywhere in the string, but is case sensitive. However, our object search uses ilike so I think we should stay consistent with that.

itslenny: You can also safely do `zz?.search?.length > 0`. This will resolve to false if search is undefined or an empty string. The like won't hurt anything if the string is empty, but it also doesn't add anything.

TylerHillery: Thanks for the feedback. I initially tried implementing it as `options?.search?.length > 0` as you suggested, but that resulted in a TypeScript error:
```
'options.search.length' is possibly 'undefined'.
```

`length` can't be guaranteed to not be undefined which was preventing the comparison to 0. Instead I added another condition after checking if search was not equal to undefined so it acts as a type guard.
```ts
      if (options?.search !== undefined && options.search.length > 0) {
        query.where('name', 'ilike', `%${options.search}%`)
      }
```
That way we don't count empty strings.

src/http/routes/bucket/getAllBuckets.ts
TylerHillery: fixed 👍

itslenny (approved):

PR #710, opened 2025-06-16
fix: backport grants on buckets & objects

soedirgo:
Re-home grants on `storage.buckets` & `storage.objects` from supabase/postgres migrations into Storage's own migrations


soedirgo:
Not sure why adding a migration caused the CI failure

PR #720, opened 2025-07-09
Iceberg Bucket support

fenos:

Feature

## What is the new behavior?

Implement a multi-tenant iceberg catalog proxy, currently supported S3Tables catalog

## Additional context


migrations/tenant/0038-iceberg-catalog-flag-on-buckets.sql
```diff
@@ -0,0 +1,66 @@
...
+        END IF;
+
+        ALTER TABLE storage.buckets DROP COLUMN IF EXISTS iceberg_catalog;
+        ALTER TABLE storage.buckets ADD COLUMN IF NOT EXISTS type BucketType NOT NULL default 'STANDARD';
+
+        CREATE TABLE IF NOT EXISTS storage.iceberg_buckets (
+            id text not null primary key,
+            created_at timestamptz NOT NULL default now(),
+            updated_at timestamptz NOT NULL default now(),
+            type BucketType NOT NULL default 'ICEBERG'
```
itslenny: Since we're doing a separate table for "buckets" and "iceberg_buckets" is the `type` column even needed anymore?

src/config.ts
```diff
@@ -489,6 +498,20 @@ export function getConfig(options?: { reload?: boolean }): StorageConfigType {
...
     ),
+
+    icebergWarehouse: getOptionalConfigFromEnv('ICEBERG_WAREHOUSE') || 'warehouse',
+    icebergCatalogUrl:
+      getOptionalConfigFromEnv('ICEBERG_CATALOG_URL') ||
+      `https://s3tables.ap-southeast-1.amazonaws.com/iceberg/v1`,
+
+    icebergCatalogAuthType: getOptionalConfigFromEnv('ICEBERG_CATALOG_AUTH_TYPE') || `sigv4`,
+    icebergCatalogToken: getOptionalConfigFromEnv('ICEBERG_CATALOG_AUTH_TOKEN'),
+    icebergMaxCatalogsCount: parseInt(getOptionalConfigFromEnv('ICEBERG_MAX_CATALOGS') || '2', 10),
```
itslenny, line 515: Could we `parseInt` these and represent them as numbers here like we do in `rateLimiterRedisConnectTimeout` etc?

src/http/routes/admin/tenants.ts
```diff
@@ -415,6 +453,15 @@ export default async function routes(fastify: FastifyInstance) {
       tenantInfo.tracing_mode = tracingMode
     }
 
+    if (features?.icebergCatalog?.enabled) {
+      tenantInfo.feature_iceberg_catalog = features?.icebergCatalog?.enabled
+      tenantInfo.feature_iceberg_catalog_max_namespaces = features?.icebergCatalog
+        ?.maxNamespaces as number
```
itslenny: Is this actually a string that we're just forcing into a number type? I think we should actually parse this into a number in config.ts and just represent it as a number throughout.

src/http/routes/iceberg/catalog.ts
```diff
@@ -0,0 +1,59 @@
...
+        throw ERRORS.FeatureNotEnabled('icebergCatalog', 'iceberg_catalog')
+      }
+
+      const bucket = await request.icebergCatalog.findCatalogById({
+        tenantId: request.tenantId,
+        id: request.query.warehouse,
+      })
+
+      if (!bucket) {
+        throw ERRORS.FeatureNotEnabled(request.query.warehouse, 'iceberg_catalog')
```
itslenny: Bucket not found error instead of FeatureNotEnabled

src/http/routes/s3/commands/head-object.ts
```diff
@@ -17,11 +18,35 @@ const HeadObjectInput = {
...
     (req, ctx) => {
       const s3Protocol = new S3ProtocolHandler(ctx.storage, ctx.tenantId, ctx.owner)
 
+      const icebergBucket = ctx.req.internalIcebergBucketName
+
+      if (!icebergBucket) {
+        throw ERRORS.InvalidParameter('Iceberg bucket not found in request context')
+      }
+
+      console.dir({
```
itslenny: remove... for debugging?

src/http/routes/s3/commands/upload-part.ts
```diff
@@ -36,6 +36,72 @@ const UploadPartInput = {
...
+          return {
+            headers: {
+              etag: part.ETag || '',
+              'Access-Control-Expose-Headers': 'etag',
+            },
+          }
+        })
+      }
+
+      console.dir('UPLOAD PARTTTT!!!!', {
```
itslenny: debug log

src/storage/database/knex.ts
```diff
@@ -284,9 +323,81 @@ export class StorageKnexDB implements Database {
...
+        })
+
+      const icebergBucketsAllowedColumnNames = ['id', 'type', 'created_at', 'updated_at']
+
+      const icebergBucketsColumns = columnNames.map((name) => {
+        if (name === 'name') {
+          return 'id as name'
+        }
+        if (!icebergBucketsAllowedColumnNames.includes(name)) {
+          return knex.raw(`null as ${name}`)
```
itslenny: injecting a passed-in value into a `knex.raw` feels potentially dangerous. Could we just filter anything not in `icebergBucketsAllowedColumnNames` out of the list entirely instead?

src/config.ts
fenos: They are already parsedInt, maybe i'm missing something obvious

src/internal/queue/event.ts
```diff
@@ -41,6 +42,7 @@ export class Event<T extends Omit<BasePayload, '$version'>> {
   public static readonly version: string = 'v1'
   protected static queueName = ''
   protected static allowSync = true
+  protected static preferSync = true
```
fenos: to remove

migrations/tenant/0038-iceberg-catalog-flag-on-buckets.sql
fenos: Yes, i agree. but it will make sense in the future when we refactor the schema

src/storage/protocols/iceberg/catalog/rest-catalog-client.ts
```diff
@@ -0,0 +1,714 @@
...
+   * @returns List of namespace identifiers
+   */
+  listNamespaces(params: ListNamespacesRequest) {
+    const warehouse = this.getWarehouse()
+    return this.httpClient
+      .get<ListNamespacesResponse>(`${warehouse}/namespaces`, { params })
+      .then((response) => response.data)
+      .catch((error) => {
+        if (error instanceof AxiosError) {
+          console.error('Error fetching configuration:', error.response?.data)
```
itslenny, line 423: Several console.error and one console.log in this file. These should probably use `logSchema`

PR #721, opened 2025-07-10
Feature: Delete empty folders from file system after file operations (#333)

### LINKED ISSUE #333: Delete empty folders from File System
# Delete empty folders from File System

## Feature

With Supabase Storage, when all files from one folder are deleted, the folder becomes no longer visible from the storage but it still exists in the File System. It is possible to implement a parameter which allows to delete the folder from the File System when all files of the folder are deleted?

EndlessGoodness:
#333 Delete empty folders from File System

Feature

Closes #333

When files are deleted from Supabase Storage using the file backend adapter, empty directories remain on the file system even though they are no longer visible in the storage UI. Over time, this leads to accumulation of empty folder structures that consume disk space and clutter the file system.

The issue occurs in several scenarios:
- Deleting individual files with `deleteObject()`
- Bulk deleting files with `deleteObjects()`
- Aborting multipart uploads with `abortMultipartUpload()`

## What is the new behavior?

This PR introduces automatic cleanup of empty directories when files are deleted. The implementation includes:

- **New `cleanupEmptyDirectories()` method**: Recursively removes empty parent directories up to the storage root path
- **Integrated cleanup in deletion operations**: 
  - `deleteObject()` now cleans up empty directories after single file deletion
  - `deleteObjects()` collects unique parent directories and cleans them up after bulk deletion
  - `abortMultipartUpload()` cleans up empty multipart directories
- **Safety measures**: 
  - Prevents cleanup beyond the configured storage root path
  - Graceful error handling that doesn't affect primary deletion operations
  - Cross-platform compatibility using `fs.remove()` for directory removal

After this change, when all files in a directory are deleted, the empty directory structure is automatically removed from the file system, keeping storage clean and preventing accumulation of unused folders.

## Additional context

This feature is specifically for the file backend adapter (`FileBackend` class). The cleanup is performed asynchronously and errors during cleanup are intentionally ignored to ensure that primary file operations are not affected by filesystem permission issues or race conditions.

The implementation follows the existing codebase patterns and maintains backward compatibility - existing functionality remains unchanged while adding the new cleanup behavior.

### DIFF
--- a/src/storage/backend/file.ts
+++ b/src/storage/backend/file.ts
@@ -214,6 +214,9 @@ export class FileBackend implements StorageBackendAdapter {
     try {
       const file = path.resolve(this.filePath, withOptionalVersion(`${bucket}/${key}`, version))
       await fs.remove(file)
+      
+      // Clean up empty parent directories
+      await this.cleanupEmptyDirectories(path.dirname(file))
     } catch (e) {
       if (e instanceof Error && 'code' in e) {
         if ((e as any).code === 'ENOENT') {
@@ -273,15 +276,31 @@ export class FileBackend implements StorageBackendAdapter {
       return fs.rm(path.resolve(this.filePath, bucket, prefix))
     })
     const results = await Promise.allSettled(promises)
-
-    results.forEach((result) => {
+    
+    // Collect unique parent directories for cleanup
+    const parentDirs = new Set<string>()
+    
+    results.forEach((result, index) => {
       if (result.status === 'rejected') {
         if (result.reason.code === 'ENOENT') {
           return
         }
         throw result.reason
+      } else {
+        // Add parent directory of successfully deleted file
+        const filePath = path.resolve(this.filePath, bucket, prefixes[index])
+        parentDirs.add(path.dirname(filePath))
       }
     })
+    
+    // Clean up empty directories
+    for (const dir of parentDirs) {
+      try {
+        await this.cleanupEmptyDirectories(dir)
+      } catch (e) {
+        // Ignore cleanup errors to not affect the main deletion operation
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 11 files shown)


EndlessGoodness:
Can you please review this pull request? @fenos

src/storage/backend/file.ts
```diff
@@ -542,6 +568,42 @@ export class FileBackend implements StorageBackendAdapter {
...
+      }
+
+      // Check if directory exists
+      const exists = await fs.pathExists(dirPath)
+      if (!exists) {
+        return
+      }
+
+      // Check if directory is empty
+      const contents = await fs.readdir(dirPath)
```
itslenny: I'm a little nervous about reading the entire directory contents to see if it is empty. If the directory contains thousands or millions (!!) of files this could potentially be a performance hit. Maybe better to introduce a helper function using `fs.opendir` to just read a single entry instead.



```Typescript
function isEmptyDir(path: string): Promise<boolean> {  
    try {
      const directory = await fs.opendir(path)
      const entry = await directory.read()
      await directory.close()

      return entry === null
    } catch (error) {
      return false
    }
}
```

EndlessGoodness: Yes, it can be a major problem if used on large data. I have updated my code as per your suggestions.

EndlessGoodness:
Can you please review the changes? @itslenny

itslenny:
Hi @EndlessGoodness thanks so much for this contribution. I pulled down your code and tested it by creating/deleting a large amount of files, and it seems to work really well. Great work.

I did however run into one issue. The value of storageFilePath can be relative or absolute. Your code works perfectly when it is absolute, but fails to delete anything if the path is relative.

To resolve this, and to create consistency in the file adapter in general, please add an update to your PR to ensure that `this.filePath` is always absolute. 

Replace line 55 (where this.filePath is set in the constructor) as follows

```TypeScript
    this.filePath = path.isAbsolute(storageFilePath)
      ? storageFilePath
      : path.resolve(__dirname, '..', '..', '..', storageFilePath)
```

After this change is made I will approve this PR.

itslenny (approved):

itslenny (changes requested):
Just ran our CI and it failed on a formatting error. Please run `npm format` to auto-fix the formatting.

Also, ensure that `npm run eslint:check` and `npm test` run without errors. Note both of these will likely throw some warnings which you can ignore.

EndlessGoodness:
Hello @itslenny I think the CI is running into an authentication issue in the workflow itself now, according to the error message. The error is : 

> Error: PGRST_JWT_SECRET is undefined

itslenny (approved):

itslenny:
@EndlessGoodness Sorry, there seems to be a misconfiguration in our test runner that causes tests to fail when being merged from a fork. I ran the tests locally to ensure they pass and merged your PR. Thanks again for submitting it.

PR #740, opened 2025-07-29
feat: upgrade tus to latest + S3 Locker

fenos:

Feature

## What is the new behavior?

- Upgrade TUS to latest
- Improve Postgres Lock to handle cancellation
- Implemented distributed lock with S3


src/http/routes/tus/index.ts
```diff
@@ -112,7 +139,7 @@ function createTusServer(
     respectForwardedHeaders: true,
     allowedHeaders: ['Authorization', 'X-Upsert', 'Upload-Expires', 'ApiKey', 'x-signature'],
     maxSize: async (rawReq, uploadId) => {
-      const req = rawReq as MultiPartRequest
+      const req = rawReq.node?.req as MultiPartRequest
```
itslenny, line 146: Maybe we should throw here if `rawReq.node` is undefined otherwise it'll just crash on the very next line trying to access `upload` on undefined `req`.

PR #749, opened 2025-08-14
feat: add support for sorting in list v2 endpoint

itslenny:

Feature

The list v2 endpoint does not support sorting like the old list endpoint

## What is the new behavior?

- Support sorting by name, created_at, and updated_at to parity the v1 list endpoint
- Fix pagination issue when objects and folders have similar names
- Add robust test coverage to ensure pagination and sorting work perfectly in all scenarios

### DIFF
--- a/migrations/tenant/0039-add-search-v2-sort-support.sql
+++ b/migrations/tenant/0039-add-search-v2-sort-support.sql
@@ -24,7 +24,6 @@ DECLARE
     cursor_op text;
     cursor_expr text;
     sort_expr text;
-    collate_clause text := '';
 BEGIN
     -- Validate sort_order
     sort_ord := lower(sort_order);


src/storage/object.ts
```diff
@@ -806,16 +819,21 @@ export class ObjectStorage {
   }
 }
 
-function encodeContinuationToken(name: string) {
-  return Buffer.from(`l:${name}`).toString('base64')
+function encodeContinuationToken(sortKey: string, value: string) {
+  return Buffer.from(`${sortKey}:${value}`).toString('base64')
```
itslenny: @fenos  did `l:` have some specific meaning/purpose? 

The thinking of encoding the sort column in the token is so we can give a meaningful error message if someone tries to change sort column and pass an old cursor. Without this check it will throw at the data base if switching between name and timestamp, or just give bad/unexpected results if switching between timestamps columns.

I could make it accept `l` for the name column if we want to make sure old cursors still work.

fenos: yes,
the reason is to allow for extending the cursor with additional params, while being backwards compatible. the name was in the `l:` probably the sort should be in the `s:`

src/storage/database/knex.ts
```diff
@@ -269,27 +269,44 @@ export class StorageKnexDB implements Database {
...
+      }
     }
   ) {
     return this.runQuery('ListObjectsV2', async (knex) => {
       if (!options?.delimiter) {
         const query = knex
           .table('objects')
           .where('bucket_id', bucketId)
-          .select(['id', 'name', 'metadata', 'updated_at'])
+          .select(['id', 'name', 'metadata', 'updated_at', 'created_at', 'last_accessed_at'])
```
fenos, line 284: last_accessed_at is deprecated, no need to include it

migrations/tenant/0039-add-search-v2-sort-support.sql
```diff
@@ -0,0 +1,127 @@
...
+                    AND level = $4
+                    AND %s
+                ORDER BY %s
+                LIMIT $3
+            )
+        ) obj
+        ORDER BY %s
+        LIMIT $3
+        $sql$,
+        cursor_expr_prefixes, -- prefixes WHERE
```
fenos, line 119: The cursor expression should be on the inner queries

src/storage/database/knex.ts
```diff
@@ -269,27 +269,55 @@ export class StorageKnexDB implements Database {
       nextToken?: string
       maxKeys?: number
       startAfter?: string
+      sortOrder?: string
```
fenos: How about adding the order in the `sortBy` object?

itslenny: Updated. I originally had it outside because it sorts by name even if sortBy is undefined so this allowed setting asc/desc without including the `sortBy` object, but having it all grouped together probably makes more sense.

PR #755, opened 2025-09-04
fix: use exactly once queue semantics for specific events

fenos:

Bug fix

Previously using `stately` queue, but has a nasty bug reported here: https://github.com/timgit/pg-boss/issues/535

## What is the new behavior?

Replaced `stately` queue with `exactly_once`


src/storage/events/pgboss/move-jobs.ts
```diff
@@ -0,0 +1,126 @@
...
+        if (job.data.deleteJobsFromOriginalQueue) {
+          const deleteSql = `
+                DELETE FROM ${schema}.job
+                WHERE name = '${fromQueueName}'
+                    AND state IN ('created', 'active', 'retry')
+            `
+          await multitenantKnex.raw(deleteSql)
+        }
+
+        await multitenantKnex.raw(sql)
```
itslenny: Duplicate of line 106.... Is this intentional? If so why are you running it twice?

PR #774, opened 2025-10-03
feat: vector buckets

fenos:

Feature

## What is the new behaviour?

Implement Vector Bucket data source

**Supported Operations:**

- CreateIndex
- DeleteIndex
- GetIndex
- ListIndexes
- PutVectors
- ListVectors
- ListVectorBuckets
- QueryVectors
- DeleteVectors
- GetVectorBucket
- GetVectors

**Authentication mechanisms:**

- SignV4
- JWT service_role

**Sharding**

- Implemented sharding for resources which span in multiple underlying buckets (vector indexes)

**Analytics Buckets**

- Fixed issue when committing an iceberg transaction when using big transaction numbers
- Added specific endpoints for managing analytics buckets


src/app.ts
```diff
@@ -63,6 +63,7 @@ const build = (opts: buildOpts = {}): FastifyInstance => {
   app.register(routes.cdn, { prefix: 'cdn' })
   app.register(routes.healthcheck, { prefix: 'health' })
   app.register(routes.iceberg, { prefix: 'iceberg/v1' })
+  app.register(routes.vectors, { prefix: 'vectors' })
```
itslenny: should this be singular ("vector")? to match: bucket, object, ...

src/http/error-handler.ts
```diff
@@ -55,6 +55,10 @@ export const setErrorHandler = (
         ? 500
         : 400
 
+      if (statusCode === 500) {
+        console.log('error')
```
itslenny: debug log, remove

src/http/plugins/signature-v4.ts
```diff
@@ -30,95 +37,176 @@ type AWSRequest = FastifyRequest<{ Querystring: { 'X-Amz-Credential'?: string }
...
+      skipIfJwtToken: false,
+      enforceJwtRoles: [],
+    }
+  ) {
+    // Use preParsing when allowing to pre-calculate the sha256 of the body
+    if (opts.allowBodyHash) {
+      fastify.addHook('preParsing', async (request: AWSRequest, reply, bodyPayload) => {
+        if (
+          opts.skipIfJwtToken &&
+          request.headers.authorization?.replace('Bearer ', '')?.match(JWT_SHAPE)
```
itslenny: Maybe this could be moved into a helper function in @internal/auth/jwt.ts

src/http/plugins/signature-v4.ts
```diff
@@ -30,95 +37,176 @@ type AWSRequest = FastifyRequest<{ Querystring: { 'X-Amz-Credential'?: string }
...
+    )
+  }
+
+  const wasBodyHashed = allowBodyHash && byteHasherStream && byteHasherStream.writableEnded
+
+  const returnStream = wasBodyHashed
+    ? byteHasherStream!.toReadable({ autoCleanup: true })
+    : (body as Readable)
+
+  const { secret: jwtSecret, jwks } = await getJwtSecret(request.tenantId)
```
itslenny, line 167: If this should use the same signing secret as URL signing (not sure if that's the intention or not) you should use `urlSigningKey` instead of `secret` which uses the JWK signing key if set otherwise falls back to jwtSecret inside `getJwtSecret()`.

src/http/plugins/signature-v4.ts
```diff
@@ -11,6 +11,13 @@ import {
   ChunkSignatureV4Parser,
   V4StreamingAlgorithm,
 } from '@storage/protocols/s3/signature-v4-stream'
+// @ts-expect-error - no types for compose
```
itslenny: Probably better to manually declare typedef for compose and remove this bypass.

Adding this down by the fastify declare (line 38) seems to resolve it and maintain stricter typing

```typescript
declare module 'stream' {
  export function compose<A extends Stream, B extends Stream>(s1: A, s2: B): B & A;
}
```

src/internal/errors/codes.ts
```diff
@@ -421,6 +436,41 @@ export const ERRORS = {
...
+      code: ErrorCode.S3VectorConflictException,
+      httpStatusCode: 409,
+      message: `${resource} "${name}" already exists`,
+    })
+  },
+  S3VectorNotFoundException(resource: string, name: string) {
+    return new StorageBackendError({
+      code: ErrorCode.S3VectorNotFoundException,
+      httpStatusCode: 404,
+      message: `resource "${name}" not found`,
```
itslenny, line 451: resource unused. Should it be `${resource}` ? Otherwise maybe just remove the param

src/internal/streams/hash-stream.ts
```diff
@@ -0,0 +1,261 @@
+// HashSpillWritable.ts
```
itslenny, line 1: remove?

src/internal/streams/hash-stream.ts
```diff
@@ -0,0 +1,261 @@
...
+      if (this.fileStream && !this.fileStream.closed) {
+        try {
+          this.fileStream.destroy()
+        } catch {}
+      }
+    })
+  }
+
+  // Writable implementation
+  _write(chunk: Buffer, _enc: BufferEncoding, cb: (error?: Error | null) => void): void {
```
itslenny, line 89: add private if this is meant to be private

src/internal/streams/hash-stream.ts
```diff
@@ -0,0 +1,261 @@
...
+        this.spillToDiskAndWrite(chunk).then(() => cb(), cb)
+      } else {
+        this.writeToFile(chunk).then(() => cb(), cb)
+      }
+    } catch (err) {
+      cb(err as Error)
+    }
+  }
+
+  _final(cb: (error?: Error | null) => void): void {
```
itslenny, line 112: add private if this is meant to be private

src/http/routes/vector/list-vectors.ts
```diff
@@ -0,0 +1,60 @@
...
+      },
+    },
+    async (request, response) => {
+      if (!request.s3Vector) {
+        throw ERRORS.FeatureNotEnabled('vectorStore', 'Vector service not configured')
+      }
+
+      const indexResult = await request.s3Vector.listVectors(request.body)
+
+      return response.send(indexResult)
```
itslenny, line 57: Might be nice to align this with the existing `listObjectsV2` naming for consistency.

- include `hasNext` boolean in response
- rename to `cursor` (request) and `nextCursor` (response)

src/http/routes/vector/list-indexes.ts
```diff
@@ -0,0 +1,51 @@
...
+  },
+  summary: 'List indexes in a vector bucket',
+} as const
+
+interface listIndexRequest extends AuthenticatedRequest {
+  Body: FromSchema<(typeof listIndex)['body']>
+}
+
+export default async function routes(fastify: FastifyInstance) {
+  fastify.post<listIndexRequest>(
```
itslenny, line 27: same as listVectors. Align pagination types with existing list V2

src/storage/protocols/vector/knex.ts
```diff
@@ -0,0 +1,251 @@
...
+          await wait(20 * Math.pow(2, attempt - 1))
+          lastError = error
+          continue
+        }
+
+        throw error
+      }
+    }
+
+    throw ERRORS.TransactionError('Transaction failed after maximum retries', lastError)
```
itslenny: Might be nice to make "after maximum retries" conditional on if there were retires, and/or maybe include the retry count in the error message to avoid confusion if we need to debug this in the future.

src/storage/protocols/vector/vector-store.ts
```diff
@@ -0,0 +1,335 @@
...
+    if (!command.vectorBucketName) {
+      throw ERRORS.MissingParameter('vectorBucketName')
+    }
+
+    const vectorBucket = await this.db.findVectorBucket(command.vectorBucketName)
+
+    return {
+      vectorBucket: {
+        vectorBucketName: vectorBucket.id,
+        creationTime: Math.floor(vectorBucket.created_at?.getTime() / 1000),
```
itslenny: check `vectorBucket.created_at` is typeof Date to avoid error/NaN

src/storage/protocols/vector/vector-store.ts
```diff
@@ -0,0 +1,335 @@
...
+    const bucketResult = await this.db.listBuckets({
+      maxResults: command.maxResults,
+      nextToken: command.nextToken,
+      prefix: command.prefix,
+    })
+
+    return {
+      vectorBuckets: bucketResult.vectorBuckets.map((bucket) => ({
+        vectorBucketName: bucket.id,
+        creationTime: Math.floor(bucket.created_at?.getTime() / 1000),
```
itslenny: check `vectorBucket.created_at` is typeof Date to avoid error/NaN

src/http/plugins/signature-v4.ts
fenos: No, is not the same signing secret as the URL signing

src/http/routes/vector/list-vectors.ts
fenos: Unfortunately, this vector List endpoint is a spec, which cannot customise:

https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_ListIndexes.html

src/storage/protocols/vector/knex.ts
fenos: Not sure i understood this, but in isolation mode: 'serialize' retrying the transaction is a common pattern
if the max retries are reached means that there are too many concurrent operations, which makes this transaction fail

src/storage/protocols/vector/vector-store.ts
fenos: it should always of type Date, but i've added an additional null check before Math.floor

src/storage/protocols/vector/knex.ts
itslenny: I just meant adding the count to the error message or only saying "after maximum retries" if there were retries as currently it says that even for requests that had no retries.

src/http/routes/iceberg/table.ts
```diff
@@ -234,18 +235,24 @@ const commitTransactionSchema = {
...
               type: 'string',
-              description: 'Name of the requirement (e.g. assert-ref-snapshot-id)',
-              examples: ['assert-ref-snapshot-id'],
+              description:
+                'Type of the requirement (e.g. assert-ref-snapshot-id, assert-table-uuid)',
+              examples: ['assert-ref-snapshot-id', 'assert-table-uuid'],
             },
+            // allow arbitrary additional args specific to the requirement
+            ref: { type: 'string' },
+            // 'snapshot-id': { type: 'number', format: 'int64', bigint: true },
```
itslenny, line 248: Will this be added/uncommented in the future? Otherwise remove this line.

src/http/routes/iceberg/table.ts
```diff
@@ -254,18 +261,54 @@ const commitTransactionSchema = {
...
+            snapshot: {
+              type: 'object',
+              properties: {
+                // 'snapshot-id': { type: 'string', format: 'int64', bigint: true },
+                // 'parent-snapshot-id': {
+                //   type: 'integer',
+                //   format: 'int64',
+                //   bigint: true,
+                //   nullable: true,
+                // },
```
itslenny, line 280: remove?

src/http/routes/iceberg/table.ts
```diff
@@ -424,27 +513,64 @@ export default async function routes(fastify: FastifyInstance) {
...
-    },
-    async (request, response) => {
-      if (!request.icebergCatalog) {
-        throw ERRORS.FeatureNotEnabled('icebergCatalog', 'iceberg_catalog')
+  fastify.register(async (fastify) => {
+    fastify.addContentTypeParser('application/json', {}, (_request, payload: unknown, done) => {
+      try {
+        if (typeof payload === 'string') return done(null, JSONBigint.parse(payload))
+        if (Buffer.isBuffer(payload)) return done(null, JSONBigint.parse(payload.toString('utf8')))
+        if (payload && typeof (payload as any).on === 'function') {
```
itslenny, line 521: This check is kinda icky... maybe replace with a type guard

```TypeScript
function isReadableStream(obj: unknown): obj is NodeJS.ReadableStream {
  return (
    obj != null &&
    typeof obj === 'object' &&
    typeof (obj as NodeJS.ReadableStream).on === 'function'
  )
}
```

Then you can do this and remove the `as NodeJS.ReadableStream`

```TypeScript
if (isReadableStream(payload)) {
  // ...
}
```

src/start/server.ts
```diff
@@ -47,12 +53,26 @@ main()
  * Start Storage API server
  */
 async function main() {
-  const { databaseURL, isMultitenant, pgQueueEnable, dbMigrationFreezeAt } = getConfig()
+  const { databaseURL, isMultitenant, pgQueueEnable, dbMigrationFreezeAt, vectorS3Buckets } =
+    getConfig()
+
+  // Sharding for special buckets (vectors, analytics)
+  const sharding = new ShardCatalog(new KnexShardStoreFactory(multitenantKnex))
```
itslenny, line 60: should this be inside the `isMultitenant` if block?

src/storage/protocols/iceberg/catalog/rest-catalog-client.ts
```diff
@@ -562,7 +585,13 @@ export class RestCatalogClient {
...
         },
       })
-      .then((response) => response.data)
+      .then((response) => {
+        // console.log({
+        //   url: response.request.,
+        //   headers: response.request.headers,
+        // })
+        return response.data
+      })
```
itslenny, line 594: revert debug code

src/storage/protocols/iceberg/catalog/rest-catalog-client.ts
```diff
@@ -696,14 +725,28 @@ export class SignV4Auth {
...
+    //       `
+    // curl -X ${req.method?.toUpperCase() || 'GET'} \\
+    // ${Array.from(signedReq.headers.entries())
+    //   .map(([name, value]) => `  -H '${name}: ${value}' \\`)
+    //   .join('\n')}
+    //   "${(req.baseURL || '') + req.url + (queryString ? `?${queryString}` : '')}"${
+    //         req.data ? ` \\\n  -d '${JSON.stringify(req.data)}'` : ''
+    //       }
+    // `.trim()
+    //     )
```
itslenny, line 747: remove debug code

PR #776, opened 2025-10-06
fix: add X-Robots-Tag header to prevent robot indexing and following

itslenny:

Bug fix

Many objects are indexed by web crawlers as there are no x-robot-tag headers

## What is the new behavior?

- Add default `x-robots-tag: none` header to prevent search engines from indexing / following
- Add ability to specify custom `x-robots-tag` header when uploading objects
- Use `x-robots-tag` value specified in object metadata in headers when serving files
- Fix eslint warnings in edited files

## Additional context

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Robots-Tag

### DIFF
--- a/src/storage/uploader.ts
+++ b/src/storage/uploader.ts
@@ -12,6 +12,7 @@ import { getConfig } from '../config'
 import { logger, logSchema } from '@internal/monitoring'
 import { Readable } from 'stream'
 import { StorageObjectLocator } from '@storage/locator'
+import { validateXRobotsTag } from './validators/x-robots-tag'
 
 const { storageS3Bucket, uploadFileSizeLimitStandard } = getConfig()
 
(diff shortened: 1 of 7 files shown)


src/storage/uploader.ts
```diff
@@ -112,6 +113,10 @@ export class Uploader {
         request.signal
       )
 
+      if (request.file.xRobotsTag) {
```
fenos, line 117: I think we should start adding strict validation for these headers.
Let's start adding the validation for the x-robot-tags

itslenny: I've added validation to where the values are stored via API and before applying the header to our responses. So, we won't serve a file with an invalid header in the case an invalid value was set in the DB directly (bypassing the API check).

src/storage/validators/x-robots-tag.ts
```diff
@@ -0,0 +1,188 @@
+import { ERRORS } from '@internal/errors'
+
+const SIMPLE_RULES = [
```
ferhatelmas, line 3: Ignore me if this is already discussed but otherwise wanted let you know that `nocache`/`noarchive` seem still in use by some (i.e. [Bing](https://www.bing.com/webmasters/help/which-robots-metatags-does-bing-support-5198d240)) and so excluding them can fail normally valid requests in validation

PR #783, opened 2025-10-17
fix: log connections that timeout, abort, or send unparsable data

itslenny:

Bug fix

There is no log entry created for connections that don't trigger a request. 

- connections that timeout (at the node:http level)
- connections that are aborted before a full http request is received
- connections that send data that can't be parsed

## What is the new behavior?

Create a log entry for all connections that don't result in a request event

### Disconnect before data sent

```bash
# close connection after 1 second
timeout 1 tail -f /dev/null | nc localhost 5000
```

Screenshot 2025-10-17 at 2 30 03 PM

### Node http connection timeout

```bash
# leave connection open triggers node:http timeout
tail -f /dev/null | nc localhost 5000
```

Screenshot 2025-10-17 at 2 33 01 PM

### Send unparsable data

```bash
printf "saldfijasdflijasdflisadfsldif" | nc localhost 5000
```

Screenshot 2025-10-17 at 2 30 17 PM

### Send incomplete data

```bash
printf "POST /object/mybucket/file.txt HTTP/1.1\r\nHost: localhost:5000\r\nUser-Agent: MyClient/1.0\r\nx-forwarded-host: bjhaohmqunupljrqypxz.local.com\r\nAuthorization: Bearer xxx\r\n" | nc localhost 5000
```

Screenshot 2025-10-17 at 2 30 38 PM

### Send complete headers with no payload

```bash
# Triggers ABORTED REQ (existing behavior)
printf "POST /object/mybucket/file.txt HTTP/1.1\r\nHost: localhost:5000\r\nUser-Agent: MyClient/1.0\r\nx-forwarded-host: bjhaohmqunupljrqypxz.local.com\r\nAuthorization: Bearer xxx\r\n\r\n" | nc localhost 5000
```

Screenshot 2025-10-17 at 2 30 57 PM


src/http/plugins/log-request.ts
```diff
@@ -22,13 +27,65 @@ declare module 'fastify' {
...
 /**
  * Request logger plugin
  * @param options
  */
 export const logRequest = (options: RequestLoggerOptions) =>
   fastifyPlugin(
     async (fastify) => {
+      // Watch for connections that timeout or disconnect before complete HTTP headers are received
+      // Log if socket closes before request is triggered
+      fastify.server.on('connection', (socket) => {
```
fenos: we would need to remove these event handlers at the end of the request otherwise we will create memory leaks

fenos: Also we need to consider that we use `keep-alive` so connections might be reused hence the listeners would be removed in the current implementation

src/http/plugins/log-request.ts
```diff
@@ -23,13 +26,79 @@ declare module 'fastify' {
...
+        // Track when data arrives for a potential request
+        const onData = () => {
+          // Reset tracking for each new request on keep-alive connections
+          if (!hasReceivedData || requestLogged) {
+            hasReceivedData = true
+            currentRequestStart = Date.now()
+            requestLogged = false
+          }
+        }
+        socket.on('data', onData)
```
ferhatelmas, line 70: cleanup is registered once and it's removed after first request but keep-alive on the same connection can revert requestLogged then new request wouldn't have the callback and would do extra log

src/http/plugins/log-request.ts
```diff
@@ -23,13 +26,79 @@ declare module 'fastify' {
...
+            const req = createPartialLogRequest(fastify, socket, currentRequestStart)
+
+            doRequestLog(req, {
+              excludeUrls: options.excludeUrls,
+              statusCode: 'ABORTED CONN',
+              responseTime: (Date.now() - req.startTime) / 1000,
+            })
+          }
+        })
+      }
```
ferhatelmas, line 93: I think this can fire before raw.once(close) on partial body then we can log twice again since callback wouldn't be running. We should set requestLogged inside as well to prevent  other log

ferhatelmas:
It would be really nice two have some test coverage that we can run automatically in CI 🙏🏻

PR #796, opened 2025-11-09
feat: add analytics bucket sharding

fenos:

Feature

Analytics buckets currently use a single undeline catalogue to store tables and namespaces.
Sometimes, Iceberg catalogues have limitations in the maximum number of iceberg tables they can accommodate.

## What is the new behavior?

We implement sharding for analytics buckets, meaning that we can shard iceberg tables across multiple catalogues and being able to multiply the number of available tables that we can create and offer. Thus, it becomes a federated catalogue.

## Additional context

sharding is optional, but needed when offering analytics buckets as multi-tenant service


migrations/multitenant/0022-iceberg-catalog-sharding.sql
```diff
@@ -0,0 +1,62 @@
...
+
+
+        FOREACH i_shard_key IN ARRAY iceberg_shards
+            LOOP
+                INSERT INTO shard (kind, shard_key, capacity) VALUES ('iceberg-table', i_shard_key, 1000)
+                ON CONFLICT (kind, shard_key) DO NOTHING;
+            END LOOP;
+
+
+        IF array_length(iceberg_shards, 1) = 0 THEN
```
itslenny: this check could probably be before the foreach

src/internal/sharding/strategy/single-shard.ts
```diff
@@ -9,6 +9,37 @@ export class SingleShard implements Sharder {
     }
   ) {}
 
+  listShardByKind(): Promise<ShardRow[]> {
+    return Promise.resolve([
+      {
+        id: 1,
+        kind: 'iceberg-table',
+        shard_key: this.singleShard.shardKey,
```
itslenny, line 17: I noticed below you changed `this.singleShard.shardKey` to `1` in a few places. Should this be the same? I'm assuming the value is 1 either way, but not sure why this isn't consistent.

src/storage/protocols/iceberg/catalog/reconciler.ts
```diff
@@ -0,0 +1,272 @@
...
+    const sharding = new ShardCatalog(new KnexShardStoreFactory(multitenantKnex))
+    const shards = await sharding.listShardByKind('iceberg-table')
+
+    await Promise.allSettled(
+      shards.map(async (shard) => {
+        const namespaces = this.listNamespaces(shard.shard_key)
+
+        for await (const nsBatch of namespaces) {
+          for (const namespace of nsBatch) {
+            const tables = this.listTables(namespace[0], shard.shard_key)
```
itslenny, line 48: is it possible that namespace is an empty array (or undefined)? If so we should check to be safe.

itslenny (approved):

src/storage/protocols/iceberg/catalog/reconciler.ts
fenos: not possible that the namespace is an empty array, but I can add that check in

src/storage/protocols/iceberg/catalog/tenant-catalog.ts
```diff
@@ -360,42 +434,33 @@ export class TenantAwareRestCatalog extends RestCatalogClient {
...
-        })
+      const namespaceCount = await store.countNamespaces({
+        tenantId: this.tenantId,
+        limit: this.options.limits.maxNamespaceCount + 1,
+      })
 
-        if (namespaceCount > this.options.limits.maxNamespaceCount) {
-          throw ERRORS.IcebergMaximumResourceLimit(this.options.limits.maxNamespaceCount)
-        }
+      if (namespaceCount > this.options.limits.maxNamespaceCount) {
```
ferhatelmas:
```suggestion
      if (namespaceCount >= this.options.limits.maxNamespaceCount) {
```

fenos: Good catch thanks,
Ignore: ~EDIT: actually, in the query above i'm adding +1 to the limit - so technically this condition is correct:~

PR #801, opened 2025-11-17
fix: remove duplicate slash in tus upload url generator

itslenny:

Bug fix

Paths returned by TUS during multi-part uploads include a duplicate slash when adding the forwarded path prefix

Example: `/storage/v1//upload/resumable/abc123`

## What is the new behavior?

Remove duplicate slash (`/storage/v1/upload/resumable/abc123`)

## Additional context

Thanks for the report @aantti

### DIFF
--- a/src/storage/protocols/s3/signature-v4.ts
+++ b/src/storage/protocols/s3/signature-v4.ts
@@ -424,8 +424,8 @@ export class SignatureV4 {
     signedHeaders: string[]
   ) {
     const method = request.method
-    const canonicalUri = new URL(`http://localhost:8080${request.prefix || ''}${request.url}`)
-      .pathname
+    const prefix = request.prefix ? request.prefix.replace(/\/+$/, '') : ''
+    const canonicalUri = new URL(`http://localhost:8080${prefix}${request.url}`).pathname
     const canonicalQueryString = this.constructCanonicalQueryString(request.query || {})
     const canonicalHeaders = this.constructCanonicalHeaders(request, signedHeaders)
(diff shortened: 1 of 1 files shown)


src/http/routes/tus/lifecycle.ts
```diff
@@ -126,7 +126,8 @@ export function generateUrl(
 
   const forwardedPath = req.headers['x-forwarded-prefix']
   if (requestAllowXForwardedPrefix && typeof forwardedPath === 'string') {
-    basePath = forwardedPath + path
+    // Remove trailing slash from forwardedPath to avoid double slashes
```
ferhatelmas, line 129: [Similar pattern could happen](https://github.com/supabase/storage/blob/4f66e94ee1bce185b6eb44b8ecf03b1853246909/src/http/plugins/signature-v4.ts#L116-L119) for [signV4 of canonical request](https://github.com/supabase/storage/blob/4f66e94ee1bce185b6eb44b8ecf03b1853246909/src/storage/protocols/s3/signature-v4.ts#L427-L428) if this feature `requestAllowXForwardedPrefix` is enabled or a configuration value ending with a slash is given. I believe, as a result, it can cause `SignatureDoesNotMatch` errors

aantti: Yes.

itslenny: Good catch. This (and the original issue) does not impact our hosted service because we do not have a trailing slash, but for self hosted people may want to configure it (nginx, envoy, etc) with a trailing slash and that would cause S3 signature mismatch.

PR #805, opened 2025-11-18
fix: make storage migrations idempotent

TylerHillery:

The `storage.migrations` table can sometimes fall out of sync, requiring migrations to be replayed on databases that already applied them. This PR makes all migrations fully idempotent, allowing you to clear `storage.migrations` and safely rerun the entire migration chain without errors, ensuring the final database state matches a clean first-run.

## Current behavior
- Several functions use `drop function if exists` without `cascade` causing issues when it has dependencies like RLS policies
- Some roles are created without checking for existence
- `migrations/tenant/0038-iceberg-catalog-flag-on-buckets.sql` creates an index on `bucket_id` without verifying that the column exists. This column gets renamed later on so it causes issues on the second run.
- iceberg catalog id fks only get created if the `catalog_id` column doesn't exist, so if you run the migration again after clearing out the migrations table it's going to drop the fk constraint but it wont recreate it because the column is still there.

## New behavior

- My initial logic for creating the functions in an idempotent way is tricky because `create or replace` is already idempotent but it doesn't work if the return type of the function changes like it does with `storage.get_size_by_bucket()` so this needs to be dropped first. Dropping first introduces a problem though if there something else dependent on that function. If there is we can't drop it without `cascade` but that means we would remove anything dependent on that function including RLS policies. In general I think we should avoid `cascade` where we can and rely on `create or replace` if we make updates to a function such as changing return type we might want to add a suffix with `_v2` instead

- Added existence checks when creating roles (same approach as super_user)

- Updated `0038-iceberg-catalog-flag-on-buckets.sql` to check for `bucket_id` before creating the index, since it’s renamed to `bucket_name` in migration `0048`

- Added step in CI to drop the migrations table and rerun migrations to ensure we don't generate any new migrations that are not idempotent. I run pg_dump after the initial migration, then delete the migrations table and rerun the migrations taking another pg_dump comparing the two. I ignore the `storage.migrations` table as the times the migrations took place are always going to differ.

- Separate the `alter table add column catalog_id` for both `iceberg_namespaces` and `iceberg_tables` so that the fk creation happens in a separate statement to ensure it gets created no matter how many times you run the migrations in a row by first checking if it doesn't exist.

- migration `0040` drop triggers if they exists before creating them

- New step in GitHub Action CI job to catch these ahead of time. It does the following:
  - will take a `pg_dump` of the database after running all migrations for the first time
  - Runs new script `test-migration-idempotency.ts` that runs all migrations one by one, resets them and reruns them. To do this I had to to make the following updates
    - Include `create-migrations-tables` to the `DBMigration` type generation. This was needed because the `resetMigration` requires you to pass in the previous migration if you want to clear out the current migration. Well for the first migration there is actually the `create-migration-tables` migration first but since the that wasn't in the `keyof typeof DBMigration` TypeScript was complaining. 
    - I also modified `resetMigration` to make `tenantId` optional like `runMigrationsOnTenant` does. I then updated the function to only update the tenant state if the tenant id was being passed in. There is no "tenants" table when running in single tenant mode so updating that state doesn't make sense.
    - Takes another `pg_dump` and compare the before and after to make sure the db ends up in the same state

## Additional context

Also to clarify `resetMigration` doesn't do a "down" migration, it simply clears `migrations` table.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,8 +102,7 @@ jobs:
             --restrict-key=test \
             > before.sql
           
-          psql "$DATABASE_URL" -c "TRUNCATE TABLE storage.migrations;"
-          npm run migration:run
+          npm run migration:test-idempotency
           
           pg_dump "$DATABASE_URL" \
             --exclude-table-data=storage.migrations \
--- a/migrations/tenant/0040-fix-prefix-race-conditions-optimized.sql
+++ b/migrations/tenant/0040-fix-prefix-race-conditions-optimized.sql
@@ -240,6 +240,10 @@ BEGIN
 END;
 $$;
 
+DROP TRIGGER IF EXISTS objects_delete_cleanup ON storage.objects;
+DROP TRIGGER IF EXISTS prefixes_delete_cleanup ON storage.prefixes;
+DROP TRIGGER IF EXISTS objects_update_cleanup ON storage.objects;
+
 -- Trigger bindings
 CREATE TRIGGER objects_delete_cleanup
     AFTER DELETE ON storage.objects
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "format": "prettier -c --write src/**",
     "lint": "prettier -v && prettier -c src/**",
     "migration:run": "tsx src/scripts/migrate-call.ts",
+    "migration:test-idempotency": "tsx src/scripts/test-migration-idempotency.ts",
     "migrations:types": "tsx src/scripts/migrations-types.ts",
     "docs:export": "tsx ./src/scripts/export-docs.ts",
     "test:dummy-data": "tsx -r dotenv/config ./src/test/db/import-dummy-data.ts",
--- a/src/internal/database/migrations/migrate.ts
+++ b/src/internal/database/migrations/migrate.ts
@@ -358,7 +358,7 @@ export async function runMigrationsOnTenant({
 }
 
 export async function resetMigration(options: {
-  tenantId: string
+  tenantId?: string
   untilMigration: keyof typeof DBMigration
   markCompletedTillMigration?: keyof typeof DBMigration
   databaseUrl: string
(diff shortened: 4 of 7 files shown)


migrations/tenant/0006-change-column-name-in-get-size.sql
```diff
@@ -1,4 +1,4 @@
-DROP FUNCTION storage.get_size_by_bucket();
+DROP FUNCTION IF EXISTS storage.get_size_by_bucket();
```
fenos, line 1: We should probably remove this DROP FUNCTION?

migrations/tenant/0038-iceberg-catalog-flag-on-buckets.sql
```diff
@@ -52,6 +54,7 @@ DO $$
           updated_at timestamptz NOT NULL default now()
         );
 
+        DROP INDEX IF EXISTS idx_iceberg_tables_namespace_id;
```
itslenny, line 57: does this index change? Wondering if it could be "create if not exists" like the above ones or if there is some reason to drop / re-create if it already exists.

migrations/tenant/0006-change-column-name-in-get-size.sql
TylerHillery: I called this one out in PR description
> it doesn't work if the return type of the function changes like it does with storage.get_size_by_bucket() so this needs to be dropped first

`CREATE OR REPLACE` doesn't work on functions where the return type changes. 

in migration 005 return type:

```sql
 RETURNS TABLE (
    size BIGINT,
    bucket text
  )
```

in migration 006 return type:

```sql
 RETURNS TABLE (
    size BIGINT,
    bucket_id text
  )
```

So unfortunately I don't think there is much we can do here.

migrations/tenant/0038-iceberg-catalog-flag-on-buckets.sql
TylerHillery: It does change in migration 0048 where we add `catalog_id` which is why I opted for the drop index and recreate.
```sql
-- migration 0038
DROP INDEX IF EXISTS idx_iceberg_tables_namespace_id;
CREATE UNIQUE INDEX idx_iceberg_tables_namespace_id ON storage.iceberg_tables (namespace_id, name);
 -- migration 0048
DROP INDEX IF EXISTS idx_iceberg_namespaces_bucket_id;
CREATE UNIQUE INDEX IF NOT EXISTS idx_iceberg_tables_namespace_id ON storage.iceberg_tables (catalog_id, namespace_id, name);
```

itslenny (approved):

PR #819, opened 2025-12-31
feat: otel metrics

fenos:

Feature

Currently metrics are gathered using the prom-client and exposed to a metric endpoint

## What is the new behavior?

- All metrics are now otel metrics and can be pushed to a 3rd party otel backend
- We still can optionally expose prometheus metrics in the /metrics endpoint using the otel prom exporter (`ENABLE_PROMETHEUS_METRICS`)
- All children spans will now include the tenant.ref attribute from a top level span
- Revamped grafana dashboard using otel metrics

## Grafana dashboard
Running completely locally

png (2)


itslenny (approved):

monitoring/grafana/dashboards/storage-otel.json
```diff
@@ -0,0 +1,5743 @@
...
+            "uid": "local_prometheus"
+          },
+          "editorMode": "code",
+          "expr": "sum(increase(storage_api_otel_upload_started_total{region=~\"$region\", instance=~\"$instance\"}[$__range]))",
+          "hide": false,
+          "instant": false,
+          "legendFormat": "__auto",
+          "range": true,
+          "refId": "C"
+        }
```
ferhatelmas, line 617: This was supposed to be multipart I guess according to description but instead duplicates query A

src/internal/monitoring/otel-instrumentation.ts
```diff
@@ -1,5 +1,53 @@
...
+import {
+  trace,
+  Span,
+  SpanStatusCode,
+  TracerProvider,
+  MeterProvider,
+  Context,
+} from '@opentelemetry/api'
+import { ReadableSpan, SpanProcessor } from '@opentelemetry/sdk-trace-base'
+import { Span as SdkSpan } from '@opentelemetry/sdk-trace-base/build/src/Span'
```
ferhatelmas: Is it possible not to import from build?

src/internal/monitoring/otel-metrics.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  // Initialize host metrics for Node.js runtime metrics
+  const hostMetrics = new HostMetrics({
+    meterProvider,
+    name: 'storage-api-host-metrics',
+  })
+  hostMetrics.start()
+
+  // Register Node.js runtime instrumentations
+  registerInstrumentations({
```
ferhatelmas, line 174: I think we need to capture returned function `unregister` here to pass `shutdown` below. Otherwise, it might prevent process from exiting

src/internal/monitoring/system/file-descriptor-collector.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  protected internalEnable(): void {
+    this._maxFds = null
+    this._openFds = null
+
+    // Update metrics periodically
+    this.updateMetrics().catch(() => {
+      // Ignore errors
+    })
+    this._updateInterval = setInterval(() => this.updateMetrics().catch(() => {}), 5000)
```
ferhatelmas, line 116: re:https://github.com/supabase/storage/pull/819/changes#r2686374454

If we address above comment, this would be unnecessary but I would add `unref` call here

src/storage/uploader.ts
```diff
@@ -236,11 +238,13 @@ export class Uploader {
 
         await Promise.all(events)
 
-        FileUploadedSuccess.inc({
-          is_multipart: uploadType === 'resumable' ? 1 : 0,
-          is_resumable: uploadType === 'resumable' ? 1 : 0,
-          is_standard: uploadType === 'standard' ? 1 : 0,
-          is_s3: uploadType === 's3' ? 1 : 0,
+        fileUploadedSuccess.add(1, {
+          is_multipart: String(uploadType === 'resumable' ? 1 : 0),
```
ferhatelmas: There is an inconsistency in the label values of `fileUploadStarted` (https://github.com/supabase/storage/pull/819/changes#diff-27bcc650282e1d4d6ab25be5c1bf343e75aa41e76844d93be70a146b4391a82bR83). Its `is_multipart` can be `true` or `false` but here `1` or `0`. Is this intentional?

src/internal/monitoring/otel-metrics.ts
fenos: Thanks! I have double checked but this function is not returning anything.
I guess it registers it internally, and on stop, it will clean it up

src/internal/monitoring/system/file-descriptor-collector.ts
fenos: This Otel collector implemention uses `enabled()` and `disbaled()` interface functions to start and stop the timer. They are called internally when otel start / stop. This should be already be handled ok

src/storage/uploader.ts
fenos: Agree fixed

src/internal/monitoring/otel-metrics.ts
ferhatelmas: It returns a [cleanup function](https://github.com/open-telemetry/opentelemetry-js/blob/b97b488a04dd38a20505194e1c4e7a0d8b4e81e0/experimental/packages/opentelemetry-instrumentation/src/autoLoader.ts#L25-L49), I believe we need to call it before meter shutdown for clean exit because it will stop timer for metric reader but won't touch the timer inside custom instrument. Another approach could be making timer unblocking as in https://github.com/supabase/storage/pull/819#discussion_r2686382468

PR #821, opened 2025-12-31
fix: use no cache for /object/info endpoint to prevent stale metadata

### LINKED ISSUE #728: `/object/info` is cached and the cache not busted after upserting
# Bug report

## Describe the bug

It seems like the `.info()` request that calls `/object/info` is cached and not busted after an upsert on the same path.

I came across this when migrating the existence check in supabase cache helpers to use `.exists()` + `.info()` instead of `.list()` for performance reasons (pr with the fix https://github.com/psteinroe/supabase-cache-helpers/pull/626)

If this is wanted, I would love some documentation on it.

## Reproduce
1. Upload file x 
2. call `/info` -> returns metadata for file x
3. Upsert file x with file y
4. call `/info` -> returns metadata for file x
5. call `/info?bust=true` -> returns metadata for file y

## Expected behavior

Should return metadata for file y after the upsert.

7ttp:
## Problem
The `/object/info` endpoint was using the object's `Cache-Control` metadata (e.g., `max-age=3600`) as the HTTP response header. This caused browsers/CDNs to cache the info response, returning stale metadata after upserts.

## Solution
Changed the `Cache-Control` header to `no-cache` so clients always revalidate with the server before using cached responses.

## related
- Closes https://github.com/supabase/storage/issues/728

Edit (saw this old pr) : 
not sure if this is the right fix, if not would be happy to document it as needed :)


ferhatelmas:
Thank you for your PR but it's addressed in the gateway level and not needed.

PR #825, opened 2026-01-23
feat: add user metadata support for  RLS policies

### LINKED ISSUE #757: Can't use `user_metadata` in RLS policies when uploading a file
# Bug report

## Describe the bug

I can't use custom values in `user_metadata` in RLS policies because they are not available in that context.
From what I've read it's because `storage.can_insert_object(...)` is not being called with `user_metadata` and so it doesn't _fake_ insert that info.

## To Reproduce

Create a RLS policy that just checks any value in `user_metadata`.

## Expected behavior

I can access in the RLS policy the custom values I defined during the upload call.

## System information

- OS: macOs
- Browser (if applies):
- Version of supabase-js: 2.56.0
- Version of Node.js: 22.12

TylerHillery:

resolves #757 by adding support for referencing `user_metadata` field in RLS policies for inserts

Currently `user_metadata` is not passed into `canUpload` checks, so RLS policies can't reference it during file uploads.

## What is the new behavior?

 Now `user_metadata` is passed through all upload paths (standard HTTP, S3, TUS, multipart) so RLS policies can use it.

## Additional context

- The main change I made was to `canUpload` to accept `userMetatdata`. I updated all callers to pass in `userMetadata`.  Because `upload` gets called in `prepareUpload` which gets called in `upload` I also updated all `upload` calls to ensure `userMetadata` is passed through. I found it difficult to ensure I caught every cause though, e.g. it was very easy to miss `uploadNewObject` needed to be updated in the `uploadFromRequest`. I'm open to feedback to see if there is a more robust solution we could implement that could be caught with typescript types.
- The part I found most confusing to update was the `canUpload` checks for multipart uploads. It required changing the order of certain operations around so that `user_metadata` was available. Wasn't entirely sure if these changes were okay.
- Added some context as to why I moved `userMetadata` from the `FileUpload` interface to the `UploadRequest` but wasn't sure if this was the right approach so would appreciate some feedback

### DIFF
--- a/src/storage/uploader.ts
+++ b/src/storage/uploader.ts
@@ -116,7 +116,10 @@ export class Uploader {
       owner: request.owner,
       isUpsert: request.isUpsert,
       userMetadata: request.userMetadata,
-      metadata: { mimetype: request.file.mimeType, contentLength: request.file.declaredContentLength ?? request.file.contentLength },
+      metadata: {
+        mimetype: request.file.mimeType,
+        contentLength: request.file.declaredContentLength ?? request.file.contentLength,
+      },
       uploadType: request.uploadType,
     })
 


src/storage/uploader.ts
```diff
@@ -48,7 +48,9 @@ export class Uploader {
     private readonly location: StorageObjectLocator
   ) {}
```
TylerHillery, line 61: I wasn't sure if moving `userMetadata` property from `FileUpload` to `UploadRequest` was the right call. Originally, I tried leaving it but that required adding `& {userMetadata: Record<string, known>` to many places, because `file` is often omitted from many functions where `canUpload` is called.

The most awkward place would have been in the `s3-handler.ts` it would have required to pass in userMetadata twice both in the `file` and the `UploadRequest` which felt wrong so I opted moving `userMetadata` to the `UploadRequest` interface.

```typescript
    const upload = await uploader.upload({
      bucketId: command.Bucket as string,
      file: {
        body: command.Body as Readable,
        cacheControl: command.CacheControl!,
        mimeType: command.ContentType!,
        isTruncated: options.isTruncated,
       // if I left userMetadata on FileUpload interface userMetadata would have to be added here
      },
      objectName: command.Key as string,
      userMetadata: command.Metadata,
      owner: this.owner,
      isUpsert: true,
      uploadType: 's3',
      signal: options.signal,
    })
```

I also deliberately made userMetadata required (but allowing undefined) because I was worried it would be easy to forget to pass it.

src/http/routes/tus/lifecycle.ts
```diff
@@ -67,6 +67,20 @@ export async function onIncomingRequest(rawReq: Request, id: string) {
...
+  // it matches the value on Upload.metadata when inserted in storage.objects
+  // link: https://github.com/tus/tus-node-server/blob/1a6482a7a55e1587bda8c6887250f36cf9d606bd/packages/server/src/handlers/PostHandler.ts#L46
+  let customMd: Record<string, string | null> | undefined = undefined
+  const uploadMetadataHeader = req.headers['upload-metadata']
+
+  if (uploadMetadataHeader && typeof uploadMetadataHeader === 'string') {
+    try {
+      customMd = Metadata.parse(uploadMetadataHeader)
+    } catch (e) {
+      req.log.warn({ error: e }, 'Failed to parse user metadata')
```
TylerHillery: should I just do a //no-op here like we do on `onUploadFinish` parsing the metadata or are we okay with the req.log.warn?

fenos: I think this is incorrect.
The userMetadata will under `customMd.metadata` which should be JSON.parsed

TylerHillery: good catch, this has been updated:
```typescript
      const parsedMetadata = Metadata.parse(uploadMetadataHeader)
      if (parsedMetadata?.metadata) {
        customMd = JSON.parse(parsedMetadata.metadata)
      }
```

src/http/routes/object/getSignedUploadURL.ts
```diff
@@ -69,10 +70,19 @@ export default async function routes(fastify: FastifyInstance) {
...
+
+      if (typeof customMd === 'string') {
+        userMetadata = parseUserMetadata(customMd)
+      }
+
       const signedUpload = await request.storage
         .from(bucketName)
         .signUploadObjectUrl(objectName, urlPath as string, uploadSignedUrlExpirationTime, owner, {
           upsert: request.headers['x-upsert'] === 'true',
+          userMetadata: userMetadata,
```
fenos: I think we don't actually support userMetadata on upload signed urls (which we should) so I'm glad you found this. 

I believe we need to store the userMetadata in the returned JWT payload; otherwise, this information will be lost during the signedUpload request. We need to be careful, though, if we decide to store the metadata in the JWT, because it can be abused with large payloads. We can for now restrict the amount of data that we can store in the JWT to something like 1MB i think we already have a constant MAX_CUSTOM_METADATA_SIZE, which can be reused (re-refactored as a validation function, potentially in limits.ts)

But we also need to add the second part (on the upload route) to store the userMetadata received from the JWT. This way, the metadata can't be tempered.

src/storage/uploader.ts
fenos: This change is good 👍

src/http/routes/object/getSignedUploadURL.ts
TylerHillery: I will do a follow up PR to handle the upload part.

src/http/routes/object/getSignedUploadURL.ts
```diff
@@ -69,10 +72,27 @@ export default async function routes(fastify: FastifyInstance) {
...
+
+      const contentType = request.headers['content-type']
+      const contentLengthHeader = request.headers['content-length']
+      const contentLength = contentLengthHeader ? Number(contentLengthHeader) : undefined
+
       const signedUpload = await request.storage
         .from(bucketName)
         .signUploadObjectUrl(objectName, urlPath as string, uploadSignedUrlExpirationTime, owner, {
           upsert: request.headers['x-upsert'] === 'true',
+          userMetadata: userMetadata,
```
ferhatelmas:
```suggestion
          userMetadata,
```

src/http/routes/tus/lifecycle.ts
```diff
@@ -67,6 +67,20 @@ export async function onIncomingRequest(rawReq: Request, id: string) {
 
   req.upload.resources = [`${uploadID.bucket}/${uploadID.objectName}`]
 
+  let customMd: Record<string, string> | undefined = undefined
```
ferhatelmas: nit: I would use `customMetadata`

ferhatelmas: or maybe directly `userMetadata`

src/storage/uploader.ts
```diff
@@ -22,19 +22,32 @@ interface FileUpload {
...
   xRobotsTag?: string
-  userMetadata?: Record<string, unknown>
+  contentLength?: number
 }
 
 export interface UploadRequest {
   bucketId: string
   objectName: string
   file: FileUpload
+  userMetadata: Record<string, unknown> | undefined
```
ferhatelmas: why do we diverge from optionality? `userMetadata?: Record<string, unknown>`

src/test/scanner.test.ts
```diff
@@ -24,11 +24,11 @@ describe('ObjectScanner', () => {
         bucketId: bucket.id,
         objectName: randomUUID() + `-test-${i}.text`,
         uploadType: 'standard',
+        userMetadata: {},
```
ferhatelmas: if types were more idiomatic, we wouldn't expect any changes in this test right?

src/http/routes/object/getSignedUploadURL.ts
ferhatelmas: I see there are a lot of places we could apply this simplification so for now, feel free to leave as it is for consistency. We might want to address globally later with a linter.

src/http/routes/tus/lifecycle.ts
```diff
@@ -96,11 +110,20 @@ export async function onIncomingRequest(rawReq: Request, id: string) {
     req.upload.storage.location
   )
 
+  const uploadLength = req.headers['upload-length']
+  const contentLength = uploadLength ? Number(uploadLength) : undefined
+  const contentType = req.headers['content-type']
```
ferhatelmas: This isn't completely right? If it's an upsert (patch), headers on the incoming request will be different than creation. 

For example, creation is an `image/png` which passes RLS (`metadata->>'mimetype' LIKE 'image/%'`) then patch probably will be `application/offset+octet-stream` and it would fail

src/http/routes/tus/lifecycle.ts
ferhatelmas: I see in some places it's named as `customMd` so fine to leave as it is again for consistency.

src/storage/database/knex.ts
```diff
@@ -902,7 +903,8 @@ export class StorageKnexDB implements Database {
             version,
             upload_signature: signature,
             owner_id: owner,
-            user_metadata: metadata,
+            user_metadata: userMetadata,
+            metadata: metadata,
```
ferhatelmas: don't we need a guard rule for new column similar to `user_metadata`?

src/storage/protocols/s3/s3-handler.ts
ferhatelmas, line 707: It seems it pulls content length from here but it's not passed. If there is a RLS policy, it could fail

src/http/routes/tus/lifecycle.ts
TylerHillery: I agree I thought `customMd` was confusing but used it because that's what it was called in the `onUploadFinish` function. I will leave it for now for consistency

src/storage/uploader.ts
TylerHillery: While implementing this change I found it very easy to forget to pass `userMetadata` somewhere when it should be passed so I thought I make it required but allow `undefined` if it's not needed.

I called this out in the PR as wasn't sure if there was a better way and wasn't 100% confident about this solution. If you think optionality is better here I am happy to change back.

>  calls to ensure userMetadata is passed through. I found it difficult to ensure I caught every cause though, e.g. it was very easy to miss uploadNewObject needed to be updated in the uploadFromRequest. I'm open to feedback to see if there is a more robust solution we could implement that could be caught with typescript type

src/test/scanner.test.ts
TylerHillery: correct, this goes back to the optionality change, I replied back to your comment on this to get your thoughts.

src/http/routes/object/getSignedUploadURL.ts
TylerHillery: good catch, I can change this one for now but agree we should add a lint rule for this to fix the rest of them

src/storage/database/knex.ts
TylerHillery: yes you're right. I added a guard rule but it feels a bit hacky. The reason I didn't add it to `rules` in `normalizedColumns` is because `metadata` column does exist in `objects` table. It's my understanding any `metadata` column would get remove regardless of the table it's selecting from. 

So instead I checked if latest migration was >= so s3-multipart-uploads-metadata and added it to the list of columns to select.

I almost feel `normalizedColumns` should be table dependent. This might be better in a follow up PR thought.

src/storage/protocols/s3/s3-handler.ts
TylerHillery: added in most recent commit 👍

src/http/routes/tus/lifecycle.ts
TylerHillery: good catch, I have updated to handle the difference between the first upload vs patch. I use the `Datastore` to get the initial contenttype and size.

I have also updated the `rls.test.ts` to include a new `tus.upload` to test this.

src/http/routes/object/getSignedUploadURL.ts
```diff
@@ -20,6 +21,8 @@ const getSignedUploadURLHeadersSchema = {
   type: 'object',
   properties: {
     'x-upsert': { type: 'string' },
+    'content-type': { type: 'string' },
+    'content-length': { type: 'string' },
```
ferhatelmas, line 26: don't we need `x-metadata` as well since we try to read below?

src/http/routes/object/getSignedUploadURL.ts
```diff
@@ -69,10 +72,27 @@ export default async function routes(fastify: FastifyInstance) {
 
       const urlPath = `${bucketName}/${objectName}`
 
+      let userMetadata: Record<string, unknown> | undefined
+
+      const customMd = request.headers['x-metadata']
+
+      if (typeof customMd === 'string') {
+        userMetadata = parseUserMetadata(customMd)
```
ferhatelmas, line 83: this parser returns `Record<string, string>` by casting due to its expectation (and S3 compatibility) but there is no validation. Values could be anything. 

This is an existing helper, feel free to add a TODO to be addressed later

ferhatelmas: I looked into supabase-js and s3 part of this, type is wrong but handling is correct and we return `x-amz-meta-missing` header for non-string/non-printable

src/storage/uploader.ts
ferhatelmas: I see your point but I would go with optional since more idiomatic. Maybe for cleanup afterwards

src/http/routes/object/getSignedUploadURL.ts
ferhatelmas: Another thing to note is that when it's set when content-type is `multipart/form-data`, we check/limit its size but not when it comes from a header. 

~However, the value is 1kb and a header can definitely be larger than that as well~ saw value wrong, it's 1mb so this is a valid assumption. We could reuse the code, instead of two parsers as a cleaup but not in this PR.

src/storage/protocols/s3/s3-handler.ts
```diff
@@ -578,15 +585,18 @@ export class S3ProtocolHandler {
     const maxFileSize = await getFileSizeLimit(this.storage.db.tenantId, bucket?.file_size_limit)
 
     const uploader = new Uploader(this.storage.backend, this.storage.db, this.storage.location)
+
+    const multipart = await this.shouldAllowPartUpload(UploadId, ContentLength, maxFileSize)
```
ferhatelmas: this ordering can cause issues for valid uploads because it mutates the state.

For example (numbers are arbitrary)
1. in progress 50mb and new part is 5mb
2. in progress becomes 55mb
3. RLS fails 
4. next request in progress becomes 60mb
5. RLS checks but size fails because it was expecting at most 55mb

src/storage/protocols/s3/s3-handler.ts
```diff
@@ -735,14 +746,16 @@ export class S3ProtocolHandler {
 
     const multipart = await this.storage.db
       .asSuperUser()
-      .findMultipartUpload(UploadId, 'id,version')
+      .findMultipartUpload(UploadId, 'id,version,user_metadata,metadata')
```
ferhatelmas, line 753: we need the column guard for these reads similar to create as well

src/storage/object.ts
```diff
@@ -339,6 +340,8 @@ export class ObjectStorage {
       objectName: destinationKey,
       owner,
       isUpsert: upsert,
+      userMetadata: userMetadata,
```
ferhatelmas: similar to metadata, aren't we supposed to use `origin.user_metadata` if `copyMetadata` is false because it's the written value eventually?

src/storage/database/knex.ts
ferhatelmas: totally agree, also see https://github.com/supabase/storage/pull/825#discussion_r2847979213

src/storage/uploader.ts
TylerHillery: reverted back to optional

src/http/routes/object/getSignedUploadURL.ts
TylerHillery: yes, added

src/http/routes/object/getSignedUploadURL.ts
TylerHillery: added TODO

src/storage/protocols/s3/s3-handler.ts
TylerHillery: go catch, changed the order to have `canUpload` called before `shouldAllowPartUpload`. What I don't like about this is that this adds an extra call to `findMultipartUpload` as this is also called in `shouldAllowPartUpload` I thought about moving `canUpload` within in `shouldAllowPartUpload` but it uses `asSuperUser` before starting the transaction which would bypass RLS.

Can you confirm my understanding? Is there anyway to do this that I'm missing that doesn't involve extra `findMultipartUpload` call?

src/storage/protocols/s3/s3-handler.ts
TylerHillery: added the guard to the `findMultipartUpload` with a TODO comment to refactor once we change `normalizeColumns` to be table aware.

src/storage/object.ts
TylerHillery: yes we are, changed it to the same logic we use for metadata

TylerHillery:
> Can we also add more tests please?
> copy with copyMetadata=true uses correct user metadata for RLS check
> multipart s3 RLS with metadata
> multipart progress accounting on auth failure
> signed upload RLS with metadata
> migration compatibility

Added all the above except for `migration compatibility` what did you have in mind for this? In [CI](https://github.com/supabase/storage/blob/master/.github/workflows/ci.yml#L96) we do ensure all migrations run and are deleted and rerun to ensure they're idempotent. Does this satisfy the migration compatibility?

src/storage/object.ts
TylerHillery: I take it back, using the same logic breaks some tests for `userMetadata` we don't want to merge the two. Either use all of the origin if copy or use the new destination metadata

PR #827, opened 2026-01-24
feat: optimised search query

fenos:
### Overview                                                                                                                                                  
                                                                                                                                                            
This PR introduces a major performance optimization for the storage object listing/search functionality by replacing the prefixes table and its associated triggers with an efficient hybrid skip-scan algorithm that derives folder structure on-the-fly from the objects table.                                   
                                                                                                                                                            
  Problem Statement                                                                                                                                         
                                                                                                                                                            
  The current implementation maintains a separate prefixes table that tracks folder hierarchies. This approach has several issues:                          
                                                                                                                                                            
  - Slow writes: Every INSERT/UPDATE/DELETE on objects triggers cascading operations to maintain the prefixes table                                         
  - Lock contention: The prefix triggers acquire row-level locks that cause contention under concurrent writes                                              
  - Race conditions: Complex trigger logic to handle prefix reference counting led to edge cases and race conditions                                        
  - Storage overhead: The prefixes table duplicates information that can be derived from object names                                                       
  - Pagination degradation: The search function uses OFFSET-based pagination which degrades linearly with depth (O(n) for page n)                           
                                                                                                                                                            
### Solution                                                                                                                                                  
                                                                                                                                                            
  New Algorithm: *Hybrid Skip-Scan*                                                                                                                          
                                                                                                                                                            
 The new implementation uses a hybrid peek-then-batch algorithm that provides O(k × log n) performance for listing k items from n total objects:           
                                                                                                                                                            
   1. PEEK (LIMIT 1)     → Find next item (index-only scan)   │                                                                                           
  │  2. Is it a folder?                                         │                                                                                           
  │     YES → Emit folder name, skip to next prefix range       │                                                                                           
  │     NO  → Batch fetch files until next folder               │                                                                                           
  │  3. Repeat until max_keys reached                                                                                                   
                                                                                                                                                            
  Key optimizations:                                                                                                                                        
  - Static SQL for peek queries (hot path): Plan cached, executes in microseconds                                                                           
  - Dynamic SQL for batch queries (cold path): Overhead amortized over many rows                                                                            
  - Cursor-based pagination: Constant time regardless of page depth                                                                                         
  - No heap access for folders: Only needs the name column from the index                                                                                   
                                                                                                                                                            
  Performance Results (60M rows benchmark)                                                                                                                  
                                                                                                                                                            
  Standard Queries (1000 items)                                                                                                                             
                                                                                                                                                            
  | Scenario                   | V1 (old)  | V2 (new) | Improvement      |                                                                                  
  |----------------------------|-----------|----------|------------------|                                                                                  
  | List 1000 files (flat)     | 2.42ms    | 1.79ms   | **1.35x faster** |                                                                                  
  | List 1000 folders (dense)  | 11.78ms   | 10.02ms  | **1.18x faster** |                                                                                  
  | List 1000 tenant folders   | 18.81ms   | 13.67ms  | **1.38x faster** |                                                                                  
  | Page 20 (offset 19000)     | 34.66ms   | 2.34ms   | **14.8x faster** |                                                                                  
                                                                                                                                                            
  Pagination Performance                                                                                                                                    
                                                                                                                                                            
  The most dramatic improvement is in pagination. The old OFFSET-based approach degrades linearly:                                                          
                                                                                                                                                            
  | Page | V1 (OFFSET) | V2 (Cursor) |                                                                                                                      
  |------|-------------|-------------|                                                                                                                      
  | 1    | 2.5ms       | 2.2ms       |                                                                                                                      
  | 10   | 18.2ms      | 2.4ms       |                                                                                                                      
  | 20   | 34.7ms      | 2.3ms       |                                                                                                                      
                                                                                                                                                            
  V2 maintains constant ~2.3ms regardless of page depth.                                                                                                    
                                                                                                                                                            
  Changes                                                                                                                                                   
                                                                                                                                                            
  New Functions                                                                                                                                             
                                                                                                                                                            
  - storage.list_objects_with_delimiter - Core S3-compatible listing with hybrid skip-scan                                                                  
  - storage.search_v2 - Main entry point, routes to appropriate implementation                                                                              
  - storage.search - Updated legacy function with same optimizations                                                                                        
  - storage.search_by_timestamp - Aggregation-based implementation for non-name sorting                                                                     
  - storage.get_common_prefix - Helper to extract folder prefixes from object names                                                                         
                                                                                                                                                            
  Removed Components                                                                                                                                        
                                                                                                                                                            
  - storage.prefixes table - No longer needed                                                                                                               
  - 6 triggers on objects table: objects_insert_create_prefix, objects_update_create_prefix, objects_delete_cleanup, etc.                                   
  - 12 helper functions: add_prefixes, delete_prefix, get_prefix, lock_top_prefixes, etc.                                                                   
  - level column on objects table and associated index                                                                                                      
  - 2 test files: prefix-functions.test.ts, prefixes.test.ts (1827 lines removed)                                                                           
                                                                                                                                                            
  Index Changes                                                                                                                                             
                                                                                                                                                            
  - Dropped: idx_objects_lower_name, objects_bucket_id_level_idx                                                                                            
  - Relies on existing: idx_objects_bucket_id_name with COLLATE "C" for bytewise ordering                                                                   
                                                                                                                                                            
  Migration Strategy                                                                                                                                        
                                                                                                                                                            
  The migration is designed to be backwards compatible:                                                                                                     
                                                                                                                                                            
  1. New functions are created with the same signatures                                                                                                     
  2. Triggers are dropped (writes become faster immediately)                                                                                                
  3. prefixes table is dropped in a subsequent migration                                                                                                    
  4. Old indexes are dropped after functions no longer reference them                                                                                       

  Testing                                                                                                     
                                                                                                                                                            
  - All existing tests pass                                                                                                                                 
  - New test file object-list-v2.test.ts covers:                                                                                                            
    - ASC/DESC sorting                                                                                                                                      
    - With/without delimiter                                                                                                                                
    - With/without prefix                                                                                                                                   
    - Pagination across all sorting modes                                                                                                                   
    - created_at and updated_at sorting                                                                                                                     
                                                                                                                                                            
  Trade-offs                                                                                                                                                
                                                                                                                                                            
  | Aspect            | Before                          | After                              |                                                              
  |-------------------|---------------------------------|------------------------------------|                                                              
  | Write performance | Slow (trigger overhead)         | Fast (no triggers)                 |                                                              
  | Read performance  | Fast for first page             | Fast for all pages                 |                                                              
  | Storage           | Extra `prefixes` table          | None                               |                                                              
  | Complexity        | High (triggers, race conditions)| Low (single algorithm)             |                                                              
  | Worst case        | Pagination degrades             | Interleaved file/folder pattern*   |                                                              
                                                                                                                                                            
  *The worst case for the new algorithm is a pattern with alternating files and folders where each folder contains only 1 file. In this adversarial case,   
  performance is similar to V1 for the first page, but V2 is still dramatically faster for subsequent pages.


migrations/tenant/0050-search-v2-optimised.sql
```diff
@@ -0,0 +1,804 @@
...
+-- Parameters:
+--   bucket_id: The bucket identifier
+--   prefix_param: Filter objects starting with this prefix
+--   delimiter_param: The delimiter for grouping common prefixes (e.g., '/')
+--   max_keys: Maximum number of results to return
+--   start_after: Start listing after this key (for pagination)
+--   next_token: Continuation token (same as start_after for compatibility)
+--   sort_order: 'asc' or 'desc' (default 'asc')
+-- ============================================================================
+DROP FUNCTION IF EXISTS storage.search(prefix text, bucketname text, limits integer, levels integer, offsets integer);
```
itslenny: If the function signature is not changing we don't need to `drop` and can just `create or replace` in place.

If a project's SQL references on one of these functions (e.g. in a trigger or rls) the drop function will fail causing this entire migration to fail. So if we CAN avoid the drop we probably should.

...If the signature does change then we have to use drop.

migrations/tenant/0052-drop-not-used-indexes-and-functions.sql
```diff
@@ -0,0 +1,33 @@
...
+DROP FUNCTION IF EXISTS storage.delete_prefix(_bucket_id text, _name text);
+DROP FUNCTION IF EXISTS storage.add_prefixes(_bucket_id text, _name text);
+DROP FUNCTION IF EXISTS storage.get_prefix;
+DROP FUNCTION IF EXISTS storage.objects_update_level_trigger();
+DROP FUNCTION IF EXISTS storage.objects_update_prefix_trigger();
+DROP FUNCTION IF EXISTS storage.lock_top_prefixes(bucket_ids text[], names text[]);
+DROP FUNCTION IF EXISTS storage.search_v1_optimised(prefix text, bucketname text, limits integer, levels integer, offsets integer, search text, sortcolumn text, sortorder text);
+
+
+DROP TABLE IF EXISTS storage.prefixes cascade;
```
itslenny: 🥳

migrations/tenant/0050-search-v2-optimised.sql
fenos: Yhea, if i don't drop this function i was ending up with 2 search_v2 functions.
I've added a try catch in case so that the migration can still run without failing

PR #841, opened 2026-02-04
Perf/query cancellation stricter timeouts

fenos:

Performance improvement

- No query cancellation when the request aborts
- No strict timeouts
- Update nodejs base image from 22 - 24

## What is the new behavior?

- Implement query cancellation using the request abort signal. In order for this to work i applied the changes of this [PR](https://github.com/knex/knex/pull/6253) for `knex`.
- We also used a different way to cancel queries, by default knex uses `pg_terminate()` but this isn't supported by pgBouncer. So we are using the Postgres native cancel protocol to cancel queries which works with pgBouncer too.

- Added stricter timeout to all queries to up to 30s. It is already generous but can be reconfigured via ENV
- General small performance improvements

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   "author": "Supabase",
   "license": "ISC",
   "engines": {
-    "node": ">= 14.0.0"
+    "node": ">= 22.0.0"
   },
   "dependencies": {
     "@aws-sdk/client-ecs": "^3.948.0",
--- a/src/config.ts
+++ b/src/config.ts
@@ -65,6 +65,7 @@ type StorageConfigType = {
   storageFileEtagAlgorithm: 'mtime' | 'md5'
   storageS3InternalTracesEnabled?: boolean
   storageS3MaxSockets: number
+  storageS3UploadQueueSize: number
   storageS3Bucket: string
   storageS3Endpoint?: string
   storageS3ForcePathStyle?: boolean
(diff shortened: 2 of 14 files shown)


src/config.ts
```diff
@@ -436,7 +441,8 @@ export function getConfig(options?: { reload?: boolean }): StorageConfigType {
     // Queue
     pgQueueEnable: getOptionalConfigFromEnv('PG_QUEUE_ENABLE', 'ENABLE_QUEUE_EVENTS') === 'true',
     pgQueueEnableWorkers: getOptionalConfigFromEnv('PG_QUEUE_WORKERS_ENABLE') !== 'false',
-    pgQueueReadWriteTimeout: Number(getOptionalConfigFromEnv('PG_QUEUE_READ_WRITE_TIMEOUT')) || 0,
+    pgQueueReadWriteTimeout:
```
ferhatelmas, line 447: When set to zero, it would disable previously but now will default to 5s. Is this intentional because disabling won't be possible now?

package.json
```diff
@@ -93,6 +94,7 @@
     "xml2js": "^0.6.2"
   },
   "devDependencies": {
+    "patch-package": "^8.0.0",
```
ferhatelmas, line 97: after knex is released, are we going to drop this?

fenos: yes we will

src/config.ts
fenos: Good catch

fenos: fixed

ferhatelmas: Since it's optional, it can be undefined and returns NaN. In that case, we want to pass default timeout but it won't.

fenos: @copilot fix it 😄

fenos: Ok, It should be fixed for good now, I did better than Co-Pilot 😄

src/storage/database/knex.ts
```diff
@@ -1022,18 +1036,20 @@ export class StorageKnexDB implements Database {
...
     const startTime = process.hrtime.bigint()
     const recordDuration = () => {
       const duration = Number(process.hrtime.bigint() - startTime) / 1e9
       dbQueryPerformance.record(duration, {
         name: queryName,
+        tenantId: this.tenantId,
       })
     }
 
+    const abortSignal = this.connection.getAbortSignal()
```
ferhatelmas, line 1055: It seems abortSignal is set in request scope correctly but while calling for super user, another tenant connection is created and it doesn't get the abort signal. Since main connection is passed in options, we could/should fall back here to it if abort signal is undefined. 

Something like:

```js
const abortSignal = this.connection.getAbortSignal() ?? this.options.parentConnection?.getAbortSignal()
```

It could be copied as well but I think this is simpler

package.json
ferhatelmas: good to update here as well? 

because no guarantee if older versions will continue working and also prevents us from adapting new features

fenos: Yes

PR #854, opened 2026-02-09
Improved adapter interface and explicit signal cancellation

fenos:
                                                                                                                                                            
  Refactor                                                                                                                                                  
                                                                                                    
  - Backend adapter methods (`S3Backend`, `FileBackend`) use **positional parameters** (e.g., `read(bucket, key, version, headers, signal)`), making call sites hard to read and fragile when parameters change.
  - Database methods use **positional parameters** or inline `Pick<>` types with no named input interfaces.
  - Abort signal is set once at the **connection level** via `TenantConnection.setAbortSignal()`, meaning all queries on a connection are cancellable, including cleanup queries in catch blocks, upload completion transactions, and internal maintenance queries. This has caused bugs where critical operations (like finalizing an upload) get cancelled when a client disconnects mid-request.
  - `Storage` and `ObjectStorage` methods use positional parameters.

  ## What is the new behavior?

  ### Single object parameter pattern with named input interfaces

  All layers now use a single object parameter with named input interfaces:

  | Layer | Before | After |
  |-------|--------|-------|
  | Backend | `read(bucket, key, version, headers)` | `read({ bucket, key, version, headers })` |
  | Database | `findBucketById(id, columns, filters)` | `findBucketById({ bucketId, columns, filters })` |
  | Storage | `findBucket(id, columns)` | `findBucket({ bucketId, columns })` |
  | ObjectStorage | `deleteObject(name)` | `deleteObject({ objectName })` |

  Input interfaces are defined in:
  - `src/storage/backend/adapter.ts` — `ReadObjectInput`, `WriteObjectInput`, `CopyObjectInput`, etc.
  - `src/storage/database/adapter.ts` — `FindBucketByIdInput`, `DeleteObjectInput`, `CreateMultipartUploadInput`, etc.

  ### Per-query abort signal instead of per-connection

  Abort signal now flows **explicitly through method inputs** instead of being bound to the connection:

  Route → Storage/ObjectStorage → DB method → .abortOnSignal(signal) per query

  **What changed:**

  - Removed `setAbortSignal()` / `getAbortSignal()` from `TenantConnection`
  - Removed connection-level signal setup from the `db.ts` plugin
  - All DB input interfaces extend a `Cancellable` base interface (`{ signal?: AbortSignal }`)
  - `withTransaction` propagates signal into child transactions so inner queries auto-inherit
  - Removed the `AbortController` workaround in the uploader that existed to prevent upload finalization from being cancelled

  **Signal routing by route type:**

  | Route type | Signal source |
  |------------|--------------|
  | Standard REST routes | `request.signals.disconnect.signal` |
  | S3 reads/lists/deletes | `ctx.signals.response` |
  | S3 uploads | `ctx.signals.body` |
  | TUS routes | `req.upload.signal` (set from `req.signals.disconnect.signal`) |

  **Intentionally unsignalled operations** — catch blocks, upload completion (`completeMultiPartUpload`, `onUploadFinish`), and queue event handlers do not
  receive a signal, ensuring they always run to completion regardless of client disconnect.

  ### File renames

  - `src/storage/backend/file.ts` → `src/storage/backend/file/file-adapter.ts`
  - `src/storage/backend/s3/adapter.ts` → `src/storage/backend/s3/s3-adapter.ts`
  - `src/storage/backend/s3/backup.ts` → `src/storage/backend/s3/s3-backup.ts`

  ## Additional context

  - Tests updated to match new single-object-parameter mock assertions

### DIFF
--- a/src/http/routes/object/listObjects.ts
+++ b/src/http/routes/object/listObjects.ts
@@ -81,6 +81,7 @@ export default async function routes(fastify: FastifyInstance) {
             order: sortBy?.order,
           },
         },
+        signal: request.signals.disconnect.signal,
       })
 
       return response.status(200).send(results)
--- a/src/http/routes/s3/commands/copy-object.ts
+++ b/src/http/routes/s3/commands/copy-object.ts
@@ -38,24 +38,29 @@ export default function CopyObject(s3Router: S3Router) {
     (req, ctx) => {
       const s3Protocol = new S3ProtocolHandler(ctx.storage, ctx.tenantId, ctx.owner)
 
-      return s3Protocol.copyObject({
-        Bucket: req.Params.Bucket,
-        Key: req.Params['*'],
-        CopySource: req.Headers['x-amz-copy-source'],
-        ContentType: req.Headers['content-type'],
-        CacheControl: req.Headers['cache-control'],
-        MetadataDirective: req.Headers['x-amz-metadata-directive'] as MetadataDirective | undefined,
-        Expires: req.Headers.expires ? new Date(req.Headers.expires) : undefined,
-        ContentEncoding: req.Headers['content-encoding'],
-        CopySourceIfMatch: req.Headers['x-amz-copy-source-if-match'],
-        CopySourceIfModifiedSince: req.Headers['x-amz-copy-source-if-modified-since']
-          ? new Date(req.Headers['x-amz-copy-source-if-modified-since'])
-          : undefined,
-        CopySourceIfNoneMatch: req.Headers['x-amz-copy-source-if-none-match'],
-        CopySourceIfUnmodifiedSince: req.Headers['x-amz-copy-source-if-unmodified-since']
-          ? new Date(req.Headers['x-amz-copy-source-if-unmodified-since'])
-          : undefined,
-      })
+      return s3Protocol.copyObject(
+        {
+          Bucket: req.Params.Bucket,
+          Key: req.Params['*'],
+          CopySource: req.Headers['x-amz-copy-source'],
+          ContentType: req.Headers['content-type'],
+          CacheControl: req.Headers['cache-control'],
+          MetadataDirective: req.Headers['x-amz-metadata-directive'] as
+            | MetadataDirective
+            | undefined,
+          Expires: req.Headers.expires ? new Date(req.Headers.expires) : undefined,
+          ContentEncoding: req.Headers['content-encoding'],
+          CopySourceIfMatch: req.Headers['x-amz-copy-source-if-match'],
+          CopySourceIfModifiedSince: req.Headers['x-amz-copy-source-if-modified-since']
+            ? new Date(req.Headers['x-amz-copy-source-if-modified-since'])
+            : undefined,
+          CopySourceIfNoneMatch: req.Headers['x-amz-copy-source-if-none-match'],
+          CopySourceIfUnmodifiedSince: req.Headers['x-amz-copy-source-if-unmodified-since']
+            ? new Date(req.Headers['x-amz-copy-source-if-unmodified-since'])
+            : undefined,
+        },
+        ctx.signals.response
+      )
     }
   )
 }
--- a/src/internal/monitoring/otel-class-instrumentations.ts
+++ b/src/internal/monitoring/otel-class-instrumentations.ts
@@ -5,7 +5,7 @@ import { ObjectStorage } from '@storage/object'
 import { Uploader } from '@storage/uploader'
 import { Storage } from '@storage/storage'
 import { Event as QueueBaseEvent } from '@internal/queue'
-import { S3Backend } from '@storage/backend'
+import { S3Adapter } from '@storage/backend'
 import { StorageKnexDB } from '@storage/database'
 import { TenantConnection } from '@internal/database'
 import { S3Store } from '@tus/s3-store'
@@ -67,7 +67,7 @@ export const classInstrumentations = [
     },
   }),
   new ClassInstrumentation({
-    targetClass: S3Backend,
+    targetClass: S3Adapter,
     enabled: true,
     methodsToInstrument: [
       'getObject',
--- a/src/storage/backend/file/file-adapter.ts
+++ b/src/storage/backend/file/file-adapter.ts
@@ -51,10 +51,10 @@ const METADATA_ATTR_KEYS = {
 }
 
 /**
- * FileBackend
+ * FileAdapter
  * Interacts with the file system with this FileBackend adapter
  */
-export class FileBackend implements StorageBackendAdapter {
+export class FileAdapter implements StorageBackendAdapter {
   client = null
   filePath: string
   etagAlgorithm: 'mtime' | 'md5'
--- a/src/storage/backend/s3/s3-adapter.ts
+++ b/src/storage/backend/s3/s3-adapter.ts
@@ -60,10 +60,10 @@ export interface S3ClientOptions {
 }
 
 /**
- * S3Backend
+ * S3Adapter
  * Interacts with a s3-compatible file system with this S3Adapter
  */
-export class S3Backend implements StorageBackendAdapter {
+export class S3Adapter implements StorageBackendAdapter {
   client: S3Client
   agent: InstrumentedAgent
 
--- a/src/storage/database/adapter.ts
+++ b/src/storage/database/adapter.ts
@@ -3,6 +3,10 @@ import { ObjectMetadata } from '../backend'
 import { TenantConnection } from '@internal/database'
 import { DBMigration } from '@internal/database/migrations'
 
+export interface Cancellable {
+  signal?: AbortSignal
+}
+
 export interface SearchObjectOption {
   search?: string
   sortBy?: {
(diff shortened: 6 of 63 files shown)


src/storage/backend/file/file-adapter.ts
ferhatelmas, line 69: due to nesting, this location would change. Is it intentional?

ferhatelmas, line 299: calculated value isn't used, wrong value here

ferhatelmas: added https://github.com/supabase/storage/issues/858 to address with a test

src/internal/monitoring/otel-class-instrumentations.ts
ferhatelmas, line 96: since methods changed, we need to update this list

src/storage/backend/s3/s3-adapter.ts
```diff
@@ -521,53 +458,47 @@ export class S3Backend implements StorageBackendAdapter {
...
     let location = key
-    let bucket = bucketName
+    let resultBucket = bucket
 
     if (opts?.removePrefix) {
       const locationParts = key.split('/')
       locationParts.shift() // tenant-id
 
-      bucket = keyParts.shift() || ''
+      resultBucket = keyParts.shift() || ''
```
ferhatelmas, line 467: We should use `locationParts` here or remove it. Using smaller scope is better though

src/http/routes/s3/commands/copy-object.ts
```diff
@@ -38,24 +38,29 @@ export default function CopyObject(s3Router: S3Router) {
...
+            | undefined,
+          Expires: req.Headers.expires ? new Date(req.Headers.expires) : undefined,
+          ContentEncoding: req.Headers['content-encoding'],
+          CopySourceIfMatch: req.Headers['x-amz-copy-source-if-match'],
+          CopySourceIfModifiedSince: req.Headers['x-amz-copy-source-if-modified-since']
+            ? new Date(req.Headers['x-amz-copy-source-if-modified-since'])
+            : undefined,
+          CopySourceIfNoneMatch: req.Headers['x-amz-copy-source-if-none-match'],
+          CopySourceIfUnmodifiedSince: req.Headers['x-amz-copy-source-if-unmodified-since']
+            ? new Date(req.Headers['x-amz-copy-source-if-unmodified-since'])
```
ferhatelmas, line 59: probably we need to track validating and short circuiting all these headers

src/storage/backend/s3/s3-adapter.ts
```diff
@@ -485,19 +427,14 @@ export class S3Backend implements StorageBackendAdapter {
...
+  async completeMultipartUpload(input: CompleteMultipartUploadInput) {
+    const { bucket, key, uploadId, version, opts } = input
+    let { parts } = input
     const keyParts = key.split('/')
 
     if (parts.length === 0) {
       const listPartsInput = new ListPartsCommand({
-        Bucket: bucketName,
+        Bucket: bucket,
         Key: version ? key + '/' + version : key,
```
ferhatelmas, line 438: Not relevant to this diff

Below we're using a helper `withOptionalVersion` but not here

ferhatelmas: also line 448

src/storage/backend/s3/s3-adapter.ts
```diff
@@ -521,53 +458,47 @@ export class S3Backend implements StorageBackendAdapter {
...
+      input
+    const uploadPartCopyCmd = new UploadPartCopyCommand({
+      Bucket: bucket,
       Key: withOptionalVersion(key, version),
-      UploadId,
-      PartNumber,
-      CopySource: `${storageS3Bucket}/${withOptionalVersion(sourceKey, sourceKeyVersion)}`,
+      UploadId: uploadId,
+      PartNumber: partNumber,
+      CopySource: `${bucket}/${withOptionalVersion(sourceKey, sourceKeyVersion)}`,
```
ferhatelmas, line 497: url encoding is missing here

src/storage/database/adapter.ts
```diff
@@ -53,6 +59,199 @@ export interface ListBucketOptions {
...
+      after?: string
+    }
+  }
+}
+
+export interface ListMultipartUploadsInput extends Cancellable {
+  bucketId: string
+  options?: {
+    prefix?: string
+    deltimeter?: string
```
ferhatelmas, line 107: delimeter

src/storage/database/adapter.ts
```diff
@@ -69,160 +268,79 @@ export interface Database {
...
-      Bucket,
-      'id' | 'name' | 'public' | 'owner' | 'file_size_limit' | 'allowed_mime_types'
-    >
-  ): Promise<Pick<Bucket, 'id'>>
+  createBucket(input: CreateBucketInput): Promise<Pick<Bucket, 'id'>>
 
-  createAnalyticsBucket(data: Pick<Bucket, 'name'>): Promise<IcebergCatalog>
+  createAnalyticsBucket(input: CreateAnalyticsBucketInput): Promise<IcebergCatalog>
 
   findBucketById<Filters extends FindBucketFilters = FindObjectFilters>(
```
ferhatelmas, line 280: isn't it supposed to be `FindBucketFilters` for default?

src/storage/backend/adapter.ts
```diff
@@ -54,156 +178,71 @@ export abstract class StorageBackendAdapter {
...
-   * @param headers
    */
-  async getObject(
-    bucketName: string,
-    key: string,
-    version: string | undefined,
-    headers?: BrowserCacheHeaders,
-    signal?: AbortSignal
-  ): Promise<ObjectResponse> {
+  async read(input: ReadObjectInput): Promise<ObjectResponse> {
```
ferhatelmas, line 189: error messages don't match method name anymore

src/storage/events/upgrades/sync-catalog-ids.ts
```diff
@@ -50,8 +50,11 @@ export class SyncCatalogIds extends UpgradeBaseEvent<SyncCatalogIdsPayload> {
...
           })
 
-          const tenantBuckets = await storage.listAnalyticsBuckets('id,name', {
-            limit: 1000,
+          const tenantBuckets = await storage.listAnalyticsBuckets({
+            columns: 'id,name',
+            options: {
+              limit: 1000,
+            },
           })
```
ferhatelmas, line 58: This needs pagination, right? For now, it might be fine but it will break at some point

src/storage/events/objects/backup-object.ts
```diff
@@ -88,15 +88,15 @@ export class BackupObjectEvent extends BaseEvent<BackupObjectEventPayload> {
...
-        await storage.backend.deleteObject(
-          storageS3Bucket,
-          storage.location.getKeyLocation({
+        await storage.backend.remove({
+          bucket: storageS3Bucket,
+          key: storage.location.getKeyLocation({
             tenantId,
             bucketId: job.data.bucketId,
             objectName: job.data.name,
           }),
```
ferhatelmas, line 97: this could be `s3Key` directly

src/http/routes/object/listObjects.ts
```diff
@@ -70,14 +70,18 @@ export default async function routes(fastify: FastifyInstance) {
...
+      const results = await request.storage.from(bucketName).searchObjects({
+        prefix,
+        options: {
+          limit,
+          offset,
+          search,
+          sortBy: {
+            column: sortBy?.column,
+            order: sortBy?.order,
+          },
```
ferhatelmas, line 82:
```suggestion
          sortBy,
```

src/storage/database/adapter.ts
```diff
@@ -33,6 +37,7 @@ export interface TransactionOptions {
   retry?: number
   readOnly?: boolean
   timeout?: number
+  signal?: AbortSignal
```
ferhatelmas, line 40: `extends Cancellable`?

src/storage/database/adapter.ts
```diff
@@ -43,6 +48,7 @@ export interface DatabaseOptions<TNX> {
   tnx?: TNX
   parentTnx?: TNX
   parentConnection?: TenantConnection
+  signal?: AbortSignal
```
ferhatelmas, line 51: this one to extend cancellable as well

src/storage/backend/s3/s3-adapter.ts
ferhatelmas, line 340: similar pattern for this method?

src/test/object.test.ts
```diff
@@ -2240,8 +2240,8 @@ describe('testing move object', () => {
       },
     })
     expect(response.statusCode).toBe(200)
-    expect(S3Backend.prototype.copyObject).toHaveBeenCalled()
-    expect(S3Backend.prototype.deleteObjects).toHaveBeenCalled()
+    expect(S3Backend.prototype.copy).toHaveBeenCalled()
+    expect(S3Backend.prototype.removeMany).toHaveBeenCalled()
```
ferhatelmas, line 2244: success is checking `removeMany` but failures below are checking for `remove` mistakenly

PR #865, opened 2026-02-19
feat: support pgbouncer for multitenant database

fenos:

Feat

No support for pooler on multitenant database

## What is the new behavior?

- Support pooler on multitenant database
- Pre-configured pgbouncer using AWS instance size calculations to determine the right configuration
- Configuring Queue to use the connection pooler
- Added optional configuration to AWS S3 Client to turn off Checksums
- Updated CI to publish and build the new image storage-otel

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -59,6 +59,15 @@ const build = (opts: buildOpts = {}): FastifyInstance => {
   app.addSchema(schemas.authSchema)
   app.addSchema(schemas.errorSchema)
 
+  // Fixes security vulnerability:
+  // https://hackerone.com/reports/3464114
+  app.addHook('onRequest', async (request, reply) => {
+    const contentType = request.headers['content-type']
+    if (contentType && contentType.includes('\t')) {
+      return reply.code(400).send({ error: 'Invalid Content-Type header' })
+    }
+  })
+
   app.register(plugins.signals)
   app.register(plugins.tenantId)
   app.register(
--- a/src/config.ts
+++ b/src/config.ts
@@ -183,6 +183,7 @@ type StorageConfigType = {
   prometheusMetricsEnabled: boolean
   prometheusMetricsIncludeTenantId: boolean
   otelMetricsEnabled: boolean
+  otelMetricsTemporality: 'DELTA' | 'CUMULATIVE'
   otelMetricsExportIntervalMs: number
   cdnPurgeEndpointURL?: string
   cdnPurgeEndpointKey?: string
@@ -458,6 +459,7 @@ export function getConfig(options?: { reload?: boolean }): StorageConfigType {
     prometheusMetricsIncludeTenantId:
       getOptionalConfigFromEnv('PROMETHEUS_METRICS_INCLUDE_TENANT') === 'true',
     otelMetricsEnabled: getOptionalConfigFromEnv('OTEL_METRICS_ENABLED') === 'true',
+    otelMetricsTemporality: getOptionalConfigFromEnv('OTEL_METRICS_TEMPORALITY') || 'CUMULATIVE',
     otelMetricsExportIntervalMs: parseInt(
       getOptionalConfigFromEnv('OTEL_METRICS_EXPORT_INTERVAL_MS') || '60000',
       10
--- a/src/internal/database/migrations/migrate.ts
+++ b/src/internal/database/migrations/migrate.ts
@@ -211,14 +211,16 @@ export async function areMigrationsUpToDate(tenantId: string) {
   )
 }
 
-export async function obtainLockOnMultitenantDB<T>(fn: () => Promise<T>) {
+export async function obtainLockOnMultitenantDB<T>(fn: (tnx: Knex.Transaction) => Promise<T>) {
+  const trx = await multitenantKnex.transaction()
   try {
-    const result = await multitenantKnex.raw(`SELECT pg_try_advisory_lock(?);`, [
-      '-8575985245963000605',
+    const result = await trx.raw(`SELECT pg_try_advisory_xact_lock(?) AS locked;`, [
+      -8575985245963000605,
     ])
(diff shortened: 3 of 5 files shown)


.docker/pgbouncer/entrypoint.sh
```diff
@@ -0,0 +1,252 @@
...
+    MIN_POOL_SIZE=$((DEFAULT_POOL_SIZE / 10))
+    RESERVE_POOL_SIZE=$((DEFAULT_POOL_SIZE / 20))
+
+    # Floor values
+    [ "$MIN_POOL_SIZE" -lt 1 ] && MIN_POOL_SIZE=1
+    [ "$RESERVE_POOL_SIZE" -lt 1 ] && RESERVE_POOL_SIZE=1
+
+    MAX_DB_CONNECTIONS=0
+
+    echo "pgbouncer: instance=${AWS_DB_INSTANCE_TYPE} memory=${memory_gb}GB rds_maxk_conn≈${rds_max_conn}"
```
ferhatelmas:
```suggestion
    echo "pgbouncer: instance=${AWS_DB_INSTANCE_TYPE} memory=${memory_gb}GB rds_max_conn=${rds_max_conn}"
```

.github/workflows/release.yml
```diff
@@ -116,6 +116,59 @@ jobs:
...
+      - name: Login to ECR account - prod
+        uses: docker/login-action@v2
+        with:
+          registry: 646182064048.dkr.ecr.us-east-1.amazonaws.com
+
+      - uses: docker/build-push-action@v3
+        with:
+          push: true
+          context: .docker/pgbouncer
+          platforms: linux/amd64
```
ferhatelmas, line 167: don't we also need arm64 since auto tuning supports it?

src/internal/queue/queue.ts
```diff
@@ -30,19 +31,24 @@ export abstract class Queue {
...
         throw new Error(
           'running storage in multi-tenant but DB_MULTITENANT_DATABASE_URL is not set'
         )
       }
-      url = multitenantDatabaseUrl
+      url = multitenantDatabasePoolUrl || multitenantDatabaseUrl
+
+      if (multitenantDatabasePoolUrl) {
+        migrate = false
+      }
```
ferhatelmas, line 46: Why is this changing? How will migration be enforced?

src/storage/backend/s3/adapter.ts
```diff
@@ -601,6 +606,11 @@ export class S3Backend implements StorageBackendAdapter {
     if (options.forcePathStyle) {
       params.forcePathStyle = true
     }
+
+    if (storageS3DisableChecksum) {
+      params.requestChecksumCalculation = 'WHEN_REQUIRED'
+      params.responseChecksumValidation = 'WHEN_REQUIRED'
+    }
```
ferhatelmas, line 613: why do we need add this change within pgbouncer?

.github/workflows/release.yml
fenos: not atm

src/internal/queue/queue.ts
fenos: we can't run migrations with pgbouncer because it might be set in transaction mode.
To run migrations, we will need to not set the pooler URL for now.

I'll think of a better way later

src/storage/backend/s3/adapter.ts
fenos: It is something I want to experiment with to disable it. It was a small change, so I put it in. Completely optional thing

.docker/pgbouncer/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY --chown=postgres:postgres pgbouncer.ini.template /etc/pgbouncer/pgbouncer.ini.template
+COPY --chown=postgres:postgres entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
+
+EXPOSE 6432
+
+USER postgres
+
+HEALTHCHECK --interval=10s --timeout=3s --start-period=5s --retries=3 \
+    CMD printf "SHOW VERSION;\n" | nc 127.0.0.1 6432 >/dev/null 2>&1 || exit 1
```
ferhatelmas, line 15: I would drop `printf` since it doesn't validate protocol anyway

ferhatelmas:
There is a code path where pgbouncer has an issue:
* when async migrations is using full fleet mode, `multitenant_knex` is used where it's trying to get `pg_try_advisory_lock` but it's not guaranteed to hit the same backend or unlock can silently fail and leak the lock
* full fleet mode isn't used according to config I see but it's possible so would be good to address (even if not right now)

.docker/pgbouncer/pgbouncer.ini.template
```diff
@@ -0,0 +1,35 @@
...
+max_db_connections = {{MAX_DB_CONNECTIONS}}
+
+server_lifetime = 3600
+server_idle_timeout = 600
+server_connect_timeout = 10
+server_login_retry = 1
+query_timeout = 0
+query_wait_timeout = 120
+client_idle_timeout = 0
+ignore_startup_parameters = extra_float_digits,statement_timeout,lock_timeout,idle_in_transaction_session_timeout,options
```
ferhatelmas, line 25: since `statement_timeout` ignored, don't we need to pass `query_timeout` to enforce `PG_QUEUE_READ_WRITE_TIMEOUT` config?

fenos: Good catch! I've now changed the queue worker to set a local statement_timeout which will work with pgBouncer transaction mode.

https://github.com/supabase/storage/commit/a2fc6c26bafdda17aca0ade53f425b71ef129f83#diff-44f8af38d4051c26681221324f8e674e9d96639e1ccc79c01354cd6a4ed230d9R43-R56

This will allows us to keep the statement_timeout as ignore option

PR #868, opened 2026-02-24
fix: handle error gracefully

fenos:

Bug fix

## What is the new behavior?

- Handle errors gracefully such as timeout errors

## Additional context

### DIFF
--- a/src/internal/queue/database.ts
+++ b/src/internal/queue/database.ts
@@ -39,6 +39,15 @@ export class QueueDB extends EventEmitter implements Db {
     }
 
     const client = await this.pool.connect()
+
+    // Create a promise that rejects if the client emits an error
+    // (e.g. connection lost, statement_timeout at the backend level)
+    let clientError: Error | undefined
+    const onError = (e: Error) => {
+      clientError = e
+    }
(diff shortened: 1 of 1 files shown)


src/internal/queue/database.ts
```diff
@@ -47,13 +56,19 @@ export class QueueDB extends EventEmitter implements Db {
...
+      if (clientError) {
+        throw clientError
+      }
+
       await client.query('COMMIT')
       return result
     } catch (err) {
       await client.query('ROLLBACK').catch(() => {})
-      throw err
+      throw clientError || err
```
ferhatelmas: better to combine them?

ferhatelmas:
```js
const rollbackErr = await client.query('ROLLBACK').catch((e) => e as Error)

const errors = [err as Error, clientError, rollbackErr].filter(
    (e): e is Error => Boolean(e)
  )

if (errors.length === 1) throw errors[0]
throw new AggregateError(errors, 'Queue transaction failed')
```

fenos: Nice!

PR #869, opened 2026-02-24
feat(pgbouncer): set application_name on server and client connections

TylerHillery:

Feature

Three things
- Application name is not set for the MT connection (was only set for the tenant db connection)
- The `application_name` for `new PgBoss` doesn't do anything because it connections to a `pg` pool. This name needs to be set on the application pooling layer
- PgBouncer wasn't setting `application_name` when connecting to postgres

## What is the new behavior?

- Properly passed in the application name when connecting to the MT database
- Removed the application name from the PgBoss configuration and change the format of the PgQueue DB to match format for the Storage API
- Added the `APP_NAME` to the `entrypoint.sh` which gets used in the `pgbouncer.ini.template`


.docker/pgbouncer/pgbouncer.ini.template
```diff
@@ -1,5 +1,5 @@
 [databases]
-{{DB_NAME}} = host={{DB_HOST}} port={{DB_PORT}} dbname={{DB_NAME}} auth_user={{DB_USER}}
+{{DB_NAME}} = host={{DB_HOST}} port={{DB_PORT}} dbname={{DB_NAME}} auth_user={{DB_USER}} application_name='{{APP_NAME}}'
```
ferhatelmas, line 2: https://www.pgbouncer.org/config.html#application_name_add_host

shall we add this to have a better visibility in db side?

.docker/pgbouncer/pgbouncer.ini.template
ferhatelmas, line 25: I thought we had added application_name into this list before but it's not here so it will be passed upstream then above suggestion isn't needed

PR #874, opened 2026-02-26
fix: Add query timeouts to multitenant db queries to prevent stuck mutexes

itslenny:

Bug fix

Under very specific conditions a query inside a mutex can get stuck and never release the mutex. This causes all subsequent calls for that tenant to get queued indefinitely.

## What is the new behavior?

- Add a timeout of 10s to all queries run within a mutex to ensure the mutex is always released
- Add timeout to other multi-tenant db queries
- Add `multitenant_pgbouncer` to docker compose
- Add `DATABASE_MULTITENANT_POOL_URL` to .env.sample to support local pgBouncer testing

## Additional context

This issue was observed in production during a resize operation on our multi tenant database. This caused several tenants to have stuck mutexes which caused the `getTenantConfig()` call to never resolve which caused any call requiring the tenant config (anything that needs to access the project database) to hang indefinitely resulting in timeouts (`ABORTED REQ`)

### Reproduction

To reproduce this locally I had to simulate a network disruption while the query was in progress. I was able to consistently reproduce this locally on all mutexes with the following process:

- Add a delay to the query (`.whereRaw('pg_sleep(0.5) IS NOT NULL')`)
- Send request to effected endpoint (e.g. `/tenants/:id/health`)
- Use `pfctl` to disrupt all traffic to database
- Send additional requests (each would queue another mutex that never releases)
- Disable `pfctl` traffic block
- Additional requests would timeout as the mutex was never released. Other requests that hit the DB but not that mutex succeed

### Testing

- Verified consistent repro on all mutex endpoints
- Verified fix with direct multi tenant db connection
- Verified fix with pgBouncer enabled (both disrupting connection to pgBouncer and to underlying postgres)
- Used `pg_sleep(30)` and `pg_stat_activity` to confirm `abortOnSignal()` cancels the query at the pg level both with direct connection and when using pgBouncer

### DIFF
--- a/.docker/docker-compose-infra.yml
+++ b/.docker/docker-compose-infra.yml
@@ -39,7 +39,7 @@ services:
     ports:
       - 6453:6432
     depends_on:
-      - tenant_db      
+      - tenant_db
     environment:
       POSTGRESQL_USERNAME: postgres
       POSTGRESQL_HOST: tenant_db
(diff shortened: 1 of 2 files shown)


src/storage/protocols/s3/credentials/store-knex.ts
```diff
@@ -38,6 +41,7 @@ export class S3CredentialsManagerStoreKnex implements S3CredentialsManagerStore
       .where('tenant_id', tenantId)
       .where('access_key', accessKey)
       .first()
+      .abortOnSignal(AbortSignal.timeout(multitenantDatabaseQueryTimeout))
```
ferhatelmas, line 46: why don't we add this timeout into other queries here?

.docker/docker-compose-infra.yml
```diff
@@ -47,6 +47,22 @@ services:
       PGBOUNCER_IGNORE_STARTUP_PARAMETERS: "extra_float_digits, options"
       PGBOUNCER_STATS_USERS: postgres
 
+  multitenant_pgbouncer:
+    image: bitnamilegacy/pgbouncer:latest
+    ports:
+      - 6454:6432
+    depends_on:
+      - multitenant_db
```
ferhatelmas, line 56: how about adding this for pgbouncer/tenant_db above?

src/config.ts
```diff
@@ -407,6 +408,10 @@ export function getConfig(options?: { reload?: boolean }): StorageConfigType {
       ),
       10
     ),
+    multitenantDatabaseQueryTimeout: parseInt(
+      getOptionalConfigFromEnv('DATABASE_MULTITENANT_QUERY_TIMEOUT') || '10000',
+      10
+    ),
```
ferhatelmas: using existing `envNumber` is a bit safer by handling malformed values

itslenny: Hmmm that's new... Looks like `envNumber` was added 2 days ago. Nice.

src/storage/protocols/s3/credentials/store-knex.ts
itslenny: I was trying to keep it pretty minimal (just the mutex queries), but there isn't really any reason any of these queries should take that long so I've added it to the rest of the queries in this file and the jwks store.

PR #890, opened 2026-03-04
fix: improve metrics performance

fenos:

Performance

- Metrics are computed too many times

## What is the new behavior?

- only calculate metrics in a controlled interval as opposed for every operation
- add admin endpoint to enable / disable metrics at run time or via env
- updated otel dependencies


src/http/plugins/metrics.ts
```diff
@@ -88,19 +88,17 @@ export const httpMetrics = (options: HttpMetricsOptions = {}) =>
...
           operation: request.operation?.type || 'unknown',
           status_code: statusCode,
-          tenantId,
+          tenantId: request.tenantId || '',
         }
 
-        // Record metrics
+        // Record duration (histogram count replaces httpRequestsTotal)
         httpRequestDuration.record(durationSeconds, attributes)
-        httpRequestsTotal.add(1, attributes)
```
ferhatelmas, line 103: drop it then?

src/internal/database/pool.ts
```diff
@@ -276,50 +300,5 @@ class TenantPool implements PoolStrategy {
...
-    pool.client.pool.on('acquireRequest', (eventId: number) => {
-      pendingAcquires.set(eventId, performance.now())
-    })
-
-    pool.client.pool.on('acquireSuccess', (eventId: number) => {
-      const startTime = pendingAcquires.get(eventId)
-      if (startTime !== undefined) {
-        pendingAcquires.delete(eventId)
-        const durationSeconds = (performance.now() - startTime) / 1000
-        dbConnectionAcquireTime.record(durationSeconds, { tenant_id: tenantId })
```
ferhatelmas, line 310: good to drop this metric as well because it will stop getting any data

src/http/plugins/metrics.ts
fenos: Yes we removed it

src/internal/database/pool.ts
fenos: yes - for now let's keep them in stand by

PR #892, opened 2026-03-04
fix: improve biome lint configuration

ferhatelmas:

Refactor

Biome configuration was bare bones.

## What is the new behavior?

Slowly add new rules which don't cause many problems or big refactoring. 

## Additional context

None

### DIFF
--- a/src/internal/streams/hash-stream.ts
+++ b/src/internal/streams/hash-stream.ts
@@ -195,15 +195,20 @@ export class HashSpillWritable extends Writable {
   }
 
   private writeToFile(buf: Buffer): Promise<void> {
-    return new Promise<void>(async (resolve, reject) => {
-      try {
-        await this.ensureFile()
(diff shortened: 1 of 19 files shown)


src/internal/streams/hash-stream.ts
```diff
@@ -195,15 +195,20 @@ export class HashSpillWritable extends Writable {
...
+          if (ok) {
+            resolve()
+            return
+          }
+
+          this.fileStream!.once('drain', resolve)
+        })
+        .catch((error) => {
+          reject(error)
+        })
```
TylerHillery, line 211: Why did this get rewritten to .then() syntax vs try catch?

ferhatelmas: Rule has a good description: https://biomejs.dev/linter/rules/no-async-promise-executor/#description

PR #900, opened 2026-03-09
feat: add storagePublicUrl for tus

aantti:

Add a new env var - `STORAGE_PUBLIC_URL` that will allow the users of [self-hosted Supabase](https://supabase.com/docs/guides/self-hosting) to avoid various complicated gymnastics around proper configuration for resumable uploads.

Current configuration options for building the `Location` header properly for TUS include `REQUEST_ALLOW_X_FORWARDED_PATH` and `TUS_URL_PATH`. (Also, one more platform-specific option).

While manageable in a default no-proxy [configuration](https://github.com/supabase/supabase/tree/master/docker), when adding a proxy on top of the API gateway (currently, Kong), it becomes quite hard to either explain concisely, or debug various edge cases that people are running into. See, e.g., issue [supabase#40686](https://github.com/supabase/supabase/issues/40686).

Arguably, it was possible to have a fairly straightforward workaround with proxying the traffic to Storage directly (e.g., see [here](https://github.com/supabase/supabase/blob/6695a9eeb0d07e210d2264f10e8415d6229829fd/docker/volumes/proxy/caddy/Caddyfile#L20)), and rely on `REQUEST_ALLOW_X_FORWARDED_PATH` to build the `Location` properly in response, with the new API keys / new asymmetric auth being added to self-hosted Supabase - this isn't going to be feasible anymore, because traffic to Storage will have to pass via the API gateway for proper [header substitution](https://github.com/supabase/supabase/pull/43554). The problem of handling various combinations of external protocol and port against those configured for Kong becomes quite painful again.

## What is the new behavior?

Add `STORAGE_PUBLIC_URL` for self-hosted Supabase and CLI only that could be configured in `.env`. Works similarly to how Studio uses `SUPABASE_PUBLIC_URL` to rewrite internal URLs for the browser - allows Storage to always construct a proper Location header without depending on `X-Forwarded-Proto`, `X-Forwarded-Host`, or `X-Forwarded-Port` headers, which Kong can overwrite with internal values when sitting behind a proxy. Does not replace `REQUEST_ALLOW_X_FORWARDED_PATH` or `TUS_URL_PATH` - works alongside them for the path prefix.


src/http/routes/tus/lifecycle.ts
```diff
@@ -113,7 +114,12 @@ export function generateUrl(
     throw ERRORS.InvalidParameter('url')
   }
 
-  proto = process.env.NODE_ENV === 'production' ? 'https' : proto
+  if (parsedPublicUrl) {
+    proto = parsedPublicUrl.protocol.replace(':', '')
```
itslenny, line 117: I wonder if `process.env.NODE_ENV === 'production'` should superced `parsedPublicUrl`. I feel like we should maybe still force https for "production"

aantti: Moved it out to supercede.

ferhatelmas: Right now, we're overriding silently. It can be a bit confusing.

Documentation can help for clarification but I think it might be better to fail with an explicit error message after parsing the URL if not https in NODE_ENV=production.

PR #902, opened 2026-03-10
fix: improve performance, upgraded otel fastify

fenos:

Improvement

## What is the new behavior?

- Upgraded deprecated otel library `@opentelemetry/instrumentation-fastify` to `https://www.npmjs.com/package/@fastify/otel` 
- Removed duplicates metrics 
- Improve allocations and performance on logs filters
- Use batch observe to monitor pool metrics


src/internal/monitoring/otel-tracing.ts
```diff
@@ -81,11 +84,21 @@ if (tracingEnabled && traceExporter && spanProcessors.length > 0) {
...
+      // It auto-sets http.route, http.request.method, url.path on spans.
+      // Other attributes (tenant.ref, trace.mode) are set in Fastify hooks
+      // via trace.getActiveSpan() (see tracing.ts, log-request.ts).
+      new FastifyOtelInstrumentation({
+        enabled: true,
+        registerOnInitialization: true,
+        ignorePaths: (routeOpts) => {
+          return ignoreRoutes.includes(routeOpts.url)
+        },
+      }),
```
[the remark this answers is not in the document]
ferhatelmas: this is a valid problem: https://github.com/fastify/otel?tab=readme-ov-file#fastifyotelinstrumentationoptionsrequesthook-function

src/http/plugins/tracing.ts
```diff
@@ -26,6 +27,16 @@ export const tracing = fastifyPlugin(
         } else {
           request.tracingMode = defaultTracingMode
         }
+
+        const span = trace.getActiveSpan()
```
ferhatelmas: there is a [comment below from coderabbit](https://github.com/supabase/storage/pull/902/changes#r2911006882) 

in short, this is the span for the hook itself, not the root request span so it won't propagate

src/internal/monitoring/metrics.ts
```diff
@@ -240,11 +216,45 @@ export const httpPoolErrors = withMetricLabels(
...
+    name,
+    type: 'gauge',
+    enabled: !disabledMetrics.has(name),
+  }
+  metricsRegistry.set(name, entry)
+
+  return {
+    instrument: meter.createObservableGauge(name, options),
+    entry,
+  }
```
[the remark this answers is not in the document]
ferhatelmas: this is valid as well

src/internal/monitoring/otel-metrics.ts
```diff
@@ -210,6 +212,18 @@ if (otelMetricsEnabled) {
         instrumentName: 'db_connection_acquire_seconds',
         aggregation: histogramAggregation,
       },
+      // Drop duplicate HTTP metrics from auto-instrumentations — we have our own in metrics.ts
```
ferhatelmas: just noting, we weren't using so it's fine but this will drop outbound http client metrics as well

src/internal/monitoring/logger.ts
```diff
@@ -165,70 +165,74 @@ export function buildTransport(): pino.TransportMultiOptions {
...
+  const qIdx = url.indexOf('?')
+
+  // Fast path: no query string, nothing to redact
+  if (qIdx === -1) return url
+
+  const query = url.slice(qIdx + 1)
+  // Fast path: no sensitive params present
+  if (!params.some((p) => query.includes(p))) return url
 
+  const lUrl = new URL(url, `${req.protocol}://${req.hostname}`)
```
ferhatelmas, line 235: changes in this file is a nice boost 👍🏻

monitoring/grafana/dashboards/storage-otel.json
```diff
@@ -19,7 +19,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
-  "id": 2,
```
ferhatelmas, line 22: commenting here but how about dropping the other dashboard (storage.json)?

src/internal/database/pool.ts
```diff
@@ -72,22 +85,108 @@ const tenantPools = new TTLCache<string, PoolStrategy>({
...
+let cachedPoolStats: PoolStatsSnapshot = {
+  poolCount: 0,
+  totalConnections: 0,
+  totalInUse: 0,
+  perTenant: [],
+}
+let collectInProgress = false
+
+async function collectPoolStats() {
+  if (collectInProgress) return
```
ferhatelmas, line 107: is it worth of tracking how much stale?

src/internal/database/pool.ts
```diff
@@ -72,22 +85,108 @@ const tenantPools = new TTLCache<string, PoolStrategy>({
...
+  collectInProgress = true
+
+  try {
+    const entries = [...tenantPools.entries()]
```
ferhatelmas: There are some conditions below where a pool wasn't added into the cache. Missing them is intentional/insignificant because they are short-lived?

fenos: what do you mean?

src/internal/database/pool.ts
fenos: just a small precaution

monitoring/grafana/dashboards/storage-otel.json
fenos: Will drop in another PR

src/internal/database/pool.ts
ferhatelmas: was referring to [this condition](https://github.com/supabase/storage/pull/902/changes#diff-da48fb498d52277cba8f3ee946a0dfd2f1f6c893c0cc5c10e119be2c5c366820R219)

src/internal/monitoring/metrics.ts
```diff
@@ -44,41 +38,74 @@ export function setMetricsEnabled(changes: { name: string; enabled: boolean }[])
...
+    meter.createHistogram('http_request_duration_seconds', {
+      description: 'HTTP request duration in seconds',
+      unit: 's',
+    })
 )
 
-export const httpRequestsTotal = withMetricLabels(
-  'http_requests_total',
-  'counter',
+export const httpRequestsTotal = registerMetric('http_requests_total', 'counter', () =>
```
ferhatelmas: good to drop at this point?

monitoring/grafana/dashboards/storage-otel.json
```diff
@@ -1162,35 +1162,43 @@
             "type": "prometheus",
             "uid": "local_prometheus"
           },
-          "expr": "sum(rate(storage_api_otel_http_requests_total{region=~\"$region\", instance=~\"$instance\"}[$__rate_interval]))",
```
ferhatelmas, line 1165: related to https://github.com/supabase/storage/pull/902#discussion_r2919098895

src/internal/monitoring/metrics.ts
```diff
@@ -44,41 +38,74 @@ export function setMetricsEnabled(changes: { name: string; enabled: boolean }[])
...
+
+// ============================================================================
+// Meter & registration
+// ============================================================================
+export const meter = metrics.getMeter('storage-api')
+
+function stripTenantAttrs(attrs: Attributes): Attributes {
+  const { tenantId, tenant_id, ...rest } = attrs as Record<string, unknown>
+  return rest as Attributes
+}
```
ferhatelmas, line 54: is it worth of checking the config and not adding in the first place?

src/http/plugins/metrics.ts
```diff
@@ -90,7 +90,6 @@ export const httpMetrics = (options: HttpMetricsOptions = {}) =>
 
         const attributes = {
           method,
-          route,
```
ferhatelmas, line 93: what is the reason? cardinality + covered by operation?

sounds good but there are still references to it in dashboard though

src/internal/monitoring/metrics.ts
fenos: yes, but it would be too manny if in too many places

src/http/plugins/metrics.ts
fenos: Yes, cardinality, we do use the operation only, i don't think we reference the route

ferhatelmas: Top 10 panels refer it

src/internal/monitoring/metrics.ts
ferhatelmas: yes, it will be a bit messier since will clutter usage. We can do it as a follow up (if needed)

PR #903, opened 2026-03-11
feat: add RLS operation ergonomics

ferhatelmas:

Feature

Users need to write raw comparisons which can be brittle.

## What is the new behavior?

A bit higher level helpers to ease writing a policy. 

## Additional context

* single and any exact case-sensitive match
* storage prefix is optional
* moved legacy operations to operations.ts and linked to docs for easier discoverability
* docs: https://github.com/supabase/supabase/pull/43859

### DIFF
--- a/src/http/routes/operations.ts
+++ b/src/http/routes/operations.ts
@@ -1,3 +1,6 @@
+// This file defines all the operation names used in the application for better maintainability and to avoid typos.
+// This file is linked from the docs to provide a list of all available operations.
+// Any change to this file should be reflected in the docs if applicable.
 export const ROUTE_OPERATIONS = {
   // Bucket
   CREATE_BUCKET: 'storage.bucket.create',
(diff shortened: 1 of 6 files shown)


src/http/routes/operations.ts
```diff
@@ -14,8 +17,10 @@ export const ROUTE_OPERATIONS = {
   DELETE_OBJECTS: 'storage.object.delete_many',
   GET_PUBLIC_OBJECT: 'storage.object.get_public',
   GET_AUTH_OBJECT: 'storage.object.get_authenticated',
-  INFO_AUTH_OBJECT: 'storage.object.info_authenticated',
+  INFO_AUTH_OBJECT: 'storage.object.info_authenticated', // not used
   INFO_PUBLIC_OBJECT: 'storage.object.info_public',
+  GET_AUTH_OBJECT_INFO: 'object.get_authenticated_info', // legacy
+  HEAD_AUTH_OBJECT_INFO: 'object.head_authenticated_info', // legacy
```
ferhatelmas, line 23: @fenos these are the two operations that don't follow others

PR #905, opened 2026-03-12
fix: stream upload to S3

fenos:

Improvement

Currently, we upload to S3 using the `Upload` class form the `@aws-sdk/lib-storage`. 
This library seems to be buffering in memory each chunk before uploading them to S3, (that's because it needs to calculate the size of the stream) 

## What is the new behavior?

We do know the content length in most cases, so we can simply pipe the body into the `PutObject` method so that we stream it all at once

### DIFF
--- a/src/test/uploader.test.ts
+++ b/src/test/uploader.test.ts
@@ -4,6 +4,47 @@ import { ErrorCode, isStorageError } from '../internal/errors'
 import { fileUploadFromRequest } from '../storage/uploader'
 
 describe('fileUploadFromRequest', () => {
+  test('keeps multipart/form-data file size undefined even when the request content-length exceeds 5GB', async () => {
+    const file = Readable.from(['payload']) as Readable & { truncated: boolean }
+    file.truncated = false
+
+    const requestFile = jest.fn().mockResolvedValue({
+      file,
+      fields: {
+        cacheControl: { value: '3600' },
+        contentType: { value: 'image/png' },
+        metadata: { value: '{"source":"multipart"}' },
+      },
+      mimetype: 'application/octet-stream',
+    })
+
+    const upload = await fileUploadFromRequest(
+      {
+        headers: {
+          'content-type': 'multipart/form-data; boundary=abc123',
+          'content-length': String(5 * 1024 * 1024 * 1024 + 512),
+        },
+        file: requestFile,
+        tenantId: 'stub-tenant',
+      } as unknown as FastifyRequest,
+      {
+        objectName: 'test.txt',
+        fileSizeLimit: 150,
+      }
+    )
+
+    expect(requestFile).toHaveBeenCalledWith({ limits: { fileSize: 150 } })
+    expect(upload.body).toBe(file)
+    expect(upload.contentLength).toBeUndefined()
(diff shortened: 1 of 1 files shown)


src/storage/backend/s3/adapter.ts
```diff
@@ -141,12 +146,94 @@ export class S3Backend implements StorageBackendAdapter {
...
     cacheControl: string,
-    signal?: AbortSignal
+    signal?: AbortSignal,
+    contentLength?: number
   ): Promise<ObjectMetadata> {
     if (signal?.aborted) {
       throw ERRORS.Aborted('Upload was aborted')
     }
 
+    if (typeof contentLength !== 'number' || contentLength > MAX_PUT_OBJECT_SIZE) {
```
fenos, line 156: we might want to strengthen the condition here with: 

```js
if (
       typeof contentLength !== 'number' || 
       (typeof contentLength === 'number' && contentLength > MAX_PUT_OBJECT_SIZE)
) {
  ...
}

```

ferhatelmas: isn't it same?

fenos: After a second thought this might not be required, can ignore

src/storage/backend/s3/adapter.ts
```diff
@@ -141,12 +146,94 @@ export class S3Backend implements StorageBackendAdapter {
...
+        key,
+        version,
+        body,
+        contentType,
+        cacheControl,
+        signal
+      )
+    }
+
+    // Use PutObject directly when content-length is known and within S3's single-object limit (5GB).
```
itslenny, line 169: What about a `multipart/form-data` upload >5GB ? The logic in `storage/uploader.ts` sets it to undefined intentionally, but if the putObject limit is 5GB won't that cause a problem if the file is larger?

src/storage/uploader.ts
```diff
@@ -357,6 +360,8 @@ export async function fileUploadFromRequest(
 
       const file = formData.file
       body = file
+      // multipart/form-data content-length includes boundary overhead and cannot be trusted as file size,
+      // so we intentionally leave fileContentLength undefined and let the backend stream via multipart upload.
```
itslenny, line 364: Related to other comment... Won't this result in always calling `this.putObject()` in `s3/adapter.ts` even for files > 5GB ?

src/storage/backend/s3/adapter.ts
ferhatelmas: It will go for buffered multipart approach if unknown or >5GB so it should hit the first condition in the previous check

fenos: @itslenny we don't support `multipart/form-data` over 5GB

ferhatelmas: pushed a unit test to confirm the condition

src/storage/uploader.ts
fenos: mmmh this is unfortunate though :( i'm wondering if there is no way to get the size reliably, but i guess not

ferhatelmas: With buffering, yes but it would defeat the purpose here

PR #917, opened 2026-03-18
feat: support for watt process runner

fenos:

Feature

## What is the new behavior?

- Add Watt support for starting the server in a monitored and multi-CPU system
- Corrected shutdown order and prevented error messages on shutdown

## Additional context

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,17 +92,17 @@
     "@types/busboy": "^1.3.0",
     "@types/cloneable-readable": "^2.0.3",
     "@types/fs-extra": "^9.0.13",
-    "@types/jest": "^29.5.14",
+    "@types/jest": "^30.0.0",
     "@types/js-yaml": "^4.0.5",
     "@types/json-bigint": "^1.0.4",
     "@types/node": "^22.19.0",
     "@types/pg": "^8.6.4",
     "@types/stream-buffers": "^3.0.7",
     "@types/xml2js": "^0.4.14",
-    "babel-jest": "^29.7.0",
+    "babel-jest": "^30.3.0",
     "esbuild": "^0.25.8",
     "form-data": "^4.0.0",
-    "jest": "^29.7.0",
+    "jest": "^30.3.0",
     "js-yaml": "^4.1.0",
     "json-schema-to-ts": "^3.0.0",
     "patch-package": "^8.0.0",
@@ -110,11 +110,10 @@
     "prettier": "^2.8.8",
     "resolve-tspaths": "^0.8.19",
     "stream-buffers": "^3.0.2",
-    "ts-jest": "^29.3.2",
-    "ts-node-dev": "^1.1.8",
+    "ts-jest": "^29.4.6",
     "tsx": "^4.16.0",
     "tus-js-client": "^3.1.0",
-    "typescript": "5.2.2"
+    "typescript": "5.9.3"
   },
   "version": "1.11.2",
   "description": "Supabase Storage Service",
@@ -130,9 +129,6 @@
     },
     "glob@7": {
       "minimatch": "^3.1.4"
-    },
-    "ts-node@9": {
-      "diff": "^4.0.4"
     }
   },
   "bin": "./dist/server.js",
--- a/migrations/tenant/0057-s3-multipart-uploads-metadata.sql
+++ b/migrations/tenant/0057-s3-multipart-uploads-metadata.sql
@@ -1,0 +1,1 @@
+ALTER TABLE storage.s3_multipart_uploads ADD COLUMN IF NOT EXISTS metadata jsonb NULL;
--- a/package-lock.json
+++ b/package-lock.json
@@ -83,17 +83,17 @@
         "@types/busboy": "^1.3.0",
         "@types/cloneable-readable": "^2.0.3",
         "@types/fs-extra": "^9.0.13",
-        "@types/jest": "^29.5.14",
+        "@types/jest": "^30.0.0",
         "@types/js-yaml": "^4.0.5",
         "@types/json-bigint": "^1.0.4",
         "@types/node": "^22.19.0",
         "@types/pg": "^8.6.4",
         "@types/stream-buffers": "^3.0.7",
         "@types/xml2js": "^0.4.14",
-        "babel-jest": "^29.7.0",
+        "babel-jest": "^30.3.0",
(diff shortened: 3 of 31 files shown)


package.json
```diff
@@ -77,14 +71,16 @@
     "jose": "^6.0.10",
     "json-bigint": "^1.0.0",
     "knex": "^3.1.0",
-    "lru-cache": "^11.2.0",
+    "lru-cache": "^10.2.0",
```
ferhatelmas: this is bumped to v11 on master and now reverting it, can we rebase?

src/start/server.ts
```diff
@@ -135,7 +138,7 @@ async function main() {
       `[Cluster] Cluster size changed to ${data.size}`
     )
     TenantConnection.poolManager.rebalanceAll({
-      clusterSize: data.size,
+      clusterSize: Math.max(Math.floor(data.size / effectiveNumWorkers), 1),
```
ferhatelmas: instead of divide, shouldn't we increase (multiply)? tasks vs tasks * workers

ferhatelmas: and related, what happens if there is no change event is triggered?

src/start/server.ts
```diff
@@ -135,7 +138,7 @@ async function main() {
       `[Cluster] Cluster size changed to ${data.size}`
     )
     TenantConnection.poolManager.rebalanceAll({
-      clusterSize: data.size,
+      clusterSize: Math.max(Math.floor(data.size * effectiveNumWorkers), 1),
```
ferhatelmas: could be `data.size * effectiveWorkers`

src/start/server.ts
```diff
@@ -135,7 +136,7 @@ async function main() {
       `[Cluster] Cluster size changed to ${data.size}`
     )
     TenantConnection.poolManager.rebalanceAll({
-      clusterSize: data.size,
+      clusterSize: Math.max(Math.floor(data.size / numWorkers), 1),
     })
```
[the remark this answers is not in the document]
ferhatelmas: not valid

Dockerfile
```diff
@@ -34,6 +34,8 @@ COPY migrations migrations
 COPY --from=production-deps /app/node_modules node_modules
 # Copy build artifacts from the build stage
 COPY --from=build /app/dist dist
+COPY --from=build /app/watt.json /app
+
 
 EXPOSE 5000
 CMD ["node", "dist/start/server.js"]
```
ferhatelmas, line 41: update this as well to in local, CLI etc to use without override ?

src/start/server.ts
ferhatelmas: I think we should put this computation inside Cluster so that wherever Cluster.size is referenced, no logic changes

fenos: Yes, I'm now passing numWorkers to the cluster directly.

Dockerfile
fenos: not sure i understand what you mean.
Watt is completely optional setup here

ferhatelmas: Thinking it would be default. If optional, that's fine.

src/start/server.ts
ferhatelmas: looks good now

src/internal/cluster/cluster.ts
```diff
@@ -23,7 +30,8 @@ export class Cluster {
     }
 
     if (cluster) {
-      Cluster.size = await cluster.getClusterSize()
+      const rawSize = await cluster.getClusterSize()
+      Cluster.size = Cluster.effectiveSize(rawSize)
```
ferhatelmas: we don't set this when no cluster so if workers are used, division per worker won't happen in disabled cluster discovery

src/start/shutdown.ts
```diff
@@ -18,23 +18,36 @@ export function bindShutdownSignals(serverSignal: AsyncAbortController) {
...
+    logSchema.info(logger, `[Server] Received ${signal}, shutting down`, {
       type: 'shutdown',
     })
     try {
       await shutdown(serverSignal)
-      logSchema.info(logger, '[Server] SIGTERM Shutdown successfully', {
+      logSchema.info(logger, `[Server] ${signal} Shutdown successfully`, {
         type: 'shutdown',
       })
+      process.exit(0)
```
ferhatelmas, line 39: why do we prefer this over `process.exitCode = 0` ?

fenos: when using `process.exitCode = 0;` didn't seem to exit properly

src/internal/cluster/cluster.ts
fenos: Agree! good catch!
I moved this logic even lower level in the pool. Which would work regardless of whether it is clustered or not.

src/start/shutdown.ts
ferhatelmas: I think we should add metric/tracing into shutdown. I already added them into global

```ts
type OTelGlobalState = typeof globalThis & {
  __otelMetricsShutdown?: () => Promise<void>
  __otelTracingShutdown?: () => Promise<void>
}

await Promise.allSettled([
  otel.__otelMetricsShutdown?.(),
  otel.__otelTracingShutdown?.(),
])
```

watt.json
```diff
@@ -0,0 +1,25 @@
...
+      "enabled": "{WATT_HEALTH_ENABLED}",
+      "maxELU": "{WATT_HEALTH_MAX_ELU}",
+      "gracePeriod": "{WATT_HEALTH_GRACE_PERIOD}",
+      "maxUnhealthyChecks": "{WATT_HEALTH_MAX_UNHEALTHY_CHECKS}",
+      "interval": "{WATT_HEALTH_INTERVAL}"
+    },
+    "server": {
+      "hostname": "{SERVER_HOST}",
+      "port": "{SERVER_PORT}"
+    },
```
ferhatelmas: in multitenant, we're listening on two different ports, wondering what will happen to admin

src/start/shutdown.ts
fenos: We can do this in subsequent PR

watt.json
fenos: Yeah! That is true!

fenos: Confirmed is no longer needed the server key in watt.json in the latest version

PR #921, opened 2026-03-19
fix: validate response headers before sending, prevent invalid transform input

itslenny:

Bug fix

- transform query parameters on HEAD and info routes are not validated these values are passed to the x-transformation headers as-is allowing invalid input to be passed
- if invalid characters are sent in response headers node http module throws `ERR_INVALID_CHAR` which is an uncaught exception resulting in a process crash

## What is the new behavior?

Validate response headers before sending and throw an http error instead of allowing node to crash. This acts as a backstop for any invalid response header type errors.

Add missing query parameters validation for transform operations to prevent the invalid input from being set in the first place

## Additional context

This issue could be triggered via `format`, `resize`, and `quality` query params on all of the following endpoints:

- `HEAD /object/public/bucket-name/file-name.jpg`
- `HEAD /object/authenticated/bucket-name/file-name.jpg`
- `HEAD/object/bucket-name/file-name.jpg`
- `GET /object/info/public/bucket-name/file-name.jpg`
- `GET /object/info/authenticated/bucket-name/file-name.jpg`
- `GET /object/info/bucket-name/file-name.jpg`

---

Without header validation any invalid chars in a header results in an uncaught exception

Screenshot 2026-03-19 at 11 06 53 AM

---

With header validation we handle the characters gracefully and log the offending header

Screenshot 2026-03-19 at 11 06 17 AM

---

With query param validation this is handled earlier by fastify validation (FST_ERR_VALIDATION)

Screenshot 2026-03-19 at 11 53 46 AM

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -69,7 +69,7 @@ const build = (opts: buildOpts = {}): FastifyInstance => {
   )
   app.register(plugins.tracing)
   app.register(plugins.logRequest({ excludeUrls: excludedRoutesFromMonitoring }))
-  app.register(plugins.headerValidator)
+  app.register(plugins.headerValidator({ excludeUrls: excludedRoutesFromMonitoring }))
   app.register(routes.tus, { prefix: 'upload/resumable' })
   app.register(routes.bucket, { prefix: 'bucket' })
   app.register(routes.object, { prefix: 'object' })
--- a/src/http/plugins/header-validator.ts
+++ b/src/http/plugins/header-validator.ts
@@ -10,25 +10,39 @@ import fastifyPlugin from 'fastify-plugin'
  */
 const INVALID_HEADER_CHAR_PATTERN = /[^\t\x20-\x7e\x80-\xff]/
 
+interface HeaderValidatorOptions {
+  excludeUrls?: string[]
+}
+
 /**
  * Validates response headers before they're sent to prevent ERR_INVALID_CHAR crashes.
  *
  * Node.js throws ERR_INVALID_CHAR during writeHead() if headers contain control characters.
  * This hook validates headers in onSend (before writeHead) and throws InvalidHeaderChar error
  */
-export const headerValidator = fastifyPlugin(
-  async function headerValidatorPlugin(fastify: FastifyInstance) {
-    fastify.addHook('onSend', async (_request: FastifyRequest, reply: FastifyReply, payload) => {
-      const headers = reply.getHeaders()
-
-      for (const [key, value] of Object.entries(headers)) {
-        if (typeof value === 'string' && INVALID_HEADER_CHAR_PATTERN.test(value)) {
-          throw ERRORS.InvalidHeaderChar(key, value)
+export const headerValidator = (options: HeaderValidatorOptions = {}) =>
+  fastifyPlugin(
+    async function headerValidatorPlugin(fastify: FastifyInstance) {
+      fastify.addHook('onSend', async (request: FastifyRequest, reply: FastifyReply, payload) => {
+        if (options.excludeUrls?.includes(request.url.toLowerCase())) {
+          return payload
         }
-      }
 
-      return payload
-    })
-  },
-  { name: 'header-validator' }
-)
+        const headers = Object.entries(reply.getHeaders())
+        for (const [key, value] of headers) {
+          if (typeof value === 'string' && INVALID_HEADER_CHAR_PATTERN.test(value)) {
+            throw ERRORS.InvalidHeaderChar(key, value)
+          } else if (Array.isArray(value)) {
+            for (let item of value) {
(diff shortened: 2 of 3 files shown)


src/http/plugins/header-validator.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Node.js throws ERR_INVALID_CHAR during writeHead() if headers contain control characters.
+ * This hook validates headers in onSend (before writeHead) and throws InvalidHeaderChar error
+ */
+export const headerValidator = fastifyPlugin(
+  async function headerValidatorPlugin(fastify: FastifyInstance) {
+    fastify.addHook('onSend', async (_request: FastifyRequest, reply: FastifyReply, payload) => {
+      const headers = reply.getHeaders()
+
+      for (const [key, value] of Object.entries(headers)) {
+        if (typeof value === 'string' && INVALID_HEADER_CHAR_PATTERN.test(value)) {
```
ferhatelmas: `reply.getHeaders()` type includes string array. 

I think we don't use but shall we add coverage since this is generic

src/http/plugins/header-validator.ts
```diff
@@ -0,0 +1,34 @@
...
+          throw ERRORS.InvalidHeaderChar(key, value)
```
ferhatelmas: do we also need to remove that header from reply object?

src/internal/errors/codes.ts
```diff
@@ -260,6 +260,14 @@ export const ERRORS = {
       message: `Invalid X-Robots-Tag header: ${message}`,
     }),
 
+  InvalidHeaderChar: (headerName: string, headerValue: string) =>
+    new StorageBackendError({
+      error: 'invalid_header_char',
+      code: ErrorCode.InvalidRequest,
+      httpStatusCode: 400,
+      message: `Invalid character in response header "${headerName}": ${headerValue.substring(0, 50)}`,
```
ferhatelmas, line 268: Just thinking loud, is it a good idea to the drop value? 

It's truncated and header name would be enough to debug, I guess

src/test/validators.test.ts
```diff
@@ -1,5 +1,74 @@
...
+    await app.close()
+  })
+
+  it('should reject response with newline in header value', async () => {
+    app.get('/test', async (_request, reply) => {
+      reply.header('x-test', 'value\nwith\nnewlines')
+      return { ok: true }
+    })
+
+    const response = await app.inject({ method: 'GET', url: '/test' })
```
ferhatelmas, line 25: This simulates in memory, right? 

Because I think we might not be testing actual Node HTTP stack where the crash (writeHead) is happening

src/app.ts
```diff
@@ -69,6 +69,7 @@ const build = (opts: buildOpts = {}): FastifyInstance => {
   )
   app.register(plugins.tracing)
   app.register(plugins.logRequest({ excludeUrls: excludedRoutesFromMonitoring }))
+  app.register(plugins.headerValidator)
```
ferhatelmas: Probably negligible and fine, I would use `excludedRoutesFromMonitoring` to skip this.

Aside we should probably switch it into a set that also contains `/` variants

src/http/plugins/header-validator.ts
```diff
@@ -0,0 +1,34 @@
...
+import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify'
+import fastifyPlugin from 'fastify-plugin'
+
+/**
+ * Matches invalid HTTP header characters per RFC 7230 field-vchar specification.
+ * Valid: TAB (0x09), visible ASCII (0x20-0x7E), obs-text (0x80-0xFF).
+ * Invalid: control characters (0x00-0x1F except TAB) and DEL (0x7F).
+ * @see https://tools.ietf.org/html/rfc7230#section-3.2
+ */
+const INVALID_HEADER_CHAR_PATTERN = /[^\t\x20-\x7e\x80-\xff]/
```
ferhatelmas, line 11: Btw, we have this regex for [s3](https://github.com/supabase/storage/blob/003d5f5dfb1af77a71c6a02663cea9025f54101f/src/storage/protocols/s3/s3-handler.ts#L1361)

src/http/plugins/header-validator.ts
itslenny: No. We throw so the reply object isn't sent. It's also nice because it is still included in the `res` part of the log (see screenshot above)

src/internal/errors/codes.ts
itslenny: The full value is still included in the log under the `res` key. This is just what is sent back as the response to the caller which doesn't need to be the full value.

src/test/validators.test.ts
itslenny: It doesn't really matter if we hit the uncaught exception path as the test expectation ensures that the invalid headers results in the correct error type (which confirms that we throw before sending)

PR #926, opened 2026-03-25
chore: replace usage of PAT with GH app

staaldraad:

chore

Pins the github actions.
Replaces usage of a PAT with a dedicated GH app
Safes template strings in shell commands

### DIFF
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -50,6 +50,8 @@ jobs:
           git checkout -b "${{ env.service_to_update }}/${GH_REF}"
 
       - name: Update file
+        env:
+          GH_REF: ${{ github.ref_name }}
         run: |
           cd '${{ env.cli_repo }}'
           sed -i "s|${{ env.cli_repo_owner }}/${{ env.service_to_update }}:v[0-9]*\.[0-9]*\.[0-9]*|${{ env.cli_repo_owner }}/${{ env.service_to_update }}:${GH_REF}|" internal/utils/misc.go


.github/workflows/cli.yaml
```diff
@@ -30,33 +39,40 @@ jobs:
...
+          token: ${{ steps.app-token.outputs.token }}
           path: "${{ env.cli_repo }}"
 
       - name: Create a new branch
+        env:
+          GH_REF: ${{ github.ref_name }}
         run: |
           cd '${{ env.cli_repo }}'
-          git checkout -b '${{ env.service_to_update }}/${{ github.ref_name }}'
+          git checkout -b '${{ env.service_to_update }}/${GH_REF}'
```
ferhatelmas: is this expanding under single quotes?

aside, what is wrong with github expression and we use shell variable now?

staaldraad: good catch! 
Best practice is to not do any template expansion in shell commands. `ref_name` can be user controlled as it comes from the branch name, this could be dangerous

.github/workflows/cli.yaml
```diff
@@ -30,33 +39,40 @@ jobs:
...
         run: |
           cd '${{ env.cli_repo }}'
-          git checkout -b '${{ env.service_to_update }}/${{ github.ref_name }}'
+          git checkout -b "${{ env.service_to_update }}/${GH_REF}"
 
       - name: Update file
         run: |
           cd '${{ env.cli_repo }}'
-          sed -i 's/${{ env.cli_repo_owner }}\/${{ env.service_to_update }}:v[0-9]*\.[0-9]*\.[0-9]*/${{ env.cli_repo_owner }}\/${{ env.service_to_update }}:${{ github.ref_name }}/' internal/utils/misc.go
+          sed -i "s|${{ env.cli_repo_owner }}/${{ env.service_to_update }}:v[0-9]*\.[0-9]*\.[0-9]*|${{ env.cli_repo_owner }}/${{ env.service_to_update }}:${GH_REF}|" internal/utils/misc.go
```
ferhatelmas, line 57: it's not defined for this step

staaldraad: ah thanks man, sharp eyes. Sorry for the incremental updates, was a long day. I should have caught these locally

PR #946, opened 2026-03-27
fix: use parameterized queries in upgrade-v10.ts

### LINKED ISSUE #945: fix: unparameterized SQL in upgrade-v10.ts (missed by #940)
## Description

PR #940 fixed template literal interpolation in `move-jobs.ts` by switching to parameterized queries, but the same pattern exists in `upgrade-v10.ts` and was not included in that fix.

In `src/storage/events/pgboss/upgrade-v10.ts`, `queue.name` and `queue.policy` are interpolated directly into raw SQL:

```ts
WHERE name = '${queue.name}'
'${queue.policy}' as policy
```

And the query runs via `multitenantKnex.raw(sql)` instead of `tnx.raw(sql, [...])`, which also means it executes outside the transaction that holds the advisory lock.

## Expected behavior

Should use parameterized queries and run inside the transaction, same as the fix applied to `move-jobs.ts` in #940:

```ts
WHERE name = ?
? as policy

await tnx.raw(sql, [queue.name, queue.policy])
```

oniani1:

Bug fix (follow-up to #940)

`upgrade-v10.ts` interpolates `queue.name` and `queue.policy` directly into raw SQL via template literals:

```ts
'${queue.policy}' as policy
...
WHERE name = '${queue.name}'
```

And executes the query with `multitenantKnex.raw(sql)` outside the transaction that holds the advisory lock.

#940 fixed this exact pattern in `move-jobs.ts` but this file was not included.

Resolves #945

## What is the new behavior?

Uses `?` placeholders with bound parameters and runs the query inside the transaction via `tnx.raw()`, matching the pattern applied in #940:

```ts
? as policy
...
WHERE name = ?

await tnx.raw(sql, [queue.policy, queue.name])
```

### DIFF
--- a/src/storage/events/pgboss/upgrade-v10.ts
+++ b/src/storage/events/pgboss/upgrade-v10.ts
@@ -84,14 +84,14 @@ export class UpgradePgBossV10 extends BaseEvent<UpgradePgBossV10Payload> {
                 createdOn,
                 keepUntil,
                 output jsonb,
-                '${queue.policy}' as policy
+                ? as policy
             FROM ${sourceSchema}.job
-            WHERE name = '${queue.name}'
+            WHERE name = ?
                 AND state = 'created'
             ON CONFLICT DO NOTHING
         `
 
-          await multitenantKnex.raw(sql)
+          await tnx.raw(sql, [queue.policy, queue.name])
         } catch (error) {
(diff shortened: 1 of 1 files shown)


ferhatelmas (approved):
Nice catch! Thanks for your contribution 🙏🏻

PR #950, opened 2026-03-30
fix: drop fs-extra for native fs/promises

ferhatelmas:

Refactor

fs-extra is used for promise based fs API.

## What is the new behavior?

Use native fs/promises and drop fs-extra.

## Additional context

[Docs](https://github.com/es-tooling/module-replacements/blob/main/docs/modules/fs-extra.md)


src/internal/fs.ts
```diff
@@ -0,0 +1,27 @@
+import { promises as fs } from 'node:fs'
```
TylerHillery: general question: out of curiosity why did you do `import {promises as fs} from 'node:fs'` here but else where you did `from node:fs/promises`

ferhatelmas: lint:fix vs autoimport in vscode, let me align

PR #952, opened 2026-03-30
fix: return user_metadata in list/search responses

### LINKED ISSUE #759: storage.list() does not return custom file metadata (only system fields)
# Bug report

## Describe the bug

When uploading a file with custom metadata (e.g. `metadata: { filename: "My Real Filename.txt" }`), the `.list()` method only returns system metadata fields (`eTag`, `size`, `mimetype`).  
Custom metadata is not included, even though it is stored correctly in `storage.objects` (and retrievable via `.info()`).

## To Reproduce

Steps to reproduce:

1. Upload a file with custom metadata:
   ```js
   const { data, error } = await supabase.storage
     .from("bucket")
     .upload("path/file.txt", file, {
       metadata: { filename: "My Real Filename.txt" },
     });
   ```

2. List files:
   ```js
   const { data } = await supabase.storage.from("bucket").list("path");
   console.log(data[0].metadata); // only shows system fields, no custom filename
   ```

3. Fetch file info:
   ```js
   const info = await supabase.storage.from("bucket").info("path/file.txt");
   console.log(info.data.metadata); // shows custom fields as expected
   ```

## Expected behavior

`storage.list()` should return both system and custom metadata, not just system fields.  
This would avoid needing `N` `.info()` calls for `N` files, which has a significant performance impact.

## System information

- OS: macOS
- Browser: Chrome
- supabase-js: 2.50.0
- Node.js: v22.18.0

## Additional context

Confirmed that custom metadata exists in `storage.objects`.  
`.info()` retrieves it correctly, but `.list()` omits it.  
This looks like a storage server issue rather than a storage-js issue.

It would also be very useful to have a first-class way to update custom user metadata after upload (for example, renaming the filename field in metadata without re-uploading the file).

oniani1:

Bug fix - resolves #759

`storage.list()` and `list-v2` only return system metadata (`size`, `mimetype`, `eTag`) for each object. Custom `user_metadata` set during upload is omitted, forcing users to make N additional `.info()` calls to retrieve it.

## What is the new behavior?

All list/search endpoints now include `user_metadata` in their response. Files return their stored `user_metadata`; folders return `null`.

### Changes

**SQL migration (`0059-search-functions-user-metadata.sql`):**
- Drops and recreates four search functions with `user_metadata jsonb` added to their `RETURNS TABLE` and `SELECT` queries:
  - `storage.list_objects_with_delimiter` (used by v2 name-sorted listing)
  - `storage.search` (used by v1 listing, both name and non-name sorting paths)
  - `storage.search_v2` (router that delegates to the above)
  - `storage.search_by_timestamp` (used by v2 timestamp-sorted listing)
- `DROP FUNCTION IF EXISTS` is required before each `CREATE OR REPLACE` because PostgreSQL cannot change a function's return type in-place
- Folders emit `NULL` for `user_metadata` in all code paths

**TypeScript (`knex.ts`):**
- Added `'user_metadata'` to the select array in `listObjectsV2` for the non-delimiter direct query path (line 383)

**Test (`object.test.ts`):**
- Uploads a file with custom metadata, calls list, asserts `user_metadata` is present on files and `null` on folders

## Verification

- Ran all 59 migrations against a fresh PostgreSQL 15 instance
- Confirmed all 4 functions return `user_metadata` via direct SQL queries
- Tested all code paths: name sorting, timestamp sorting, non-name sorting, and direct delimiter listing

## Additional context

- The `Obj` schema and route handlers already supported `user_metadata` -- the data just never left the database layer
- `user_metadata` column has existed on `storage.objects` since migration 0025
- `.info()` correctly returns it; only the search/list functions were missing it
- Backward compatible: `normalizeColumns` already strips `user_metadata` for tenants below migration 25, and migrations run sequentially so the column always exists before this migration
- No S3 protocol changes needed -- S3 ListObjects spec doesn't include custom metadata

### DIFF
--- a/migrations/tenant/0059-search-functions-user-metadata.sql
+++ b/migrations/tenant/0059-search-functions-user-metadata.sql
@@ -2,6 +2,14 @@
 -- Each function's RETURNS TABLE gains a user_metadata jsonb column,
 -- and the query bodies include user_metadata in selects.
 -- Folders return NULL for user_metadata.
+--
+-- DROP is required because CREATE OR REPLACE cannot change return types.
+
+-- Drop old function signatures (return type is changing)
+DO $$ BEGIN DROP FUNCTION IF EXISTS storage.list_objects_with_delimiter(text,text,text,integer,text,text,text); EXCEPTION WHEN OTHERS THEN NULL; END; $$;
+DO $$ BEGIN DROP FUNCTION IF EXISTS storage.search(text,text,integer,integer,integer,text,text,text); EXCEPTION WHEN OTHERS THEN NULL; END; $$;
+DO $$ BEGIN DROP FUNCTION IF EXISTS storage.search_v2(text,text,integer,integer,text,text,text,text); EXCEPTION WHEN OTHERS THEN NULL; END; $$;
+DO $$ BEGIN DROP FUNCTION IF EXISTS storage.search_by_timestamp(text,text,integer,integer,text,text,text,text); EXCEPTION WHEN OTHERS THEN NULL; END; $$;
 
 -- ============================================================================
 -- list_objects_with_delimiter: Add user_metadata support
--- a/migrations/tenant/0059-search-functions-user-metadata.sql
+++ b/migrations/tenant/0059-search-functions-user-metadata.sql
@@ -1,0 +1,722 @@
+-- Add user_metadata to all search/list SQL functions
+-- Each function's RETURNS TABLE gains a user_metadata jsonb column,
+-- and the query bodies include user_metadata in selects.
+-- Folders return NULL for user_metadata.
+
+-- ============================================================================
+-- list_objects_with_delimiter: Add user_metadata support
+-- ============================================================================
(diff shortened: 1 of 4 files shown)


oniani1:
Dug into why CI stays red after the rebase. The tests themselves pass (635/635), but the Verify migration idempotency step fails because 0059 drops and recreates search, list_objects_with_delimiter, search_v2, and search_by_timestamp with user_metadata added to the return type. Once 0059 has run, the idempotency test can't re-apply migration 10 (search-files-search-function), since that migration tries to CREATE OR REPLACE the same 8-param signature with the old return type and PG rejects the return type change.

Two ways I can see to reshape this:

1. Add new function names (search_with_metadata, etc.) and leave the existing functions untouched, switching callers over
2. Drop the SQL-function path here and surface user_metadata only through the direct knex query branch, skipping the functions

Happy to go whichever direction fits the codebase conventions better, or if there's a third preferred pattern, let me know.

PR #959, opened 2026-04-01
refactor: get rid of pg queue app name env var

TylerHillery:

Refactor , to remove the redundant `PG_QUEUE_APPLICATION_NAME`.

Currently there are two variables set that determine the application name that is used when connecting to postgres
- `PG_QUEUE_APPLICAITON_NAME` which was used during the connection to the internal pg pool and MT DB
- `DB_APPLICATION_NAME` which was used to connect to the tenant DB

## What is the new behavior?

`PG_QUEUE_APPLICATION_NAME` is redundant for what we are trying to accomplish which is distinguishing the worker connections from the API connections, we only need the `DB_APPLICATION_NAME` to do this.


monitoring/grafana/dashboards/pgbouncer.json
```diff
@@ -0,0 +1,3806 @@
...
+        "wideLayout": true
+      },
+      "pluginVersion": "12.0.2",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "local_prometheus"
+          },
+          "expr": "pgbouncer_up{job=\"discovered-internal-storage-pgbouncer\", instance=~\"$instance\"}",
```
ferhatelmas: this won't [work here](https://github.com/supabase/storage/blob/master/monitoring/prometheus/prometheus.yml#L42) because job name isn't matching

ferhatelmas: it seems we're also running [different version of exporter](https://github.com/supabase/storage/blob/master/.docker/docker-compose-monitoring.yml#L3)

ferhatelmas:
I think env changes are fine. If you can want, we can merge it and add the setup/dashboard separately.

TylerHillery:
> I think env changes are fine. If you can want, we can merge it and add the setup/dashboard separately.

removed the dashboard changes from this PR

ferhatelmas (approved):

PR #961, opened 2026-04-02
fix: use object limit on orphan object scanner to avoid error

itslenny:

Bug fix

The orphan object scanner fails because it always hits the entity expansion limit in fast-xml-parser which was [introduced by this PR](https://github.com/supabase/storage/pull/927).

By default AWS list operation returns 1000 objects each includes an etag which is in the format `"abc123"` which counts as 2 entity expansions. This causes all lists that return more than 500 objects to fail.

Additionally, errors are not sent to the orphan scanner client so this error was not obvious and I had to find it in the server logs.

Error
```
Entity expansion limit exceeded: 1002 > 1000 Deserialization error: to see the raw response, inspect the hidden field {error}.$response on this object.
```

## What is the new behavior?

- Update `@aws-sdk` which removes the 1000 limit on the xml parser and includes the patched version of `fast-xml-parser` so the override can be removed
- Pass errors back to the orphan-client and display the error message. This ensures if an error occurs doing the orphan object scan we are made aware of it. Before this change it would just swallow the error and return partial results.
- Expand tests to include > 1000 objects with special characters, this reproduces the error without this update and covers the case of parsing multiple pages of objects

### DIFF
--- a/src/test/scanner.test.ts
+++ b/src/test/scanner.test.ts
@@ -17,13 +17,15 @@ describe('ObjectScanner', () => {
       name: bucketId,
     })
 
-    const maxUploads = 20
+    const maxUploads = 1100
 
-    // Create uploads
+    // Create uploads with special XML characters that contribute to entity expansion
     const result = await eachParallel(maxUploads, async (i) => {
+      // Add special XML characters to object names to increase entity count
+      const specialChars = i % 5 === 0 ? `&'<>"` : ''
       const upload = await storage.uploader.upload({
         bucketId: bucket.id,
-        objectName: randomUUID() + `-test-${i}.text`,
+        objectName: randomUUID() + `-test${specialChars}-${i}.text`,
         uploadType: 'standard',
         file: {
           body: Readable.from(Buffer.from('test')),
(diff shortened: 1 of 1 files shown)


src/scripts/orphan-client.ts
```diff
@@ -163,12 +168,17 @@ async function writeStreamToJsonArray(
...
+    stream.on('data', (data: OrphanObject | PingObject | ErrorObject) => {
       if (data.event === 'ping') {
         console.log('Received ping event, ignoring')
         return
       }
 
+      if (data.event === 'error') {
+        console.error('Server error:', data.error)
+        return
+      }
```
ferhatelmas, line 180: I think it's better to break the script to be explicit

src/http/routes/admin/objects.ts
```diff
@@ -174,6 +181,12 @@ export default async function routes(fastify: FastifyInstance) {
           )
         }
       } catch (e) {
+        reply.raw.write(
+          JSON.stringify({
+            event: 'error',
+            error: StorageBackendError.fromError(e),
+          }) + '\n'
+        )
```
ferhatelmas, line 189: It seems this won't expose the actual error to the script

ferhatelmas:
closing in favor of #963 + #965

PR #967, opened 2026-04-05
fix: read correct property name in PUT tenant handler

### LINKED ISSUE #966: PUT tenant handler reads wrong property name for maxResolution
In the PUT handler for tenants (`src/http/routes/admin/tenants.ts`), the condition correctly checks `features?.imageTransformation?.maxResolution` but the assignment reads `features?.imageTransformation?.image_transformation_max_resolution` -- which is the DB column name, not the request body property.

Since the request schema defines the field as `maxResolution` (camelCase), the snake_case property is always `undefined`, so the max resolution setting is silently dropped on PUT.

The PATCH handler in the same file handles this correctly, reading `maxResolution` consistently.

oniani1:
Fixes #966

The PUT handler for tenants checks `features?.imageTransformation?.maxResolution` in the condition but then reads `features?.imageTransformation?.image_transformation_max_resolution` in the assignment. The snake_case name is the DB column, not the request body property, so it's always `undefined`.

The PATCH handler right above does this correctly -- reads `maxResolution` in both places. This just aligns the PUT handler with the same pattern.

### DIFF
--- a/src/http/routes/admin/tenants.ts
+++ b/src/http/routes/admin/tenants.ts
@@ -471,7 +471,7 @@ export default async function routes(fastify: FastifyInstance) {
 
       if (typeof features?.imageTransformation?.maxResolution !== 'undefined') {
         tenantInfo.image_transformation_max_resolution = features?.imageTransformation
-          ?.image_transformation_max_resolution as number | undefined
+          ?.maxResolution as number | undefined
       }
 
       if (typeof features?.s3Protocol?.enabled !== 'undefined') {


src/internal/database/migrations/migrate.ts
```diff
@@ -806,7 +806,7 @@ async function doesSchemaExists(client: BasicPgClient, schemaName: string) {
       WHERE schema_name = ${schemaName}
   );`)
 
-  return result.rows.length > 0 && result.rows[0].exists === 'true'
+  return result.rows.length > 0 && result.rows[0].exists
```
fenos, line 809: Not sure why this change was committed?

fenos: Ah got it, is a change from another PR.

PR #968, opened 2026-04-05
fix: correct boolean check in doesSchemaExists

### LINKED ISSUE #928: `doesSchemaExist` will always return `false`
# Bug report

## Describe the bug

The function `doesSchemaExist` in `migrate.ts` always returns `false` no matter, whether the schema actually exists due to a type issue with the result of the SQL statement:

```ts
  return result.rows.length > 0 && result.rows[0].exists === 'true'
```

In consequence the storage service will always try to create the schema `storage` which fails, if the user does not have the necessary permissions.

## To Reproduce

1. Start from a freshly migrated database (Postgres 17)
2. apply least privileges to the `supabase_storage_admin` to allow creation of objects within the `storage` schema but not for the whole database
3. start the storage service with a proper database configuration
4. applying the migrations will fail due to "permission denied for database ..."

## Expected behavior

Because the schema already exists, the service should just apply the migrations instead of calling

```sql
CREATE SCHEMA IF NOT EXISTS storage
```

after running

```sql
GRANT CREATE ON DATABASE app TO supabase_storage_admin;
```

as `supabase_admin` the migrations are applied correctly.

## System information

- OS: Linux / Kubernetes
- Docker image: supabase/storage-api:v1.37.8

oniani1:
Fixes #928

`doesSchemaExists` compares the result of `SELECT EXISTS(...)` with the string `'true'`, but the `pg` driver returns a JavaScript boolean. So `true === 'true'` is always `false`, meaning the function always reports the schema doesn't exist.

This causes the migration to always run `CREATE SCHEMA IF NOT EXISTS storage`, which fails for self-hosted setups where the DB user has permissions within the storage schema but not `CREATE` on the database itself.

The fix aligns with how `doesTableExist` (same file, line 794) already handles the check -- just using the boolean value directly.

### DIFF
--- a/src/internal/database/migrations/migrate.ts
+++ b/src/internal/database/migrations/migrate.ts
@@ -806,7 +806,7 @@ async function doesSchemaExists(client: BasicPgClient, schemaName: string) {
       WHERE schema_name = ${schemaName}
   );`)
 
-  return result.rows.length > 0 && result.rows[0].exists === 'true'
+  return result.rows.length > 0 && result.rows[0].exists
 }
 
 /**


fenos:
Great catch @oniani1!

PR #975, opened 2026-04-06
fix: preserve charset in content-type for non-zero byte files

### LINKED ISSUE #816: Upload file content type charset is dropped
# Bug report

- [x] I confirm I have searched the Docs, GitHub Discussions, and Discord.

---

## Describe the bug

When uploading text files to Supabase Storage using the **Python SDK** with a
`Content-Type` that includes a `charset` parameter (for example
`text/markdown; charset=UTF-8`), the `charset` information is silently dropped
during the upload process.

Although the upload request explicitly includes the full `Content-Type`
(including `charset`), after the upload completes:

- The file metadata returned by the Storage `info` / `list` API
- The HTTP response headers returned by public URLs or signed URLs

**only contain the base MIME type** (e.g. `text/markdown`) and never include the
`charset` parameter.

This leads to incorrect or ambiguous character encoding handling in browsers and
HTTP clients that rely on response headers for charset detection.

---

### 1. Upload a file with charset specified

```python
from supabase import create_client

supabase = create_client(SUPABASE_URL, SUPABASE_KEY)

with open("example.md", "rb") as f:
    supabase.storage.from_("test-bucket").upload(
        "example.md",
        f,
        file_options={
            "content-type": "text/markdown; charset=UTF-8"
        }
    )
```

### 2. Verify upload request

- The actual HTTP request sent by the Python SDK is `multipart/form-data`
- The **file part** includes:

```
Content-Type: text/markdown; charset=UTF-8
```

---

### 3. Query file metadata

```python
files = supabase.storage.from_("test-bucket").list()
```

Returned metadata shows:

```
contentType: "text/markdown"
```

❌ `charset=UTF-8` is missing.

---

### 4. Access via signed URL or public URL

```python
url = supabase.storage.from_("test-bucket").create_signed_url(
    "example.md", 60
)["signedURL"]
```

```bash
curl -I "<signed-url>"
```

Response header:

```
Content-Type: text/markdown
```

❌ `charset=UTF-8` is missing.

---

## Expected behavior

If a file is uploaded with:

```
Content-Type: text/markdown; charset=UTF-8
```

Supabase Storage should either:

1. Preserve the full `Content-Type` (including `charset`) in stored metadata and
   HTTP responses
**or**
2. Clearly document that MIME parameters (such as `charset`) are intentionally
   stripped and unsupported

---

## Root cause analysis (from backend code inspection)

After inspecting the Supabase Storage backend source code, the root cause is in
`src/storage/backend/s3/adapter.ts`, specifically the `uploadObject` method:

### The Problem

1. **Content-Type is correctly extracted**: During multipart uploads, the backend
   correctly extracts the full `Content-Type` (including `charset`) from
   `formData.fields.contentType?.value` in `fileUploadFromRequest`
   (`src/storage/uploader.ts:344`).

2. **Content-Type is correctly sent to S3**: The full `Content-Type` (including
   `charset`) is passed to S3's `Upload` command as the `ContentType` parameter
   (`src/storage/backend/s3/adapter.ts:152`).

3. **Charset is lost when retrieving metadata**: After upload completes, for
   **non-zero-byte files**, the code calls `headObject` to retrieve metadata
   (`src/storage/backend/s3/adapter.ts:175`). S3's `HeadObjectCommand` returns a
   `ContentType` that has been normalized/stripped of the `charset` parameter.

4. **Normalized value overwrites original**: The code uses the `mimetype` from
   `headObject` response (`src/storage/backend/s3/adapter.ts:190`) instead of the
   original `contentType` parameter, causing the `charset` to be lost.

### Code Evidence

```typescript
// src/storage/backend/s3/adapter.ts:174-190
const metadata = hasUploadedBytes
  ? await this.headObject(bucketName, key, version)
  : {
      mimetype: contentType,
      // ...
    }

return {
  // ...
  mimetype: metadata.mimetype,
  // ...
}
```

### Why Zero-Byte Files Work

For zero-byte files, the code bypasses `headObject` and directly uses the
original `contentType` parameter (line 179), which is why `charset` is
preserved for empty files but not for files with content.

### The Fix

The code should use the original `contentType` parameter for non-zero-byte
files instead of relying on `headObject`'s normalized response, or merge the
original `contentType` with other metadata from `headObject`.

---

## Why this is a bug (not a feature request)

- `charset` is a standard, semantically significant part of `Content-Type`
- Browsers and HTTP clients rely on it for correct text decoding
- The Python SDK explicitly sends the full `Content-Type`
- Supabase Storage silently drops it without documentation or warning

---

## System information

- OS: macOS / Linux
- SDK: supabase==2.27.0, storage3==2.27.0
- Python version: 3.12
- Storage: Supabase managed Storage

SushanthMusham:
Bug fix
When uploading a non-zero byte file with a Content-Type that includes a charset parameter (e.g. text/markdown; charset=UTF-8), the charset is silently dropped. This happens because bufferedMultipartUpload calls headObject after upload, and S3 normalizes the Content-Type by stripping the charset before returning it.
Fixes #816
## What is the new behavior?
The original contentType parameter (which contains the full Content-Type including charset) is now used directly in the return value, instead of relying on the normalized response from S3's headObject. Falls back to metadata.mimetype if contentType is empty.

// Before
mimetype: metadata.mimetype

// After
mimetype: contentType || metadata.mimetype

## Additional context
Zero-byte files were already handled correctly because they bypass headObject entirely. This fix brings non-zero byte files in line with the same behaviour.

### DIFF
--- a/src/storage/backend/s3/adapter.ts
+++ b/src/storage/backend/s3/adapter.ts
@@ -283,7 +283,7 @@ export class S3Backend implements StorageBackendAdapter {
         httpStatusCode: data.$metadata.httpStatusCode || metadata.httpStatusCode,
         cacheControl,
         eTag: metadata.eTag,
-        mimetype: metadata.mimetype,
+        mimetype: contentType || metadata.mimetype,
         contentLength: metadata.contentLength,
         lastModified: metadata.lastModified,
         size: metadata.size,
--- a/src/test/object.test.ts
+++ b/src/test/object.test.ts
@@ -286,6 +286,35 @@ describe('testing GET object', () => {
  * multipart upload
  */
 describe('testing POST object via multipart upload', () => {
+  test('preserve charset in content-type for non-zero byte files', async () => {
+    const form = new FormData()
+    form.append('file', fs.createReadStream(`./src/test/assets/sadcat.jpg`)) 
+    
+    const headers = Object.assign({}, form.getHeaders(), {
+      authorization: `Bearer ${process.env.AUTHENTICATED_KEY}`,
+      'x-upsert': 'true',
+      'content-type': 'text/markdown; charset=UTF-8', 
+    })
+
+    const response = await appInstance.inject({
+      method: 'POST',
+      url: '/object/bucket2/charset-test.md',
+      headers,
+      payload: form,
+    })
+  
+    expect(response.statusCode).toBe(200)
+    
+    const client = await getSuperuserPostgrestClient()
+    const object = await client
+      .table('objects')
+      .select('*')
+      .where('name', 'charset-test.md')
+      .first()
+
+    expect(object).not.toBeFalsy()
+    expect(object?.metadata?.mimetype).toBe('text/markdown; charset=UTF-8')
+  })
   test('check if RLS policies are respected: authenticated user is able to upload authenticated resource', async () => {
     const form = new FormData()
     form.append('file', fs.createReadStream(`./src/test/assets/sadcat.jpg`))
@@ -807,7 +836,9 @@ describe('testing POST object via multipart upload', () => {
       .first()
 
     expect(objectResponse).toBe(undefined)
+    
   })
+  
 })
 
 /*


SushanthMusham:
Hey @ferhatelmas, could you take a quick look at this? Thanks!

ferhatelmas:
@SushanthMusham can we add some tests? also, we shouldn't have any diff in package lock

ferhatelmas:
I had made a change in https://github.com/supabase/storage/pull/981 so that CI can run for contributors. Please rebase your change to get this

SushanthMusham:
Thanks for the review, @ferhatelmas!

I will rebase my branch now and revert the unintended changes to the package-lock.json file. I'll also try adding the test coverage for the charset preservation and push those updates shortly.

SushanthMusham:
Hey @ferhatelmas, so sorry for the delay—I had my 6th sem exams.
I’ve rebased the branch with your latest CI fixes and updated my npm version (11.12.1) to avoid package-lock issues.
I added an integration test in src/test/object.test.ts that:
1) Uploads a non-empty file (sadcat.jpg) to bucket2
2) Sets a Content-Type with charset (text/markdown; charset=UTF-8)
3) Checks in Postgres that the charset is preserved and not removed
I used bucket2 because it doesn't have the strict mimetype RLS rules that some other test buckets have. This is my first time writing a test, please let me know if there's anything I should adjust!

SushanthMusham:
@ferhatelmas Hi , any issues with the code ?

src/storage/backend/s3/adapter.ts
```diff
@@ -283,7 +283,7 @@ export class S3Backend implements StorageBackendAdapter {
         httpStatusCode: data.$metadata.httpStatusCode || metadata.httpStatusCode,
         cacheControl,
         eTag: metadata.eTag,
-        mimetype: metadata.mimetype,
+        mimetype: contentType || metadata.mimetype,
```
ferhatelmas, line 286: This isn't the right solution because in multipart  case, our dependency `@fastify/busboy` strips key value pairs and returns only mimetype.

https://github.com/fastify/busboy/issues/127 
I think we need to follow up this issue first.

src/test/object.test.ts
```diff
@@ -286,6 +286,35 @@ describe('testing GET object', () => {
...
 describe('testing POST object via multipart upload', () => {
+  test('preserve charset in content-type for non-zero byte files', async () => {
+    const form = new FormData()
+    form.append('file', fs.createReadStream(`./src/test/assets/sadcat.jpg`)) 
+    
+    const headers = Object.assign({}, form.getHeaders(), {
+      authorization: `Bearer ${process.env.AUTHENTICATED_KEY}`,
+      'x-upsert': 'true',
+      'content-type': 'text/markdown; charset=UTF-8', 
+    })
```
ferhatelmas, line 297: This is also wrong because we need to ensure multipart upload case

ferhatelmas (changes requested):

ferhatelmas:
closing but feel free to open a new one with the suggested approach

PR #994, opened 2026-04-08
fix: empty segment listing

ferhatelmas:

Bug fix

Repeated `/` segments in object keys make list misidentify children by producing blank folders or blank names.

## What is the new behavior?

Skip empty path segments and derive the first real child name instead.

## Additional context

Validate prefix/key alignment before deriving child paths.
Make non-name sorting deterministic with explicit folder-first ordering.


ferhatelmas:
closing in favor of #1346

PR #998, opened 2026-04-09
feat: return object id in TUS upload completion response header

### LINKED ISSUE #647: Missing object_id in resumable uploads response using TUS
# Bug report

## Describe the bug

When uploading using the tus-js-client just as described in the docs: [here]()

## To Reproduce

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

oniani1:
Closes #647

## Summary

After a TUS resumable upload completes, there's currently no way to get the `object_id` of the uploaded file without making a follow-up `.list()` call. The `onSuccess` callback from `tus-js-client` only gets the upload URL, not the object ID.

The standard `POST /object` endpoint already returns the ID in the JSON body (`{ Id, Key }`). For TUS, since the response format is governed by the protocol, a response header is the natural fit -- same approach @fenos suggested in #647.

This captures the return value of `completeUpload()` in the `onUploadFinish` lifecycle hook and includes the object's UUID in a new `x-supabase-id` response header. The `@tus/server` library already supports custom headers -- both `PatchHandler` and `PostHandler` merge them into the outgoing response.

## Changes

- `src/http/routes/tus/lifecycle.ts` -- capture `completeUpload()` result, add `x-supabase-id` header
- `src/test/tus.test.ts` -- new test verifying the header is present, is a valid UUID, and matches the database record

## Test plan

- [x] Added test: TUS upload completion returns `x-supabase-id` header with valid UUID matching DB record
- [ ] Existing TUS tests pass with no regressions

### DIFF
--- a/src/http/routes/tus/lifecycle.ts
+++ b/src/http/routes/tus/lifecycle.ts
@@ -308,7 +308,7 @@ export async function onUploadFinish(rawReq: Request, upload: Upload) {
       }
     }
 
-    await uploader.completeUpload({
+    const result = await uploader.completeUpload({
       version: resourceId.version,
       bucketId: resourceId.bucket,
       objectName: resourceId.objectName,
@@ -322,6 +322,7 @@ export async function onUploadFinish(rawReq: Request, upload: Upload) {
     return {
       headers: {
         'Tus-Complete': '1',
+        'x-supabase-id': result.obj.id,
       },
     }
   } catch (e) {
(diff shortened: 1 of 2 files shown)


oniani1:
Bumping this. CI is green, happy to rework if returning the object id via header feels wrong versus the response body.

PR #999, opened 2026-04-09
feat: create admin endpoint to roll tenant url signing jwk

itslenny:

Feature

There is no admin api endpoint to roll the tenant url signing jwk. This means that rolling the url signing jwk for a tenant is possible but a clunky manual process.

## What is the new behavior?

Add admin api endpoint that rolls the url signing jwk

### DIFF
--- a/src/http/routes/admin/jwks.ts
+++ b/src/http/routes/admin/jwks.ts
@@ -1,6 +1,7 @@
 import { UrlSigningJwkGenerator } from '@internal/auth/jwks/generator'
 import { jwksManager } from '@internal/database'
 import { logSchema } from '@internal/monitoring'
+import { JwksRollUrlSigningKey } from '@storage/events'
 import { FastifyInstance, RequestGenericInterface } from 'fastify'
 import { FromSchema } from 'json-schema-to-ts'
 import apiKey from '../../plugins/apikey'
@@ -48,6 +49,12 @@ interface JwksUpdateRequestInterface extends RequestGenericInterface {
   Params: {
     tenantId: string
     kid: string
+  }
+}
+
+interface JwksRollRequestInterface extends RequestGenericInterface {
+  Params: {
+    tenantId: string
   }
 }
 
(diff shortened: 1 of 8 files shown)


src/http/routes/admin/jwks.ts
```diff
@@ -127,6 +134,23 @@ export default async function routes(fastify: FastifyInstance) {
...
+      const { tenantId } = request.params
+
+      await JwksRollUrlSigningKey.send({
+        tenantId,
+        tenant: {
+          ref: tenantId,
+        },
+      })
+
+      return reply.send({ started: true })
```
ferhatelmas, line 150: This can be a bit misleading because send can decide not to send due to disabled events. Shall we override `shouldSend` for this event?

src/internal/auth/jwks/manager.ts
```diff
@@ -66,6 +66,30 @@ export class JWKSManager {
     return { kid: createJwkKid({ kind: JWK_KIND_STORAGE_URL_SIGNING, id }) }
   }
 
+  /**
+   * Atomically rolls the URL signing JWK by deactivating the current key and creating a new one
+   * @param tenantId
+   */
+  async rollUrlSigningJwk(tenantId: string): Promise<{ oldKid: string | null; newKid: string }> {
+    return this.storage.transaction(async (trx) => {
+      const currentKeys = await this.storage.listActive(tenantId, JWK_KIND_STORAGE_URL_SIGNING)
```
ferhatelmas: Why not using `trx` for this case? 

Due to implementation, it doesn't matter at the moment but it could bite later.

src/storage/events/jwks/jwks-roll-url-signing-key.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  static async handle(job: Job<JwksRollUrlSigningKeyPayload>) {
+    const { tenantId } = job.data
+
+    try {
+      const { oldKid, newKid } = await jwksManager.rollUrlSigningJwk(tenantId)
+
+      logSchema.info(
+        logger,
+        `[Jwks] rolled url signing key for tenant ${tenantId} (old: ${oldKid}, new: ${newKid})`,
```
ferhatelmas, line 49: Just pointing out duplication since we add the project in the context, we might drop `tenantId` from the log line. Sometimes search in the message itself is useful though 🤷🏻‍♂️

ferhatelmas (approved):

PR #1016, opened 2026-04-13
fix: add lifecycle logs for empty bucket operations

itslenny:

Bug fix

There are no lifecycle logs emitted when emptying buckets

## What is the new behavior?

Emit delete object lifecycle logs for all objects removed when a bucket is emptied

### DIFF
--- a/src/storage/events/objects/object-admin-delete-all-before.ts
+++ b/src/storage/events/objects/object-admin-delete-all-before.ts
@@ -5,6 +5,7 @@ import { Job, SendOptions, WorkOptions } from 'pg-boss'
 import { getConfig } from '../../../config'
 import { Storage } from '../../index'
 import { BaseEvent } from '../base-event'
+import { ObjectRemoved } from '../lifecycle/object-removed'
 
 const DELETE_JOB_TIME_LIMIT_MS = 10_000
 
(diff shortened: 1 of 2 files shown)


TylerHillery (approved):

src/storage/events/objects/object-admin-delete-all-before.ts
```diff
@@ -87,6 +88,19 @@ export class ObjectAdminDeleteAllBefore extends BaseEvent<ObjectDeleteAllBeforeE
               }
 
               await backend.deleteObjects(storageS3Bucket, prefixes)
+
+              await Promise.allSettled(
```
ferhatelmas, line 92: can we add a test to ensure event is triggerred to prevent regressions?

src/test/webhooks.test.ts
```diff
@@ -385,10 +386,95 @@ describe('Webhooks', () => {
       })
     )
   })
+
+  it('will emit webhooks for each deleted object during empty bucket operation', async () => {
+    const emptyTestBucketName = 'bucket-empty-webhook-test'
```
ferhatelmas, line 391: might be better to make it unique as well

ferhatelmas (approved):

PR #1017, opened 2026-04-13
feat: add tenant config per request hit ratio

ferhatelmas:

feat

Tenant config cache is used multiple times in a request so hit ratio is overcounting.

## What is the new behavior?

Add a separate metric to track hits per request.

## Additional context

Update dash to add it.


ferhatelmas:
After thinking further and seeing the numbers, its benefit is low compared to overhead. It gave an idea of implementation but won't merge.

PR #1032, opened 2026-04-16
fix: validate before date in DELETE orphan-objects

### LINKED ISSUE #1031: DELETE /orphan-objects does not validate the `before` field
The DELETE handler in `src/http/routes/admin/objects.ts` accepts an arbitrary string in the `before` body field, passes it to `new Date(...)`, and proceeds even if the result is `Invalid Date`.

The GET handler on the same route validates this:

```ts
if (before && isNaN(before.getTime())) {
  return reply.status(400).send({
    error: 'Invalid date format',
  })
}
```

The DELETE handler is missing the same check.

### Reproduction
```
DELETE /tenants/:tenantId/buckets/:bucketId/orphan-objects
Content-Type: application/json

{"deleteDbKeys": true, "before": "not-a-date"}
```

Expected: `400 Invalid date format`.
Actual: returns 200 -- the Invalid Date is truthy so the default-hour branch doesn't run, and the Invalid Date propagates into `scanner.deleteOrphans`. Date comparisons against NaN all evaluate false, so the scan quietly matches nothing.

oniani1:
Closes #1031

The DELETE handler accepted any string as the `before` body field and passed it straight to `new Date()`. An invalid string becomes `Invalid Date`, which is truthy, so the `if (!before)` default-hour branch was skipped and the `Invalid Date` propagated into `scanner.deleteOrphans`. Comparisons against NaN are all false, so the scan quietly matched nothing.

The GET handler on the same route already validates this. Mirror the same `isNaN` check in DELETE, plus integration tests covering both handlers and the existing `deleteDbKeys`/`deleteS3Keys` guard.

### DIFF
--- a/src/test/admin-orphan-objects.test.ts
+++ b/src/test/admin-orphan-objects.test.ts
@@ -47,7 +47,7 @@ describe('admin orphan-objects routes', () => {
       expect(JSON.parse(response.body).error).toBe('Invalid date format')
     })
 
-    it('returns 400 when neither deleteDbKeys nor deleteS3Keys is set (unchanged behavior)', async () => {
+    it('returns 400 when neither deleteDbKeys nor deleteS3Keys is set', async () => {
       const response = await adminApp.inject({
         method: 'DELETE',
         url: '/tenants/bjhaohmqunupljrqypxz/buckets/bucket2/orphan-objects',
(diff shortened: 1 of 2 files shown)


src/test/admin-orphan-objects.test.ts
```diff
@@ -0,0 +1,70 @@
...
+          deleteS3Keys: false,
+          before: 'not-a-date',
+        }),
+      })
+
+      expect(response.statusCode).toBe(400)
+      expect(JSON.parse(response.body).error).toBe('Invalid date format')
+    })
+
+    it('returns 400 when neither deleteDbKeys nor deleteS3Keys is set (unchanged behavior)', async () => {
```
ferhatelmas:
```suggestion
    it('returns 400 when neither deleteDbKeys nor deleteS3Keys is set', async () => {
```

ferhatelmas (approved):

PR #1039, opened 2026-04-21
fix: path traversal for tus

ferhatelmas:

Bug fix

File backend prevents path traversal via #818 and #883 but TUS with file backend has still the same problem.

## What is the new behavior?

Extract secure path helper from file backend and protect upload ids for TUS uploads.

## Additional context

Add file backend to integration tests.
Add concurrency control into delete expired and guard for zero offset where it would delete complete uploads if expiration is given and gc was called.
Remove exposes more info about error now.


src/storage/protocols/tus/file-store.ts
```diff
@@ -1,10 +1,18 @@
...
+import type http from 'node:http'
+import stream from 'node:stream'
+import { ErrorCode, StorageBackendError } from '@internal/errors'
 import { ensureFile } from '@internal/fs'
 import { Configstore, FileStore as TusFileStore } from '@tus/file-store'
 import { Upload } from '@tus/server'
-import path from 'path'
+import { ERRORS as TUS_ERRORS } from '@tus/utils'
 import { FileBackend } from '../../backend'
+import { resolveSecureFilesystemPath } from '../../backend/secure-path'
```
itslenny: You can use `@storage/backend` (and for FileBackend above)

ferhatelmas: updated

itslenny (approved):

src/test/tus.test.ts
```diff
@@ -403,16 +828,19 @@ describe('Tus multipart', () => {
...
 
+      const signedAt = new Date()
+      vi.setSystemTime(signedAt)
+
       const signedUpload = await storage
         .from(bucketName)
         .signUploadObjectUrl(objectName, `${bucketName}/${objectName}`, 1)
 
-      await wait(2000)
+      vi.setSystemTime(new Date(signedAt.getTime() + 2000))
```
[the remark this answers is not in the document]
ferhatelmas: this is false positive, `setSystemTime` mocks `Date.*` and constructor without timers

PR #1041, opened 2026-04-22
feat: add sb-request-id logging

ferhatelmas:

feat

Supabase specific request id isn't supported.

## What is the new behavior?

Supabase specific request id `sb-request-id` header is read and propagated downstream and serialized as `request_id` at top level for logflare.

## Additional context

Increased coverage a bit when applicable.
Refactored a few places not using log schema. 
Also propagate the existing `reqId` where missed.
Fixed any usage in touched files except db to keep scope smaller, related to #922


src/internal/queue/constants.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Placeholder tenant for events that run across the whole fleet
+ * (e.g. pg-boss maintenance) and have no single tenant context.
+ */
+export const SYSTEM_TENANT = { ref: '', host: '' } as const
```
itslenny, line 6: would it make sense to make this something other than empty strings? Like 'SYSTEM_TENANT' (so if we see it in logs it's more obvious)

itslenny (approved):

src/internal/queue/constants.ts
ferhatelmas: Sounds fine. I will keep the constant so no logical change and serialized version is small but will log with `SYSTEM_TENANT`

ferhatelmas: updated, see the new commit

PR #1043, opened 2026-04-23
feat: add queue management handlers

ferhatelmas:

feat

Manual queue management under load while jumping between systems.

## What is the new behavior?

Admin app attaches endpoints for listing, backup and restore.
Add a client to attach it via npm scripts.

## Additional context

Kind of playbook automation/ease of use.
Disabled orioledb for restore due to targetless on conflict incompatibility, that support isn't really needed.


src/internal/strings/boolean.test.ts
```diff
@@ -0,0 +1,19 @@
+import { parseOptionalBoolean } from './boolean'
```
fenos: I would rename this package to `strings` as opposed to `parser`
it seems that this is doing strings parsing. 

Inspired by go: strings package 😄

ferhatelmas: sounds good, renamed

src/internal/database/pg-connection.ts
```diff
@@ -835,6 +838,62 @@ export function createAbortError(): Error & { code: string } {
   return error
 }
 
+function acquirePoolClientWithSignal(pool: Pool, signal: AbortSignal): Promise<PoolClient> {
```
fenos, line 922: I'm worried that this function might caused some race conditions.
For example what if we abort the signal while a query is still ongoing and we call `release`?

It might create some issue. I'm not super confident that having an abort signal on acquisition when the signal can't be passed down to `connect` (which i'm not sure is cancellable)

ferhatelmas: Yes, it isn't cancelable. I wanted to add here exactly because queue connection timeout is unbounded (for this case - incident, better to fail fast). Ideally, we should set it to a reasonable value but I didn't want to change it since it's operationally bigger, instead capped here so that we can retry.

For the case (running query and release), ownership is transferred and as usual, connection will be destroyed. 

Also, added two new tests in the last commit for more confidence

ferhatelmas:
@claude review always

ferhatelmas:
closing in favor of upcoming closer queue write because this will become obsolete hopefully

PR #1048, opened 2026-04-23
fix(s3): set NextMarker in ListObjects V1 response when truncated

### LINKED ISSUE #1047: S3 ListObjects V1 response never sets NextMarker on truncated results
## Bug

`S3ProtocolHandler.listObjects` in `src/storage/protocols/s3/s3-handler.ts` (handling `ListObjectsCommand`, the V1 API) never emits `NextMarker`, even when `IsTruncated` is true. Per the [S3 V1 spec](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html#AmazonS3-ListObjects-response-NextMarker), `NextMarker` is the cursor a client is expected to use for the next page when the response is truncated.

The handler does populate `Marker` with the V2 `NextContinuationToken` (in `cursorV1: true` mode this is the plain last-key), so AWS SDK paginators happen to still work because they fall back to `resp.Marker` / the last `Contents` key. But any client that specifically reads `resp.NextMarker` (explicit manual pagination, strict spec clients, some third-party tooling) receives `undefined` and can't advance.

## Reproduction

Against a local storage server, upload N objects and issue a V1 `ListObjectsCommand` with `MaxKeys < N`:

```js
const resp = await s3.send(new ListObjectsCommand({ Bucket, MaxKeys: 2 }))
// resp.IsTruncated === true
// resp.NextMarker === undefined  <-- spec violation
// resp.Marker === '<last key>'   <-- next-page hint leaks into wrong field
```

## Expected

When `IsTruncated === true`, `NextMarker` should carry the cursor used for the next request (for flat listings, the last key in `Contents`; for delimiter listings, the last entry across `Contents` / `CommonPrefixes`). When `IsTruncated === false`, `NextMarker` should be absent.

## Related / out of scope

The response `Marker` field is also spec-wrong today (it should echo the request's input `Marker`, not carry the next-page hint). Fixing that is a backwards-incompatible change for any client that has grown to depend on the current value, so it should be done separately after a compat decision.

oniani1:
Closes #1047.

## Summary

S3 `ListObjects` V1 has to return `NextMarker` alongside `IsTruncated` so that pagination-aware clients can advance without reverse-engineering the cursor off `Contents`. `S3ProtocolHandler.listObjects` never emitted it. The next-page hint was leaking into the `Marker` field instead, and any client that specifically read `NextMarker` got `undefined`.

Sets `NextMarker` to the V2 `NextContinuationToken`, which under `cursorV1: true` is the plain last key (`results.nextCursorKey`). Only emitted when `IsTruncated === true` and the token is present, via a conditional spread, so the XML serializer does not render an empty `<NextMarker/>` element in the non-truncated case.

`Marker` is intentionally left unchanged. It is still populated with the next-page hint rather than the echoed input `Marker` per the spec. The AWS SDK paginator prefers `NextMarker` now that it is available, and changing `Marker`'s value is a backwards-incompatible change for any client that has grown to depend on it. Tracked separately in the issue.

## Test plan

Three tests added / adjusted in `src/test/s3-protocol.test.ts` under `ListObjectCommand`:

- [x] `sets NextMarker when the response is truncated`. Flat listing, confirms `NextMarker` is defined and usable as next request's `Marker`.
- [x] `omits NextMarker when the response is not truncated`. Confirms the field is absent (not empty string).
- [x] `sets NextMarker when truncated with a delimiter`. Delimiter listing, confirms `NextMarker` carries the cursor and feeding it back returns the remaining common prefixes.

All `ListObjectCommand` tests pass against a local Docker storage stack; existing `ListObjectsV2Command` tests are unaffected.

### DIFF
--- a/src/storage/protocols/s3/s3-handler.ts
+++ b/src/storage/protocols/s3/s3-handler.ts
@@ -178,7 +178,10 @@ export class S3ProtocolHandler {
           Name: v2Result.Name,
           Prefix: v2Result.Prefix,
           Marker: v2Result.NextContinuationToken,
-          ...(v2Result.IsTruncated && v2Result.NextContinuationToken
+          // NextMarker is returned only when the response is truncated AND the
+          // Delimiter request parameter is specified, per the S3 ListObjects V1
+          // reference.
+          ...(v2Result.IsTruncated && v2Result.NextContinuationToken && command.Delimiter
             ? { NextMarker: v2Result.NextContinuationToken }
             : {}),
           MaxKeys: v2Result.MaxKeys,
--- a/src/storage/protocols/s3/s3-handler.ts
+++ b/src/storage/protocols/s3/s3-handler.ts
@@ -170,17 +170,22 @@ export class S3ProtocolHandler {
       cursorV1: true,
     })
 
+    const v2Result = list.responseBody.ListBucketResult
+
     return {
       responseBody: {
         ListBucketResult: {
-          Name: list.responseBody.ListBucketResult.Name,
-          Prefix: list.responseBody.ListBucketResult.Prefix,
-          Marker: list.responseBody.ListBucketResult.NextContinuationToken,
-          MaxKeys: list.responseBody.ListBucketResult.MaxKeys,
-          IsTruncated: list.responseBody.ListBucketResult.IsTruncated,
-          Contents: list.responseBody.ListBucketResult.Contents,
-          CommonPrefixes: list.responseBody.ListBucketResult.CommonPrefixes,
-          EncodingType: list.responseBody.ListBucketResult.EncodingType,
+          Name: v2Result.Name,
+          Prefix: v2Result.Prefix,
+          Marker: v2Result.NextContinuationToken,
+          ...(v2Result.IsTruncated && v2Result.NextContinuationToken
+            ? { NextMarker: v2Result.NextContinuationToken }
+            : {}),
+          MaxKeys: v2Result.MaxKeys,
+          IsTruncated: v2Result.IsTruncated,
+          Contents: v2Result.Contents,
+          CommonPrefixes: v2Result.CommonPrefixes,
+          EncodingType: v2Result.EncodingType,
         },
       },
     }
--- a/src/test/s3-protocol.test.ts
+++ b/src/test/s3-protocol.test.ts
@@ -571,6 +571,65 @@ describe('S3 Protocol', () => {
         expect(resp2.CommonPrefixes?.length).toBe(2)
         expect(resp2.Contents?.length).toBe(1)
       })
+
+      it('sets NextMarker when the response is truncated', async () => {
+        const bucket = await createBucket(client)
+
+        await Promise.all([
+          uploadFile(client, bucket, 'a.jpg', 1),
+          uploadFile(client, bucket, 'b.jpg', 1),
+          uploadFile(client, bucket, 'c.jpg', 1),
+        ])
+
+        const truncated = await client.send(
+          new ListObjectsCommand({ Bucket: bucket, MaxKeys: 2 })
+        )
+        expect(truncated.IsTruncated).toBe(true)
+        expect(truncated.NextMarker).toBeDefined()
+
+        // NextMarker must be usable as the next request's Marker per S3 spec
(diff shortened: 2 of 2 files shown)


src/storage/protocols/s3/s3-handler.ts
```diff
@@ -170,17 +170,22 @@ export class S3ProtocolHandler {
...
-          Marker: list.responseBody.ListBucketResult.NextContinuationToken,
-          MaxKeys: list.responseBody.ListBucketResult.MaxKeys,
-          IsTruncated: list.responseBody.ListBucketResult.IsTruncated,
-          Contents: list.responseBody.ListBucketResult.Contents,
-          CommonPrefixes: list.responseBody.ListBucketResult.CommonPrefixes,
-          EncodingType: list.responseBody.ListBucketResult.EncodingType,
+          Name: v2Result.Name,
+          Prefix: v2Result.Prefix,
+          Marker: v2Result.NextContinuationToken,
+          ...(v2Result.IsTruncated && v2Result.NextContinuationToken
```
ferhatelmas: > This element is returned only if you have the delimiter request parameter specified

I think we should handle this condition as well and adjust tests accordingly

src/storage/protocols/s3/s3-handler.ts
```diff
@@ -170,17 +170,22 @@ export class S3ProtocolHandler {
...
-          Prefix: list.responseBody.ListBucketResult.Prefix,
-          Marker: list.responseBody.ListBucketResult.NextContinuationToken,
-          MaxKeys: list.responseBody.ListBucketResult.MaxKeys,
-          IsTruncated: list.responseBody.ListBucketResult.IsTruncated,
-          Contents: list.responseBody.ListBucketResult.Contents,
-          CommonPrefixes: list.responseBody.ListBucketResult.CommonPrefixes,
-          EncodingType: list.responseBody.ListBucketResult.EncodingType,
+          Name: v2Result.Name,
+          Prefix: v2Result.Prefix,
+          Marker: v2Result.NextContinuationToken,
```
ferhatelmas, line 180:
```suggestion
          ...(command.Marker !== undefined ? { Marker: command.Marker } : {}),
```

@supabase/storage I think we should update this separately for compliance. It could be [breaking theoretically for some](https://www.hyrumslaw.com/) but I guess minimal risk

oniani1:
Thanks @ferhatelmas! Addressed the NextMarker + delimiter condition in ada90f1. Tightened the condition to require `command.Delimiter` and flipped the no-delimiter test to assert `NextMarker` is omitted, matching the spec note that NextMarker is only returned when Delimiter is set.

On the `Marker` line, good catch on the compliance gap. I held off bundling that one here because of the behavior change you flagged. The pre-existing `list all keys with pagination` test paginates via `resp.Marker`, which suggests external callers likely do the same. Swapping to `command.Marker` makes `resp.Marker` undefined on the first page and breaks those pagination loops.

Happy to either (a) add it here with a test migration to `resp.NextMarker` and a note in the PR description, or (b) split it into a follow-up PR so this one can merge cleanly. Which do you prefer?

src/storage/protocols/s3/s3-handler.ts
oniani1: Done in ada90f1. Gated the NextMarker spread on `command.Delimiter` and flipped the no-delimiter test to assert `NextMarker` is undefined.

src/storage/protocols/s3/s3-handler.ts
oniani1: Held off on this one in the current push. Details in the top-level reply above: the pre-existing `list all keys with pagination` test paginates via `resp.Marker`, and swapping to `command.Marker` makes that undefined on the first page and breaks external callers doing the same. Happy to fold it in here with a test migration, or split it into a follow-up PR so this can merge cleanly. Let me know which you prefer.

ferhatelmas: I think this change is good as it is to fix `NextMarker`. 

Let's prepare `Marker` issue separately to make it obvious so that we can refer people to the exact minimal change if they are affected and mark it breaking.

ferhatelmas (approved):

PR #1051, opened 2026-04-24
fix: drop axios from cdn cache manager

ferhatelmas:

reafactor

Axios is used in cdn cache manager.

## What is the new behavior?

Native fetch replaces axios.

## Additional context

Added some unit test coverage.

### DIFF
--- a/src/storage/cdn/cdn-cache-manager.ts
+++ b/src/storage/cdn/cdn-cache-manager.ts
@@ -1,27 +1,25 @@
 import { ERRORS } from '@internal/errors'
 import { Storage } from '@storage/storage'
-import { HttpsAgent } from 'agentkeepalive'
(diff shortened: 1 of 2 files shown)


src/storage/cdn/cdn-cache-manager.ts
```diff
@@ -42,19 +53,31 @@ export class CdnCacheManager {
...
+            ref: opts.tenant,
+          },
+          bucketId: opts.bucket,
+          objectName: opts.objectName,
+        }),
+        dispatcher,
       }
 
-      throw e
+      const response = await fetch(resolvePurgeUrl(cdnPurgeEndpointURL), requestInit)
```
ferhatelmas: let's cache it

itslenny (approved):

PR #1059, opened 2026-04-27
fix: add auth schemes for open api spec

itslenny:

fix to open api specs

Open API spec does not include auth schemes. This makes the swagger UI unusable as you cannot authenticate

## What is the new behavior?

Add auth schemes to all authenticated endpoints (api and admin api). This makes the open api json more accurate and also enables authentication in the swagger UI when developing which allows executing endpoints directly in the swagger UI (http://localhost:5000/documentation)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "test:integration": "vitest run --config vitest.integration.config.ts",
     "test:integration:watch": "vitest --config vitest.integration.config.ts",
     "test:integration:coverage": "vitest run --coverage --config vitest.integration.config.ts",
-    "test": "npm run infra:restart && npm run test:dummy-data && npm run test:integration",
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "test:integration": "vitest run --config vitest.integration.config.ts",
     "test:integration:watch": "vitest --config vitest.integration.config.ts",
     "test:integration:coverage": "vitest run --coverage --config vitest.integration.config.ts",
-    "test": "npm run infra:restart && npm run test:dummy-data && npm run test:integration",
+    "test": "npm run infra:restart && npm run test:dummy-data && npm run test:integration && npm run test:unit",
```
ferhatelmas: maybe change order to get faster feedback

ferhatelmas (approved):

PR #1061, opened 2026-04-27
feat: add error code, normalize more errors, and exclude stack traces from 4xx errors

itslenny:

Feature

- logs all stack traces. 404, 400, etc are very noisy and not useful and loading all stack traces adds up to real performance impact at scale
- error code can only be accessed via error.raw
- several errors were missing data (error code and/or status code)

## What is the new behavior?

Only log stack traces for http errors >= 500 and unknown errors

Add `errorCode` to logged errors to expose the [error codes defined in the docs](https://supabase.com/docs/guides/storage/debugging/error-codes?queryGroups=language&language=js#storage-error-codes). Prior to this change these can only be accessed by digging in to error.raw

### Normalized Errors

- Missing error code. These errors had no error code
  - TUS metadata parse error = InvalidRequest
  - S3 continuation error = InvalidParameter
  - S3 invalid XML = InvalidRequest
  - S3 AJV validation errors = InvalidRequest
  - S3 stream errors = various... see: signature-v4-stream.ts
  - Iceberg errors
    - Map values from iceberg error format which differs from other storage api errors
- Fastify Validation (map to InvalidRequest)
  - FST_ERR_VALIDATION
  - FST_ERR_CTP_INVALID_JSON_BODY
  - FST_ERR_CTP_INVALID_CONTENT_LENGTH
  - FST_ERR_CTP_INVALID_MEDIA_TYPE
  - FST_ERR_CTP_BODY_TOO_LARGE

### DIFF
--- a/src/internal/errors/codes.ts
+++ b/src/internal/errors/codes.ts
@@ -149,11 +149,11 @@ export const ERRORS = {
       originalError: opts?.error,
     }),
 
-  InvalidRequest: (error: Error) =>
+  InvalidRequest: (message: string, error?: Error) =>
     new StorageBackendError({
       code: ErrorCode.InvalidRequest,
       httpStatusCode: 400,
-      message: error.message || 'Invalid Request',
+      message: message || 'Invalid Request',
       originalError: error,
     }),
 
@@ -565,8 +565,8 @@ const ERROR_CODE_MAP: Record<string, ErrorCode> = {
   FST_ERR_CTP_EMPTY_JSON_BODY: ErrorCode.InvalidRequest,
   FST_ERR_CTP_INVALID_JSON_BODY: ErrorCode.InvalidRequest,
   FST_ERR_CTP_INVALID_CONTENT_LENGTH: ErrorCode.InvalidRequest,
-  FST_ERR_CTP_INVALID_MEDIA_TYPE: ErrorCode.InvalidRequest,
-  FST_ERR_CTP_BODY_TOO_LARGE: ErrorCode.InvalidRequest,
+  FST_ERR_CTP_INVALID_MEDIA_TYPE: ErrorCode.InvalidMimeType,
+  FST_ERR_CTP_BODY_TOO_LARGE: ErrorCode.EntityTooLarge,
 }
 
 export function isStorageError(errorType: ErrorCode, error: unknown): error is StorageBackendError {
@@ -584,10 +584,12 @@ function hasStringErrorCode(error: Error): error is Error & { code: string } {
 function getErrorCode(error: Error): string {
   if (error instanceof IcebergError && error.error) {
     return error.error
-  } else if (hasStringErrorCode(error)) {
+  }
+  if (hasStringErrorCode(error)) {
     if (KNOWN_ERROR_CODES.has(error.code)) {
       return error.code as ErrorCode
-    } else if (error.code in ERROR_CODE_MAP) {
+    }
+    if (error.code in ERROR_CODE_MAP) {
       return ERROR_CODE_MAP[error.code]
     }
   }
@@ -597,9 +599,11 @@ function getErrorCode(error: Error): string {
 function getErrorStatusCode(error: Error): number {
   if (error instanceof StorageBackendError && error.httpStatusCode) {
     return error.httpStatusCode
-  } else if (error instanceof IcebergError) {
+  }
+  if (error instanceof IcebergError) {
     return error.code
-  } else if (hasNumericStatusCode(error)) {
+  }
+  if (hasNumericStatusCode(error)) {
     // Fastify validation errors include statusCode we can use
     return error.statusCode
   }
--- a/src/internal/errors/codes.ts
+++ b/src/internal/errors/codes.ts
@@ -59,6 +59,7 @@ export enum ErrorCode {
   NoAvailableShard = 'NoAvailableShard',
   ShardNotFound = 'ShardNotFound',
 }
+const KNOWN_ERROR_CODES = new Set<string>(Object.values(ErrorCode))
 
 export const ERRORS = {
   BucketNotEmpty: (bucket: string, e?: Error) =>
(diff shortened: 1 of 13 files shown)


src/internal/errors/codes.ts
```diff
@@ -550,30 +559,70 @@ export const ERRORS = {
...
+}
+
+export function normalizeRawError(error: unknown, logLevel: string) {
+  if (error instanceof Error) {
+    const statusCode = getErrorStatusCode(error)
+    const errorCode = getErrorCode(error)
+    const includeStack =
+      logLevel === 'debug' || statusCode >= 500 || errorCode === ErrorCode.UnknownError
+        ? true
+        : false
```
ferhatelmas:
```suggestion
    const includeStack = logLevel === 'debug' || statusCode >= 500 || errorCode === ErrorCode.UnknownError
```

src/internal/errors/codes.ts
```diff
@@ -550,30 +559,70 @@ export const ERRORS = {
...
+    code = error.error
+  } else if (hasStringErrorCode(error)) {
+    if (error.code in ErrorCode) {
+      code = error.code as ErrorCode
+    } else if (error.code in ERROR_CODE_MAP) {
+      code = ERROR_CODE_MAP[error.code]
     }
+  }
+  return code
+}
```
ferhatelmas, line 597: I would use 

```js
if (...)
   return
if (...)
   return
return default
```

ferhatelmas (approved):

src/internal/errors/codes.ts
```diff
@@ -550,30 +559,70 @@ export const ERRORS = {
...
+function hasStringErrorCode(error: Error): error is Error & { code: string } {
+  return 'code' in error && typeof error.code === 'string'
+}
+
+function getErrorCode(error: Error): string {
+  let code: string = ErrorCode.UnknownError
+  if (error instanceof IcebergError && error.error) {
+    code = error.error
+  } else if (hasStringErrorCode(error)) {
+    if (error.code in ErrorCode) {
```
ferhatelmas: this doesn't work for every error if key and value are different than each other because it's checking enum key but `error.code` is the value here

for example `S3VectorConflictException = 'ConflictException'`

ferhatelmas (changes requested):
requesting changes to prevent merge by mistake

src/internal/errors/codes.ts
itslenny: good call... I had the variable because these were originally in the normalize fn and I broke them out into their own function

src/internal/errors/codes.ts
```diff
@@ -550,30 +560,66 @@ export const ERRORS = {
   },
 }
 
-export function isStorageError(errorType: ErrorCode, error: any): error is StorageBackendError {
+const ERROR_CODE_MAP: Record<string, ErrorCode> = {
+  FST_ERR_VALIDATION: ErrorCode.InvalidRequest,
+  FST_ERR_CTP_EMPTY_JSON_BODY: ErrorCode.InvalidRequest,
+  FST_ERR_CTP_INVALID_JSON_BODY: ErrorCode.InvalidRequest,
+  FST_ERR_CTP_INVALID_CONTENT_LENGTH: ErrorCode.InvalidRequest,
+  FST_ERR_CTP_INVALID_MEDIA_TYPE: ErrorCode.InvalidRequest,
```
ferhatelmas: `InvalidMimeType` ?

src/internal/errors/codes.ts
```diff
@@ -550,30 +560,66 @@ export const ERRORS = {
...
+  FST_ERR_CTP_BODY_TOO_LARGE: ErrorCode.InvalidRequest,
```
ferhatelmas: similarly `EntityTooLarge`?

src/internal/errors/codes.ts
ferhatelmas: nit: I would even drop `else` in `else if`

src/internal/errors/codes.ts
```diff
@@ -147,6 +149,14 @@ export const ERRORS = {
       originalError: opts?.error,
     }),
 
+  InvalidRequest: (error: Error) =>
```
ferhatelmas: maybe `InvalidRequest(message: string, e?: Error)` to align with other factories

src/internal/errors/codes.ts
```diff
@@ -550,30 +560,66 @@ export const ERRORS = {
...
+    const errorCode = getErrorCode(error)
+    const includeStack =
+      logLevel === 'debug' || statusCode >= 500 || errorCode === ErrorCode.UnknownError
 
     return {
       raw: JSON.stringify(error),
       name: error.name,
       message: error.message,
-      stack: error.stack,
+      stack: includeStack ? error.stack || '' : '',
```
ferhatelmas, line 624: shall we consider not logging at all if not required?

src/internal/errors/codes.ts
```diff
@@ -550,30 +560,66 @@ export const ERRORS = {
...
-    } else if (hasStatusCode(error)) {
-      // Fastify validation errors include statusCode we can use
-      statusCode = error.statusCode
+function hasStringErrorCode(error: Error): error is Error & { code: string } {
+  return 'code' in error && typeof error.code === 'string'
+}
+
+function getErrorCode(error: Error): string {
+  if (error instanceof IcebergError && error.error) {
+    return error.error
```
ferhatelmas, line 586: shall we do mapping here to canonical values? Not a must but logged `errorCode` will be a bigger set otherwise.

src/internal/errors/codes.ts
```diff
@@ -58,6 +59,7 @@ export enum ErrorCode {
   NoAvailableShard = 'NoAvailableShard',
   ShardNotFound = 'ShardNotFound',
 }
+const KNOWN_ERROR_CODES = new Set<string>(Object.values(ErrorCode))
```
ferhatelmas, line 62: I would suggest having a unit test with a bad example of this change fixed

src/internal/errors/codes.ts
itslenny: I think for now we should stick to just dropping the stack traces. Seeing logs for 4xx can be useful and expected by customers already. Changing that would need a bigger discussion.

src/internal/errors/codes.ts
itslenny: I think this is ok. These error codes come [from the iceberg spec](https://github.com/apache/iceberg/blob/57409faedb184eb3b410a9cbbab52a4e5c0b6f0a/open-api/rest-catalog-open-api.yaml#L5155) and it's probably a good idea to keep the codes consistent between the responses and logs. We can add a section to the errors doc that lists the iceberg error codes too

src/internal/errors/codes.ts
ferhatelmas: Yes, I agree about the log line. 

I was suggesting not adding `stack` field instead of empty string. 

I think we need a pass to reduce duplication. For example, `tenantId` is logged in many levels (`project`, `context.project`, `context.tenantId`)

src/internal/errors/codes.ts
ferhatelmas: Sounds good to me

ferhatelmas (approved):

PR #1064, opened 2026-04-28
fix: generate swagger ui with script

itslenny:

Bug fix

- Swagger ui github action is outdated and [fails to generate ui](https://github.com/supabase/storage/actions/runs/25011561045/job/73248354247)
- Vector and Iceberg endpoints not included in spec

## What is the new behavior?

- Replace github action with script that generates ui using existing fastify/swagger-ui
- Enable vector and iceberg when generating spec so they're included

### DIFF
--- a/src/scripts/export-docs-ui.ts
+++ b/src/scripts/export-docs-ui.ts
@@ -20,4 +20,7 @@ function swaggerInitializer(targetUrl: string) {
   await fs.cp(swaggerPath, targetPath, { recursive: true })
   await fs.cp(specPath, resolve(targetPath, specFile))
   await fs.writeFile(resolve(targetPath, 'swagger-initializer.js'), swaggerInitializer(specFile))
-})().catch(console.error)
+})().catch((e) => {
+  console.error(e)
+  process.exit(1)
+})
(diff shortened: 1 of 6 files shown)


TylerHillery (approved):

src/scripts/export-docs-ui.ts
```diff
@@ -0,0 +1,23 @@
...
+
+;(async () => {
+  const targetPath = 'swagger-ui'
+  const swaggerPath = resolve('node_modules', '@fastify', 'swagger-ui', 'static')
+  const specFile = 'api.json'
+  const specPath = resolve('static', specFile)
+  await fs.cp(swaggerPath, targetPath, { recursive: true })
+  await fs.cp(specPath, resolve(targetPath, specFile))
+  await fs.writeFile(resolve(targetPath, 'swagger-initializer.js'), swaggerInitializer(specFile))
+})().catch(console.error)
```
ferhatelmas: shall we not catch and break the build on failure? or set a non-zero exit code?

src/scripts/export-docs-ui.ts
```diff
@@ -0,0 +1,23 @@
...
+      dom_id: '#swagger-ui',
+      deepLinking: true,
+      url: '${targetUrl}'
+    }
+    const ui = SwaggerUIBundle(config)
+  }`
+}
+
+;(async () => {
+  const targetPath = 'swagger-ui'
```
ferhatelmas, line 16: shall we add into gitignore as well?

PR #1069, opened 2026-04-29
fix: allow PUT tenant to clear database_pool_url

### LINKED ISSUE #1068: Admin PUT /tenants/:id can't clear database_pool_url or database_pool_mode
The PUT handler in `src/http/routes/admin/tenants.ts` builds `tenantInfo` with truthy guards:

```ts
if (databasePoolUrl) {
  tenantInfo.database_pool_url = encrypt(databasePoolUrl)
}

if (databasePoolMode) {
  tenantInfo.database_pool_mode = databasePoolMode
}
```

The body schema marks both as `nullable: true`, but a request with `databasePoolUrl: null` or `databasePoolMode: null` skips the assignment entirely. The merge upsert then leaves the existing column unchanged, so operators can't reset pool config back to the default.

The PATCH handler at lines 396-401 already does the three-way check correctly. PUT just needs the same.

Repro:

```sh
curl -X PUT 'http://localhost:5000/tenants/abc' \
  -H "apikey: $ADMIN_API_KEYS" \
  -H 'content-type: application/json' \
  -d '{
    "anonKey": "...", "databaseUrl": "...", "jwtSecret": "...", "serviceKey": "...",
    "databasePoolUrl": "postgresql://...",
    "databasePoolMode": "transaction"
  }'

curl -X PUT 'http://localhost:5000/tenants/abc' \
  -H "apikey: $ADMIN_API_KEYS" \
  -H 'content-type: application/json' \
  -d '{
    "anonKey": "...", "databaseUrl": "...", "jwtSecret": "...", "serviceKey": "...",
    "databasePoolUrl": null,
    "databasePoolMode": null
  }'

# database_pool_url and database_pool_mode are still the encrypted/string values from the first call.
```

oniani1:
The PUT `/tenants/:id` handler used truthy guards on `databasePoolUrl` and `databasePoolMode`, so a request with `null` would skip the assignment and the merge upsert left the existing column unchanged. The schema declares both fields `nullable: true`, and the PATCH handler at lines 396-401 already does the three-way check.

This switches both checks to `!== undefined` and treats explicit `null` as a clear, mirroring PATCH. `tenantDBInterface` was widened to allow `string | null` on the two columns so the assignment typechecks.

`admin-tenants.test.ts` adds a case that PUTs a tenant with both pool fields populated, verifies the encrypted/string values land in the DB, then PUTs again with `null` for both and asserts the columns are now `NULL`. Verified locally against `multitenant_db` running in Docker; the case fails on master and passes with the fix.

Closes #1068

### DIFF
--- a/acceptance/specs/admin.test.ts
+++ b/acceptance/specs/admin.test.ts
@@ -36,7 +36,6 @@ interface TenantFeatures {
 interface TenantDetailResponse {
   anonKey?: string
   capabilities?: Record<string, unknown>
-  databasePoolMode?: string | null
   databasePoolUrl?: string | null
   databaseUrl?: string
   fileSizeLimit?: number
@@ -572,10 +571,6 @@ function buildTenantProvisionBody(
     fileSizeLimit,
     jwtSecret: requireString(sourceTenant?.jwtSecret, 'jwtSecret'),
     serviceKey: requireString(sourceTenant?.serviceKey, 'serviceKey'),
-  }
-
-  if (typeof sourceTenant?.databasePoolMode === 'string') {
-    body.databasePoolMode = sourceTenant.databasePoolMode
   }
 
   if (typeof sourceTenant?.databasePoolUrl === 'string') {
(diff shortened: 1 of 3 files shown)


ferhatelmas:
Clearing pool_url isn't useful but for completeness, we still could get it. pool_mode is dropped. If you want to update this, feel free. Otherwise, good to close.

ferhatelmas:
closing this per previous comment

ferhatelmas:
on second thought, I will update this and merge

ferhatelmas (approved):

PR #1070, opened 2026-04-29
chore: use official github pages deploy

itslenny:

Chore

Uses 3rd party action which depends on deprecated node to deploy github pages

## What is the new behavior?

Use github official github pages action

## Additional context

https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,6 +9,11 @@ on:
 permissions:
   contents: write
   pages: write
+  id-token: write
+
+concurrency:
+  group: "pages-deploy"
+  cancel-in-progress: false
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -56,8 +61,14 @@ jobs:
           VECTOR_ENABLED: true
           ICEBERG_ENABLED: true
 
-      - name: Deploy to GitHub Pages
-        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
+      - name: Setup Pages
+        uses: actions/configure-pages@v5
```
ferhatelmas: I think it won't work when this build runs because our org requires exact hashes in version

ferhatelmas: related https://github.com/supabase/storage/pull/925

ferhatelmas (approved):

PR #1094, opened 2026-05-10
feat: embedded vector store

fenos:

Feature

Currently, Vector buckets only support S3Vector, which is a cloud service and not self-hostable.
For local development and self-hosting, this might be impossible or very difficult to set up and operate. 

## What is the new behavior?

- Supporting vector buckets locally for easy development
- It uses **pgvector** as an embedded vector store
- We use the same client api


package-lock.json
```diff
@@ -102,6 +102,9 @@
       "engines": {
         "node": ">=24.0.0",
         "npm": "11.12.1"
+      },
```
ferhatelmas: we shouldn't have any changes in the lock, this is leftover from previous iteration

src/storage/protocols/vector/adapter/pgvector/filter.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+    case '$exists': {
+      if (typeof raw !== 'boolean') {
+        throw ERRORS.InvalidParameter(`$exists requires a boolean for "${fieldName}"`)
+      }
+      if (!VALID_IDENTIFIER.test(fieldName)) {
+        throw ERRORS.InvalidParameter(`Invalid metadata field name: ${fieldName}`)
+      }
+      const key = placeholder(ctx, fieldName)
+      return raw ? `${ctx.column} ? ${key}` : `NOT (${ctx.column} ? ${key})`
```
ferhatelmas: I think this (`?`) will generate broken SQL due to replacement. Easy fix is to use function form `jsonb_exists`. We have a test gap because unit tests don't check the final SQL after replacement

src/storage/protocols/vector/adapter/pgvector/filter.ts
```diff
@@ -0,0 +1,210 @@
...
+  '$gt',
+  '$gte',
+  '$lt',
+  '$lte',
+  '$in',
+  '$nin',
+  '$exists',
+])
+
+const VALID_IDENTIFIER = /^[A-Za-z_][A-Za-z0-9_]*$/
```
ferhatelmas: Isn't this overly restrictive for what we accept in [put](https://github.com/supabase/storage/blob/master/src/http/routes/vector/put-vectors.ts#L34-L39) and [query](https://github.com/supabase/storage/blob/master/src/http/routes/vector/query-vectors.ts#L17)? It means customers will put some metadata which they won't be able to query. I think we need to relax it and to parametrize keys as well

src/storage/protocols/vector/adapter/pgvector/filter.ts
```diff
@@ -0,0 +1,210 @@
...
+    case '$gt':
+    case '$gte':
+    case '$lt':
+    case '$lte': {
+      if (typeof raw !== 'number') {
+        throw ERRORS.InvalidParameter(`${op} requires a number for "${fieldName}"`)
+      }
+      checkFinite(raw)
+      const opSql = { $gt: '>', $gte: '>=', $lt: '<', $lte: '<=' }[op]
+      return `${numericField(ctx, fieldName)} ${opSql} ${placeholder(ctx, raw)}`
```
ferhatelmas: Since we do a cast here, do we need a guard to skip that row if value isn't numeric?

src/test/pgvector-adapter.test.ts
```diff
@@ -0,0 +1,180 @@
...
+      client: 'pg',
+      connection: { connectionString: TEST_DATABASE_URL, connectionTimeoutMillis: 2_000 },
+      pool: { min: 0, max: 1 },
+    })
+    try {
+      await probe.raw('CREATE EXTENSION IF NOT EXISTS vector')
+      await probe.raw('CREATE SCHEMA IF NOT EXISTS storage_vectors')
+      pgvectorAvailable = true
+    } catch (e) {
+      // eslint-disable-next-line no-console
```
ferhatelmas: we don't use eslint, this line should be unnecessary

src/storage/protocols/vector/adapter/pgvector/index.ts
```diff
@@ -0,0 +1,375 @@
...
+        await db.raw(
+          `CREATE TABLE ${SCHEMA}.??
+           (
+             key text PRIMARY KEY,
+             embedding vector(${dimension}) NOT NULL,
+             metadata jsonb NOT NULL DEFAULT '{}'::jsonb
+           )`,
+          [table]
+        )
+        await db.raw(`CREATE INDEX ?? ON ${SCHEMA}.?? USING hnsw (embedding ${choice.opClass})`, [
```
ferhatelmas: re: https://github.com/pgvector/pgvector/issues/461#issuecomment-2680114052

16k is fine here?

fenos:
@copilot review again

TylerHillery (approved):

PR #1096, opened 2026-05-11
fix: replace knex with pg directly

ferhatelmas:

Bug fix

Knex is heavy and generating too much garbage in multitenant setup.

## What is the new behavior?

Drop knex, knex-telemetry, knex-patch and patch-package in favor of direct postgres connection layer. 

## Additional context

Improve direct query cancellation and mark aborted connection not to return bad to the pool. Type abort error.
Cap pool drain with timeout.
Object lock uses CTE, no promise.race.
Normalize pool mode (`recycled` vs `recycle`).
Invalidate cache from admin. Max connections recreate the pool, safer.
Fixed reset migration lock. Admin migrations handlers apply limits correctly.
Improve ordering in shutdown by reversing (server, tenants, multitenant) because tenants can need multitenant.
Use request scope db from iceberg/vector.
Fix connection leak in bucket deleted.
Fix scanner batching waiting.
Quote identifiers.
Plumb ssl settings to MT db.
Clear owner_id at the same time with owner.
Validate database pool mode in admin handler.


src/internal/database/multitenant-pg.ts
```diff
@@ -0,0 +1,119 @@
...
+const multitenantPgPoolOwner = new MultitenantPgPoolOwner()
+
+export const multitenantPgPool = new Proxy({} as Pool, {
+  get(_target, property) {
+    const pool = multitenantPgPoolOwner.getPool()
+    const value = Reflect.get(pool, property, pool)
+
+    return typeof value === 'function' ? value.bind(pool) : value
+  },
+})
```
mcollina: You should avoid using Proxy, it's a bottleneck.

ferhatelmas: addressed

src/storage/database/pg.ts
```diff
@@ -0,0 +1,1963 @@
...
+const { isMultitenant } = getConfig()
+// Scanner cache tables are unlogged scratch tables, not session temp tables.
+// They must survive across pooled pg clients used by separate unscoped queries.
+const S3_KEYS_SCRATCH_TABLE_SCHEMA = 'storage'
+const S3_KEYS_SCRATCH_TABLE_PREFIX = '_s3_remote_keys_'
+const S3_KEYS_SCRATCH_TABLE_MAX_AGE_MS = 24 * 60 * 60 * 1000
+const S3_KEYS_SCRATCH_TABLE_PATTERN = `^${S3_KEYS_SCRATCH_TABLE_PREFIX}([0-9]{13})(?:_[A-Za-z0-9_]+)?$`
+
+export function escapeLike(str: string) {
+  return str.replace(/([%_])/g, '\\$1')
```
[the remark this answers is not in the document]
ferhatelmas: addressed

src/storage/database/pg.ts
```diff
@@ -0,0 +1,2031 @@
...
+              value
+            FROM set_lock_timeout
+          ),
+          restore_lock_timeout AS MATERIALIZED (
+            SELECT set_config('lock_timeout', value, true) AS restored_timeout
+            FROM acquire_lock
+          )
+          SELECT true AS locked
+          FROM restore_lock_timeout
+        `
```
ferhatelmas, line 1220: @GuptaManan100 this is failing with multigres and made [above workaround](https://github.com/supabase/storage/pull/1096/changes#diff-194f2b802e82311a34463f6eaea3c82c0e1ec24bebb58ed1e031fa57d19b8bfdR1187-R1190). Wanted to give a heads up if you're not aware

GuptaManan100: Yes we are working on supporting advisory locks so that they work too. Right now we don't track them so queries end up going to different connections and that breaks the entire model

fenos (approved):

PR #1097, opened 2026-05-11
fix(auth): pass alg to importJWK for jose v6 compatibility with Auth0…

cameron-michie:
Bug fix

Any authenticated Storage request fails with HTTP 403 when the configured JWKS contains RSA keys that omit the `"alg"` field.

The error thrown is [from here](https://github.com/panva/jose/blob/9c8658647f1b9c3fe4d7d15812ddc51fedbf2416/src/lib/jwk_to_key.ts#L107): `"alg" argument is required when "jwk.alg" is not present`

jose v6 removed algorithm inference from `importJWK()`. It now requires either `jwk.alg` to be set on the key object, or the algorithm passed explicitly as the second argument. Storage was calling `importJWK(jwk)` without either, causing jose to throw before the key could be imported.

## What is the new behavior?

`importJWK` is called as `importJWK(jwk, header.alg)`

## Additional context

n/a

### DIFF
--- a/src/internal/auth/jwt.ts
+++ b/src/internal/auth/jwt.ts
@@ -103,14 +103,22 @@ async function findJWKFromHeader(
       return encoder.encode(secret)
     }
 
-    // find the first key without a kid or with the matching kid, the "oct" type
-    const jwk = jwks.keys.find(
-      (key) => (!key.kid || key.kid === header.kid) && key.kty === 'oct' && key.k
-    )
+    // find the first compatible "oct" key without a kid or with the matching kid
+    let mismatchedJwk: JwksConfigKey | undefined
+    const jwk = jwks.keys.find((key) => {
+      if ((!key.kid || key.kid === header.kid) && key.kty === 'oct' && key.k) {
+        if (key.alg !== undefined && key.alg !== header.alg) {
+          mismatchedJwk ??= key
+          return false
+        }
+        return true
+      }
+      return false
+    })
 
-    if (jwk?.alg && jwk.alg !== header.alg) {
+    if (!jwk && mismatchedJwk) {
       throw ERRORS.AccessDenied(
-        `JWT algorithm "${header.alg}" does not match JWK algorithm "${jwk.alg}"`
+        `JWT algorithm "${header.alg}" does not match JWK algorithm "${mismatchedJwk.alg}"`
       )
     }
 
@@ -136,7 +144,7 @@ async function findJWKFromHeader(
     return (
       ((!key.kid && !header.kid) || key.kid === header.kid) &&
       key.kty === kty &&
-      (!key.alg || key.alg === header.alg)
+      (key.alg === undefined || key.alg === header.alg)
     )
   })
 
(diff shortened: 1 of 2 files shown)


mathiasGM:
Thanks a lot for looking into this and opening the PR, @cameron-michie. Really appreciate it!
We’ve been able to reproduce the issue consistently on our side as well, so it definitely seems like a real bug.

This would be a very valuable fix for us, and happy to help test or provide more context if useful.

src/internal/auth/jwt.ts
```diff
@@ -90,7 +90,7 @@ async function findJWKFromHeader(
     // couldn't find a matching JWK, try to use the secret
     return encoder.encode(secret)
   }
-  return await importJWK(jwk)
+  return await importJWK(jwk, header.alg)
```
ferhatelmas: Now, we also need to fix the jwk filtering above because otherwise jwk with a particular algorithm will verify a token with an algorithm that isn't created for. For example, jwk says `RS256` but header says `RS512`

And it's also better to pass `jwk.alg || header.alg` here for clearer intention

src/internal/auth/jwt.ts
```diff
@@ -90,7 +90,7 @@ async function findJWKFromHeader(
...
 }
```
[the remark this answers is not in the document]
ferhatelmas: yes, we need to see some tests as well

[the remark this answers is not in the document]
ferhatelmas: * accept asymmetric jwt when matching jwk has no alg
* reject asymmetric jwt when matching jwk has a different alg
* reject hmac jwt when matching jwk has a different alg

ferhatelmas (changes requested):

ferhatelmas:
@cameron-michie main idea of the change is good but it's too permissive. Let me know if you can work on additional changes I commented. Happy to give a hand otherwise.

cameron-michie:
Hi @ferhatelmas, thanks so much for the review and for the comments. I gave the changes you requested a go, and added the tests. Could you take another look please?

src/internal/auth/jwt.ts
```diff
@@ -103,9 +103,13 @@ async function findJWKFromHeader(
...
-    // find the first key without a kid or with the matching kid and the "oct" type
+    // find the first key without a kid or with the matching kid, the "oct" type, and a compatible alg
     const jwk = jwks.keys.find(
-      (key) => (!key.kid || key.kid === header.kid) && key.kty === 'oct' && key.k
+      (key) =>
+        (!key.kid || key.kid === header.kid) &&
+        key.kty === 'oct' &&
+        key.k &&
+        (!key.alg || key.alg === header.alg)
     )
```
ferhatelmas: if we match by kid but alg is different, we should reject  here, not fallback to static

src/internal/auth/jwt.ts
```diff
@@ -125,16 +129,20 @@ async function findJWKFromHeader(
...
+      (!key.alg || key.alg === header.alg)
+    )
   })
 
   if (!jwk) {
     // couldn't find a matching JWK, try to use the secret
     return encoder.encode(secret)
   }
-  return await importJWK(jwk)
+  return await importJWK(jwk, jwk.alg || header.alg)
```
ferhatelmas: shall we use defined instead of empty/falsy?

ferhatelmas:
@claude review always

ferhatelmas (approved):

itslenny (approved):

PR #947, opened 2026-03-30
fix: improved error handling and normalization

itslenny:
## What kind of change does this PR introduce?

Fix to add additional error handling and normalization

## What is the current behavior?

- Some 400 errors include `statusCode: 0` in the json body
- Some of the most common 500 errors aren't true 500s

## What is the new behavior?

- Add missing statusCodes to json body for 400 level errors
- Handle specific 500 errors with more accurate errors / codes

## Additional Context

**400 ERRORS**

Add missing internal `statusCode` value to error for the following errors (currently statusCode=0):

* Fastify validation
  * FST_ERR_VALIDATION = 400
  * FST_ERR_CTP_EMPTY_JSON_BODY = 400
  * FST_ERR_CTP_INVALID_JSON_BODY = 400
  * FST_ERR_CTP_INVALID_CONTENT_LENGTH = 400
  * FST_ERR_CTP_INVALID_MEDIA_TYPE = 415
* S3 protocol validation
  * S3 protocol body validation = 400

**500 ERRORS**

Better error codes for 500s that aren't true 500s

* S3 Chunk size exceeds 8388608 bytes = 400
* Project database errors
  * 08P01 - no more connections allowed (max_client_conn) = 503
  * 25006 - read only sql transaction = 503
  * 22P02 - invalid text representation = 503
  * 42703 - undefined column = 503
  * 42P01 - relation does not exist = 503

PR #1106, opened 2026-05-13
fix: correctly categorize errors related to saturated project databases

itslenny:

Bug fix

These errors caused by projects under extreme load throw as 500 errors

- 'server login has been failing, cached error: connect timeout (server_login_retry)',
- 'server login has been failing, cached error: the database system is not accepting connections (server_login_retry)',

This [previous change](https://github.com/supabase/storage/pull/947/changes#diff-230c7c8b84befeba072a468151e75decacdc7e40dc9a99f08fa41ef862b200dcR138) miscategorized `no more connections allowed` as a 503. Without this it is correctly handled in error-hanler.ts as a 429

## What is the new behavior?

Errors related to project databases being overloaded throw as `429 slow down`

### DIFF
--- a/src/http/error-handler.ts
+++ b/src/http/error-handler.ts
@@ -33,6 +33,8 @@ export const setErrorHandler = (
         'no more connections allowed',
         'sorry, too many clients already',
         'server login has been failing, try again later',
+        'server login has been failing, cached error: connect timeout (server_login_retry)',
+        'server login has been failing, cached error: the database system is not accepting connections (server_login_retry)',
       ].some((msg) => (error as DatabaseError).message.includes(msg))
     ) {
       return reply.status(429).send(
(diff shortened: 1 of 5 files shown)


src/http/error-handler.ts
```diff
@@ -33,6 +33,8 @@ export const setErrorHandler = (
         'no more connections allowed',
         'sorry, too many clients already',
         'server login has been failing, try again later',
+        'server login has been failing, cached error: connect timeout (server_login_retry)',
+        'server login has been failing, cached error: the database system is not accepting connections (server_login_retry)',
```
ferhatelmas: We have a similar list for s3 error handler, can we update it as well?

It might be better to extract this into a helper and use in both places

ferhatelmas (approved):

PR #590, opened 2024-11-20
fix: improve error message for insufficient privilege

fenos:
## What kind of change does this PR introduce?

Bug fix

## What is the new behavior?

Make the error message more descriptive when insufficient privileges error is thrown by postgres

PR #1107, opened 2026-05-14
fix: improve database error messages

itslenny:

Bug fix

All schema mismatch errors (missing column or missing table) show an error message indicating the storage migrations are out of sync. However, user created triggers/rls can cause these errors resulting in confusion and support burden

All permission denied errors are shown as RLS errors

## What is the new behavior?

- Show error details for errors originating inside of triggers / rls
- Only show "RLS" error for "RLS" errors. For other access denied pass a more detailed error

### schema mismatch in RLS

```
There is a database schema mismatch in a trigger or RLS policy: PL/pgSQL function raise_42703() line 3 at EXECUTE
```

### schema mismatch in trigger

```
There is a database schema mismatch in a trigger or RLS policy: PL/pgSQL function storage_objects_broken_trigger() line 3 at PERFORM
```

### Non rls access denied

```
insert into \"objects\" (\"bucket_id\", \"metadata\", \"name\", \"owner\", \"owner_id\", \"user_metadata\", \"version\") values ($1, $2, $3, $4, $5, DEFAULT, $6) - permission denied for table secret_table
```

### DIFF
--- a/src/storage/database/knex.ts
+++ b/src/storage/database/knex.ts
@@ -1145,7 +1145,9 @@ export class DBError extends StorageBackendError implements RenderableError {
     switch (pgError.code) {
       case '42501':
         return ERRORS.AccessDenied(
-          'new row violates row-level security policy',
+          pgError.message.includes('row-level security')
+            ? 'new row violates row-level security policy'
+            : pgError.message,
           pgError
         ).withMetadata({
           query,
--- a/src/internal/errors/codes.ts
+++ b/src/internal/errors/codes.ts
@@ -1,4 +1,5 @@
 import { IcebergError } from '@storage/protocols/iceberg/catalog/errors'
+import { DatabaseError } from 'pg'
 import { StorageBackendError } from './storage-error'
 
 export enum ErrorCode {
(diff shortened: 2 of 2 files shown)


src/storage/database/knex.ts
```diff
@@ -1145,7 +1145,9 @@ export class DBError extends StorageBackendError implements RenderableError {
     switch (pgError.code) {
       case '42501':
         return ERRORS.AccessDenied(
-          'new row violates row-level security policy',
+          pgError.message.includes('row-level security')
+            ? 'new row violates row-level security policy'
```
ferhatelmas, line 1149: Shall we update to this error message: https://github.com/supabase/storage/pull/590

itslenny: I don't think so / I think this version is better (but certainly open to the discussion). The RLS error is correct as-is when it is actually an RLS error, but there are other types of access denied getting lumped into this message (usually indicating a bigger schema/role/grant issue)

ferhatelmas: sounds good to me, then I think we should close the other PR

fenos (approved):

ferhatelmas (approved):

PR #1110, opened 2026-05-18
feat: Added separate Database application.

ShogunPanda:


watt.database.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://schemas.platformatic.dev/@platformatic/node/3.52.0.json",
+  "node": {
+    "main": "../../dist/database/index.js",
+    "hasServer": false
+  }
+}
```
mcollina, line 7: I would probably put this file in the root of the project

src/database/test-server.ts
```diff
@@ -0,0 +1,226 @@
+import { randomUUID } from 'node:crypto'
```
mcollina: what is this test server

src/internal/database/watt-connection.ts
```diff
@@ -0,0 +1,411 @@
...
+      isolationLevel: options?.isolation,
+      operationName: this.operation?.(),
+      readOnly: options?.readOnly,
+    })
+
+    return new WattPgTransaction(response.lockId, this.operation)
+  }
+}
+
+class WattPgTransaction extends PgTransaction {
```
mcollina: Should these have interfaces instead of classes we fully override?

src/internal/database/watt-connection.test.ts
```diff
@@ -0,0 +1,226 @@
...
+      send: vi.fn(async (application: string, message: string, data: Record<string, unknown>) => {
+        sent.push({ application, data, message })
+        const response = responses[message]
+        return response instanceof Promise ? response : response
+      }),
+    },
+  }
+
+  return { sent }
+}
```
mcollina: we should be providing this utility in our repo

src/internal/database/watt-connection.ts
```diff
@@ -0,0 +1,399 @@
...
+  const requestId = typeof data.requestId === 'string' ? data.requestId : randomUUID()
+  const payload = { ...data, requestId }
+  let abortListener: (() => void) | undefined
+  let settled = false
+
+  const request = getMessaging().send(databaseApplicationId, message, payload)
+  const abortPromise = new Promise<never>((_, reject) => {
+    if (!signal) {
+      return
+    }
```
mcollina: we are allocating this promise unnecessarily in this case

src/internal/database/watt-connection.ts
```diff
@@ -0,0 +1,399 @@
...
+
+type PgQueryArgument = PgQueryOptions | unknown[]
+
+const databaseApplicationId = 'database'
+
+export function hasDatabaseWattMessaging(): boolean {
+  const platformatic = (globalThis as typeof globalThis & { platformatic?: PlatformaticGlobal })
+    .platformatic
+  return Boolean(platformatic?.messaging)
+}
```
mcollina, line 47: this verifies it runs on Watt, not if there is a database app running

PR #1115, opened 2026-05-20
Encapsulate storage database transaction APIs

mcollina:
## Summary
- remove generic transaction/permission callback utilities from the public storage `Database` interface
- replace them with domain-specific database operations for permission checks, bucket/object mutations, uploads, multipart progress, and TUS lock acquisition
- keep external side effects such as backend deletes and webhook dispatch outside database transaction helpers
- narrow the newly added database operation result types instead of returning broad `Obj` shapes

## Validation
- `npx biome check src/storage/database/adapter.ts src/storage/database/knex.ts src/storage/events/objects/object-admin-delete-all-before.ts src/storage/object.ts src/storage/protocols/s3/s3-handler.ts src/storage/protocols/tus/postgres-locker.ts src/storage/storage.ts src/storage/uploader.ts src/test/s3-protocol.test.ts`
- `npm run build`

### DIFF
--- a/src/storage/database/knex.ts
+++ b/src/storage/database/knex.ts
@@ -64,10 +64,10 @@ export class StorageKnexDB implements Database {
     this.latestMigration = options.latestMigration
   }
 
-  async withTransaction<T>(
-    fn: (db: Database) => Promise<T>,
+  private async withTransaction<T extends (db: Database) => Promise<unknown>>(
+    fn: T,
     opts?: TransactionOptions
-  ): Promise<T> {
+  ): Promise<Awaited<ReturnType<T>>> {
     const tnx = await this.connection.transactionProvider(this.options.tnx, opts)()
 
     try {
@@ -80,7 +80,7 @@ export class StorageKnexDB implements Database {
       const opts = { ...this.options, tnx }
       const storageWithTnx = new StorageKnexDB(this.connection, opts)
 
-      const result = await fn(storageWithTnx)
+      const result = (await fn(storageWithTnx)) as Awaited<ReturnType<T>>
       await tnx.commit()
       return result
     } catch (e) {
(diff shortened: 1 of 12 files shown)


src/storage/protocols/s3/s3-handler.ts
```diff
@@ -1283,12 +1266,10 @@ export class S3ProtocolHandler {
...
-    const [destinationBucket] = await this.storage.db.asSuperUser().withTransaction(async (db) => {
-      return Promise.all([
-        db.findBucketById(Bucket, 'file_size_limit'),
-        db.findBucketById(sourceBucketName, 'id'),
-      ])
-    })
+    const [destinationBucket] = await Promise.all([
+      this.storage.db.asSuperUser().findBucketById(Bucket, 'file_size_limit'),
+      this.storage.db.asSuperUser().findBucketById(sourceBucketName, 'id'),
+    ])
```
mcollina, line 1272: add a new method to do this in one go

src/storage/protocols/tus/postgres-locker.ts
```diff
@@ -73,42 +73,7 @@ export class PgLock implements Lock {
...
-            } finally {
-              abortController.abort()
-            }
-          },
-          {
-            timeout: 5 * 60 * 1000, // 5 minutes
-          }
-        )
-        .catch(reject)
+      this.holdLockTransaction(stopSignal, resolve).catch(reject)
```
mcollina, line 76: possible blocker, need to verify the S3 locker

src/storage/database/knex.ts
```diff
@@ -117,6 +118,215 @@ export class StorageKnexDB implements Database {
...
+  acquireObjectLockForTransaction(
+    bucketId: string,
+    objectName: string,
+    version: string | undefined,
+    transactionOptions?: TransactionOptions
+  ) {
+    return this.withTransaction(async (db) => {
+      await db.mustLockObject(bucketId, objectName, version)
+    }, transactionOptions)
+  }
```
mcollina, line 289: this is incorrect, the lock is gone after the transaction completes

PR #1127, opened 2026-05-26
fix: put-vector relax metadata validation

fenos:

Bug fix

Metadata can be of a single type only

## What is the new behavior?

Metadata can be of any of the types

## Additional context


ferhatelmas:
closing as addressed in #1132

PR #1131, opened 2026-06-02
ci: add Multigres integration test job

GuptaManan100:
## Summary

- Add a `Test / Multigres` CI job that runs storage's tenant migrations, dummy-data seeding, and the full integration suite against a Multigres cluster, mirroring the existing OrioleDB variant.
- Add `.docker/docker-compose-infra-multigres-override.yml`, a minimal override that swaps only the `tenant_db` service for the published `ghcr.io/multigres/multigres-cluster:latest` image; everything else in `docker-compose-infra.yml` is reused unchanged.
- Add the six `:multigres` npm scripts (mirroring `:oriole`) that the job and local runs use.

## Description

Storage's database-under-test is a single Compose service (`tenant_db`). This PR follows the same swappable-database pattern OrioleDB uses: a tiny override points `tenant_db` at a Multigres cluster, and parallel npm scripts (`test:multigres`, `test:multigres:ci`, plus the `infra:*:multigres` helpers) wire the existing `migration:run` / `test:dummy-data` / `test:integration` steps to it. The Multigres team publishes the all-in-one cluster image to GHCR, so the job pulls it directly (no build step) — matching how OrioleDB pulls a published image.

The override sets `MULTIGRES_GATEWAY_PG_PORT=5432` so storage reaches the gateway on its default port, `MULTIGRES_NUM_CELLS=2` (the minimum for the shard's AtLeastN(2) durability policy to elect a leader), and a psql-based healthcheck with a generous `start_period` since cluster bootstrap is slower than a bare postgres image.

Unlike `test_oriole`, this job intentionally omits the `pg_dump`-based migration-idempotency check. `pg_dump` (PG 17+) fails during startup introspection through the Multigres gateway because session-scoped `set_config()` rejects a name argument that is not a bare literal or a bound parameter — and `pg_dump` issues `SELECT set_config(name, 'view, foreign-table', false) FROM pg_settings WHERE name = 'restrict_nonsystem_relation_kind'`, whose name argument is a column reference. This is unrelated to storage's migration/test compatibility; details and a minimal repro are in `docs/multigres-pg-dump-set-config-bug.md`. The job can adopt the idempotency check once that introspection gap is closed.

## Validation

Ran the full flow locally against `ghcr.io/multigres/multigres-cluster:latest`:

- `tenant_db` becomes healthy after cluster bootstrap.
- `migration:run` applies all 61 tenant migrations cleanly.
- `test:dummy-data` seeds successfully.
- `test:integration` passes: 33 files, 685 passed, 10 skipped, 0 failed.


ferhatelmas:
we also need to add it to acceptance test matrix + sample env but I will address in a follow up PR, thanks for the support

ferhatelmas (approved):

ferhatelmas:
added `Test / Multigres` into required checks

GuptaManan100:
@ferhatelmas Thank you so much!

PR #1133, opened 2026-06-03
feat: add ability to purge cdn for whole buckets, tenant, and transformations

itslenny:

Feature

The CDN can only be purged for individual objects using the full object path.

There is no way to specifically purge transformations

## What is the new behavior?

CDN can be purged for individual objects or all objects in an entire bucket or entire tenant.

Also, supports purging just the transformations at all three levels.

### Endpoints

- `DELETE /storage/v1/cdn/bucket-name/path/to/an/object.ext` - **Existing purge endpoint**, purges single cached object
- `DELETE /storage/v1/cdn/bucket-name/path/to/an/object.ext?transformations=true` - Purge all cached transformations for a single object
- `DELETE /storage/v1/cdn/bucket-name` - Purge all cached objects in specified bucket
- `DELETE /storage/v1/cdn/bucket-name?transformations=true` - Purge all cached transformations associated with objects in specified bucket
- `DELETE /storage/v1/cdn` - Purge all cached objects for this tenant
- `DELETE /storage/v1/cdn?transformations=true` - Purge all cached transformations for this tenant

## Additional context

Requires deploy of updated CDN cache manager worker


src/storage/cdn/cdn-cache-manager.ts
```diff
@@ -21,12 +21,48 @@ const defaultHeaders = new Headers({
...
+interface PurgeBucketInput extends BasePurgeInput {
+  type: 'bucket'
+  bucket: string
+}
+
+interface PurgeTenantInput extends BasePurgeInput {
+  type: 'tenant'
+}
+
+interface PurgeObjectTransformsInput extends BasePurgeInput {
```
ferhatelmas, line 43: why not extending `PurgeObjectInput`?

src/storage/cdn/cdn-cache-manager.ts
```diff
@@ -48,37 +84,55 @@ export class CdnCacheManager {
...
     if (!cdnPurgeUrl) {
       throw ERRORS.MissingParameter('CDN_PURGE_ENDPOINT_URL is not set')
     }
 
-    // Check if object exists
-    await this.storage.from(opts.bucket).asSuperUser().findObject(opts.objectName)
+    // Check if object exists (only for object-level purges)
+    if (opts.type === 'object' || opts.type === 'object-transforms') {
+      await this.storage.from(opts.bucket).asSuperUser().findObject(opts.objectName)
+    }
```
ferhatelmas: similarly do we need bucket existence check?

src/storage/cdn/cdn-cache-manager.ts
```diff
@@ -48,37 +84,55 @@ export class CdnCacheManager {
...
+
+    if ('bucket' in opts) {
+      requestBody.bucketId = opts.bucket
+    }
 
+    if ('objectName' in opts) {
+      requestBody.objectName = opts.objectName
+    }
+
+    console.log('PURGE CACHE!!!!!!!!... 2')
```
ferhatelmas: debugging logs to drop

src/test/cdn.test.ts
```diff
@@ -99,9 +99,132 @@ describe('CDN Cache Manager', () => {
...
     expect(body).toEqual({ message: 'success' })
     expect(purgeSpy).toHaveBeenCalledWith({
+      type: 'object',
       tenant: tenantId,
       bucket: bucketName,
       objectName,
     })
   })
+
+  it('will purge object transformations when transformations query param is true', async () => {
```
ferhatelmas, line 109: coverage is fine, but if you wish, we could add some acceptance tests as well

src/http/routes/cdn/purgeCache.ts
```diff
@@ -12,40 +12,120 @@ const purgeObjectParamsSchema = {
...
+  fastify.delete<PurgeTenantRequestInterface>(
+    '/',
+    {
+      schema: createDefaultSchema(successResponseSchema, {
+        querystring: purgeQuerySchema,
+        summary: 'Purge cache for entire tenant or tenant transformations',
+        tags: ['cdn'],
+      }),
+      config: {
+        operation: { type: ROUTE_OPERATIONS.PURGE_TENANT_CACHE },
```
ferhatelmas, line 63: These operations are always without transforms. Since you added transform operations, are you also planning to match operation via query params? Otherwise, I think, we should drop unused operations

src/storage/cdn/cdn-cache-manager.ts
itslenny: PurgeObjectInput has `type: 'object'` so PurgeObjectTransformsInput cannot extend it (without doing an Omit<type>) and it seemed cleaner to just stick to base object for everything.

acceptance/specs/cdn-render.test.ts
```diff
@@ -76,6 +76,108 @@ describeAcceptance(
...
+          expectedStatus: 200,
+          token: requireServiceKey(),
+        })
+        expect(purge.json).toMatchObject({
+          message: 'success',
+        })
+
+        const missingPurge = await client.request(
+          'DELETE',
+          `/cdn/${bucketName}/${encodePathSegments(missingBucketName)}`,
```
ferhatelmas:
```suggestion
          `/cdn/${missingBucketName}`,
```

otherwise it becomes object

src/test/cdn.test.ts
```diff
@@ -99,9 +99,132 @@ describe('CDN Cache Manager', () => {
...
+      objectName,
+    })
+  })
+
+  it('will purge entire bucket when using bucket endpoint', async () => {
+    const purgeSpy = vi.spyOn(CdnCacheManager.prototype, 'purge').mockResolvedValue(undefined)
+
+    const response = await appInstance.inject({
+      method: 'DELETE',
+      url: `/cdn/${bucketName}`,
```
ferhatelmas, line 150: can we add a test `/cdn/${bucketName}/`? It should call object wildcard path. We might want to exercise the contract

src/test/cdn.test.ts
```diff
@@ -99,9 +99,155 @@ describe('CDN Cache Manager', () => {
...
+
+    const body = await response.json()
+    expect(body).toEqual({ message: 'success' })
+    expect(purgeSpy).toHaveBeenCalledWith({
+      type: 'object',
+      tenant: tenantId,
+      bucket: bucketName,
+      objectName: '',
+    })
+  })
```
ferhatelmas: This resolves to object. If it's was bucket, we shouldn't have seen `objectName`

ferhatelmas:
Acceptance CDN stub is ignoring body shape at the moment. It didn't matter much previously since there was one shape but we might want to add validation. It also creates some coupling with another implementation and we already do in integration tests so feel free to ignore

src/test/cdn.test.ts
itslenny: The test name was a copy/paste error. This was the test you previously asked me to add to handle a trailing slash. It is counted as an object purge because of the trailing slash, but re-thinking this it should really be a 400 because an empty string is never a valid object name. I updated the validation and test.

ferhatelmas: For convenience, we could make hit to bucket route but current also works

ferhatelmas (approved):

src/storage/cdn/cdn-cache-manager.ts
```diff
@@ -52,21 +88,35 @@ export class CdnCacheManager {
...
     }
 
-    // Check if object exists
-    await this.storage.from(opts.bucket).asSuperUser().findObject(opts.objectName)
+    // Check if resource exists
+    if (opts.type === 'object' || opts.type === 'object-transforms') {
+      await this.storage.from(opts.bucket).asSuperUser().findObject(opts.objectName)
+    } else if (opts.type === 'bucket' || opts.type === 'bucket-transforms') {
+      await this.storage.asSuperUser().findBucket(opts.bucket)
+    }
```
ferhatelmas: we added this behavior for consistency but I am thinking a case, I deleted a resources and now I want to purge cache immediately but this will fail. Just pointing out for ease of use

src/storage/cdn/cdn-cache-manager.test.ts
```diff
@@ -18,11 +18,14 @@ async function importCdnCacheManager(config: CdnConfig = {}) {
 }
 
 function createStorageMock(findObject = vi.fn().mockResolvedValue({ id: 'object-id' })) {
+  const storageAsSuperUser = vi.fn(() => ({
```
ferhatelmas, line 21: I think this is not needed anymore

src/storage/cdn/cdn-cache-manager.ts
ferhatelmas, line 83: One thing I wanted to point out in terms of gc pressure. 

There is nothing request specific here other than `storage` but we create a cdn manager anyway. I think it would be better to pass `storage` when needed. Also, we register cdn in routes where we don't need it. 

No need to make changes in this PR but we need to rework hooks soonish so this is just heads up if you already want to adapt cdn in that direction

PR #1135, opened 2026-06-04
fix: always validate signed urls scope

fenos:
What kind of change does this PR introduce?

  Bug fix (security) — scopes signed-URL tokens to a single action.                                                                       

  What is the current behavior?

  Signed download and signed upload URLs are minted with the same signing key and carry no claim distinguishing one from the other. Verification only checks the url and exp claims, so the tokens are interchangeable

  - A download (read) signed-URL token can be replayed against PUT /object/upload/sign/... to overwrite the object at that path.
  - An upload token can likewise be replayed against the download/render endpoints.

Because download URLs are frequently shared widely and signed with long expiries, this allows a read-only link to be escalated into a write.

What is the new behavior?
  
Each token now carries an explicit scope claim ('download' or 'upload'), and every consuming endpoint enforces it. The check is a whitelist — accept only the matching shape, reject everything else — applied symmetrically:

## Additional context

**Backward compatibility** — no breakage on deploy:
  - Pre-existing download URLs (no scope, no upsert) keep working — important since they can be long-lived.
  - Pre-existing upload URLs keep working via the upsert-claim fallback: only the upload-signing flow ever emits upsert (Boolean(...), always serialized), so its presence reliably identifies a legacy upload token.
  - Cross-use is now rejected: a download token can't upload, and an upload/upload-shaped token can't download.

### DIFF
--- a/src/http/routes/object/getSignedObject.ts
+++ b/src/http/routes/object/getSignedObject.ts
@@ -1,9 +1,7 @@
 import { FastifyInstance } from 'fastify'
 import { FromSchema } from 'json-schema-to-ts'
 import { getConfig } from '../../../config'
-import { SignedToken, verifyJWT } from '../../../internal/auth'
-import { getJwtSecret } from '../../../internal/database'
-import { ERRORS } from '../../../internal/errors'
+import { SIGNED_URL_SCOPE_DOWNLOAD } from '../../../internal/auth'
 import { ROUTE_OPERATIONS } from '../operations'
 
 const { storageS3Bucket } = getConfig()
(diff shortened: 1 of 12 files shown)


src/http/routes/render/renderSignedImage.ts
```diff
@@ -71,6 +71,17 @@ export default async function routes(fastify: FastifyInstance) {
 
       const path = `${request.params.bucketName}/${request.params['*']}`
 
+      // Only honor download tokens: either explicitly scoped for download, or a
+      // legacy download token (no scope and no `upsert` claim — only the upload
+      // flow ever emits `upsert`). This rejects upload tokens, legacy upload-shaped
+      // tokens, and any other scope, while keeping long-lived legacy download URLs.
+      const isDownloadToken =
```
itslenny: maybe we should put this isDownloadToken helper somewhere shared and reference it in both places since it's duplicated security logic in two places.

fenos: agree!

itslenny (approved):

src/http/routes/object/getSignedObject.ts
```diff
@@ -73,6 +73,10 @@ export default async function routes(fastify: FastifyInstance) {
       const { url, exp } = payload
       const path = `${request.params.bucketName}/${request.params['*']}`
 
+      if (!isDownloadScopedToken(payload)) {
```
ferhatelmas: nit: download validation could be centralized similar to upload (`verifyObjectSignature`)

ferhatelmas (approved):

ferhatelmas:
@GuptaManan100 [create unlogged table](https://github.com/supabase/storage/actions/runs/27000576182/job/79684529099?pr=1135) seems failing with the last image, FYI

GuptaManan100:
Yes we had a discussion about it in the team, and we decided that unlogged tables are going to be hard to support because they'll just vanish when a failover happens. Is this a feature that you rely on?

fenos:
We are using unlogged tables only for a specific script that reconciles S3 objects to what's in the customer database, it is a useful script and we might add this on a recovery path at some point. The of Unlogged tables is great here, because these tables are temporary and don't need the WAL

ferhatelmas:
> Is this a feature that you rely on?

Yes. For now, we're pinning to [old version](https://github.com/supabase/storage/pull/1135/changes/37e6d491f4e43d7cb513694f4e1c06f4a03d0052) and we will try to drop it. Surprising thing at first was that it was working yesterday and now started failing in unrelated changes.

PR #1145, opened 2026-06-10
fix: harden pg connection and admin timeouts

ferhatelmas:

Bug fix

Missed a pool in fix #1144 and registered knex error listener on checked-out client.

## What is the new behavior?

crash:
 * track checked-out pg client errors to prevent socket level failures from crashing
 * add error handler for single-tenant pgvector pool for idle connections

compat/hardening/improvements:
 * map connection timeout correctly
 * reject pre-aborted query signals before checkout
 * release tx client on abort instead of queueing rollback
 * normalize invalid/legacy pool mode on read
 * trigger replace/rebalance after admin tenant writes (old bug)
 * increase admin migration job cleanup limit to 2000
 * restore keep fleet migration on caller abort timeout
 * filter soft-deleted iceberg catalog during reconcile
 * reuse table capability probes for pgvector

## Additional context

Related to #1096
Related to #1144

### DIFF
--- a/src/http/routes/admin/migrations.ts
+++ b/src/http/routes/admin/migrations.ts
@@ -1,4 +1,8 @@
-import { MigrationAdminStorePg, multitenantPgExecutor } from '@internal/database'
+import {
+  MIGRATION_ADMIN_JOB_LIMIT,
+  MigrationAdminStorePg,
+  multitenantPgExecutor,
+} from '@internal/database'
 import {
   isDBMigrationName,
   resetMigrationsOnTenants,
@@ -70,7 +74,10 @@ export default async function routes(fastify: FastifyInstance) {
     if (!pgQueueEnable) {
       return reply.code(400).send({ message: 'Queue is not enabled' })
     }
-    const data = await migrationAdminStorePg.listActiveJobs(migrationQueueName, 2000)
+    const data = await migrationAdminStorePg.listActiveJobs(
+      migrationQueueName,
+      MIGRATION_ADMIN_JOB_LIMIT
+    )
 
     return reply.send(data)
   })
@@ -79,7 +86,10 @@ export default async function routes(fastify: FastifyInstance) {
     if (!pgQueueEnable) {
       return reply.code(400).send({ message: 'Queue is not enabled' })
     }
-    const data = await migrationAdminStorePg.completeActiveJobs(migrationQueueName)
+    const data = await migrationAdminStorePg.completeActiveJobs(
+      migrationQueueName,
+      MIGRATION_ADMIN_JOB_LIMIT
+    )
 
     return reply.send(data)
   })
--- a/src/http/routes/admin/migrations.ts
+++ b/src/http/routes/admin/migrations.ts
@@ -79,7 +79,7 @@ export default async function routes(fastify: FastifyInstance) {
     if (!pgQueueEnable) {
       return reply.code(400).send({ message: 'Queue is not enabled' })
     }
-    const data = await migrationAdminStorePg.completeActiveJobs(migrationQueueName, 2000)
+    const data = await migrationAdminStorePg.completeActiveJobs(migrationQueueName)
 
     return reply.send(data)
   })
--- a/src/http/routes/admin/tenants.ts
+++ b/src/http/routes/admin/tenants.ts
@@ -4,6 +4,7 @@ import {
   getTenantCapabilities,
   getTenantConfig,
   jwksManager,
+  MIGRATION_ADMIN_JOB_LIMIT,
   MigrationAdminStorePg,
   multitenantPgExecutor,
   onTenantConfigChange,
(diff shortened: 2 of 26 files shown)


src/http/routes/admin/migrations.ts
```diff
@@ -79,7 +79,7 @@ export default async function routes(fastify: FastifyInstance) {
     if (!pgQueueEnable) {
       return reply.code(400).send({ message: 'Queue is not enabled' })
     }
-    const data = await migrationAdminStorePg.completeActiveJobs(migrationQueueName, 2000)
+    const data = await migrationAdminStorePg.completeActiveJobs(migrationQueueName)
```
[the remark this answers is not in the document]
ferhatelmas: intentional

src/http/routes/admin/tenants.ts
```diff
@@ -244,7 +255,7 @@ function listTenantMigrationJobs(tenantId: string) {
 }
 
 function deleteTenantMigrationJobs(tenantId: string) {
-  return migrationAdminStorePg.deleteTenantJobs(tenantId, migrationQueueName, 100)
+  return migrationAdminStorePg.deleteTenantJobs(tenantId, migrationQueueName)
 }
```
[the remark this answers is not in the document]
ferhatelmas: intentional

TylerHillery (approved):

src/internal/database/migration-admin-store-pg.ts
```diff
@@ -74,22 +63,14 @@ export class MigrationAdminStorePg {
...
-          LIMIT $3
-        )
-        DELETE FROM ${this.jobTable} AS job
-        USING jobs_to_delete
-        WHERE job.id = jobs_to_delete.id
+        DELETE FROM ${this.jobTable}
+        WHERE data->'tenant'->>'ref' = $1
+          AND name = $2
       `,
-      values: [tenantId, queueName, limit],
```
fenos, line 92: Any reason to remove the limit here?

src/internal/database/tenant-store-pg.ts
```diff
@@ -218,49 +218,57 @@ export class TenantConfigStorePg {
...
-    const result = await this.query<TenantCursorRow>({
-      text: `
-        SELECT id, cursor_id
-        FROM tenants
-        WHERE cursor_id > $1
-          AND (
-            (
-              migrations_version != $2
-              AND migrations_status != ALL($3::text[])
+    const result = await this.queryWithoutInternalTimeout<TenantCursorRow>(
```
fenos: I'm wondering it's it's better to reuse the same query method and pass an optional overwrite time maybe to 0 if we want to avoid a query timeout. However, I still think that having a default timeout of 5s might be beneficial

src/internal/database/tenant-store-pg.ts
```diff
@@ -273,6 +281,13 @@ export class TenantConfigStorePg {
       signal: AbortSignal.timeout(multitenantDatabaseQueryTimeout),
     })
   }
+
+  private queryWithoutInternalTimeout<T extends QueryResultRow = QueryResultRow>(
+    statement: Parameters<PgExecutor['query']>[0],
+    signal?: AbortSignal
+  ) {
+    return signal ? this.db.query<T>(statement, { signal }) : this.db.query<T>(statement)
```
fenos: Why the function says without timeout but I don't see any timeout removal logic?

src/storage/protocols/vector/adapter/pgvector/index.ts
```diff
@@ -375,6 +385,14 @@ export class PgVectorStore implements VectorStore {
     return resolveRootPgExecutor(this.executor)
   }
 
+  private tableCapabilityCacheKey(db: PgTransactionalExecutor | PgTransaction): object {
```
fenos: Method duplicated, we also have the pure function tableCapabilityCacheKey

ferhatelmas: yes, dropped this in favor of the other

src/internal/database/migration-admin-store-pg.ts
ferhatelmas: I think limit is better but brought prior logic back since no pagination logic for cleanup, if you prefer it as well, can keep it

src/internal/database/tenant-store-pg.ts
ferhatelmas: it doesn't attach timeout, that's why, will improve its name

src/internal/database/tenant-store-pg.ts
ferhatelmas: This is about fleet migration. I think your suggestion makes sense, which addresses your other comment related to naming

src/internal/database/migration-admin-store-pg.ts
fenos: yes let's keep the limit, otherwise this might be not useful or too expensive

fenos (approved):

src/internal/database/migration-admin-store-pg.ts
ferhatelmas: ok, brought it back and increased hardcoded constant to 2000 to share between routes

PR #1154, opened 2026-06-15
fix: fold timeout into scope query

ferhatelmas:

Refactor for perf

Right now, statement timeout is sent separately which introduces extra round trip.

## What is the new behavior?

Delay until the first statement (set_config) and combine them together in the same query to reduce one round trip to db.

## Additional context

Additionally, make query text for scope module level, and cache headers/jwt payload for multiple queries in the same connection.

### DIFF
--- a/src/internal/database/pg-connection.ts
+++ b/src/internal/database/pg-connection.ts
@@ -475,7 +475,7 @@ export class PgTransaction implements PgExecutor {
     private readonly clientErrorTracker?: PgClientErrorTracker,
     options: PgTransactionOptions = {}
   ) {
-    this.statementTimeoutMs = options.statementTimeoutMs
+    this.statementTimeoutMs = normalizeStatementTimeoutMs(options.statementTimeoutMs)
   }
 
   isCompleted(): boolean {
--- a/src/internal/database/pg-connection.ts
+++ b/src/internal/database/pg-connection.ts
@@ -45,6 +45,24 @@ export interface PgQueryOptions {
 
 type PgQueryArgument = PgQueryOptions | unknown[]
 
+const scopeConfigNames = [
+  'role',
+  'request.jwt.claim.role',
+  'request.jwt',
+  'request.jwt.claim.sub',
+  'request.jwt.claims',
+  'request.headers',
+  'request.method',
+  'request.path',
+  'storage.operation',
+] as const
+
+const scopeConfigStatement = buildScopeConfigStatement(scopeConfigNames)
+const scopeConfigStatementWithTimeout = buildScopeConfigStatement([
+  'statement_timeout',
+  ...scopeConfigNames,
+])
+
 export interface PgExecutor {
   query<T extends QueryResultRow = QueryResultRow>(
     statement: string | PgStatement,
@@ -303,6 +321,18 @@ function pulsePgPoolQueue(pool: Pool): void {
   ;(pool as Pool & { _pulseQueue?: () => void })._pulseQueue?.()
 }
 
+function buildScopeConfigStatement(configNames: readonly string[]): string {
+  const configStatements = configNames.map((name, index) => {
+    return `set_config('${name}', $${index + 1}, true)`
+  })
+  configStatements.push(`set_config('storage.allow_delete_query', 'true', true)`)
+
+  return `
+        SELECT
+          ${configStatements.join(',\n          ')};
+      `
+}
+
 export class PgPoolManager extends PoolManager<PgPoolStrategy> {
   protected newPool(settings: PoolStrategySettings): PgPoolStrategy {
     return new PgPoolStrategy(settings)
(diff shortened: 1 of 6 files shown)


src/internal/database/pg-connection.ts
```diff
@@ -658,32 +715,22 @@ export class PgTenantConnection {
...
+      this.options.user.payload.sub || '',
+      this.userPayload,
+      this.headersPayload,
+      this.options.method || '',
+      this.options.path || '',
+      this.options.operation?.() || '',
+    ]
+
     await tnx.query({
-      text: `
```
fenos, line 679: We lost, readability of code with this approach. 
`setScope` is an important function that i'd prefer to keep explicit and not dynamic if possible. 

I'd prefer an approach where we are still explicit on the config we are setting, but we can potentially add more dynamically. 

It's just very hard to reason about now.

fenos:
It's an interesting approach, but it might beat us in case we never call `setScope` on a query, and expect the timeout to still work. 

Not sure if it's worth the performance gain here (1ms or less) vs a bit of a hidden timeout logic

src/internal/database/pg-connection.ts
```diff
@@ -426,6 +457,16 @@ export class PgTransaction implements PgExecutor {
     return this.completed
   }
 
+  setPendingStatementTimeout(timeoutMs: number | undefined): void {
```
fenos: Wondering, why this should use getter / setter. 
Isn't the timeout a fixed statement at the beginning of the transaction and never reset?

If that's the case we could use constructor argument to set the timeout instead?

ferhatelmas: in that case, it would set the timeout before transaction exists because first query is begin

ferhatelmas:
> it might beat us in case we never call setScope on a query

that's fine and covered by a [test](https://github.com/supabase/storage/pull/1154/changes#diff-d624316485835ecb052dee8d429d361739ba5ebd47916f9c4c27f295737dcf76R537)

src/internal/database/pg-connection.ts
ferhatelmas: improved readability a bit

src/internal/database/pg-connection.ts
```diff
@@ -658,36 +751,50 @@ export class PgTenantConnection {
...
-        this.options.method || '',
-        this.options.path || '',
-        this.options.operation?.() || '',
-      ],
+      text: statementTimeout ? scopeConfigStatementWithTimeout : scopeConfigStatement,
+      values: statementTimeout ? [...scopeValues, `${statementTimeout}ms`] : scopeValues,
     })
+
+    if (isMultigres && transaction) {
+      // Multigres requires SET LOCAL for statement_timeout, and role scope setup resets it.
```
ferhatelmas: @GuptaManan100 `set_config('statement_timeout', ..., true)` does not change effective transaction timeout. Is this worked on? Workaround is making the hot code complex, it would be desirable if we wouldn't need

GuptaManan100: It doesn't!? I'll file a bug and fix it on Multigres, we shouldn't need to change code here. I'll get this fixed and let you know when the fix is in place.

ferhatelmas: Nope, I have tests which work with pg/oriole but fails with Multigres. Looking forward to your fix 🙏🏻

GuptaManan100: Verified this is indeed a bug, and I've reproduced it locally. Working on a fix now.

GuptaManan100: Here is the fix that I've settled on! - https://github.com/multigres/multigres/pull/1238
We'll try and get this merged as soon as possible! Thank you once again for finding the bug ❤️

GuptaManan100: We have the fix merged! So the new images should have the fix in a while and we can retry!

ferhatelmas: Seems we can set, nice fix but we need to use `SHOW statement_timeout` to read, `current_setting('statement_timeout', true)` doesn't return the correct value. Is it expected?

GuptaManan100: Yes, that is another area we need to resolve, I'll get on it 👍

ferhatelmas: Sounds good, thanks 👍🏻

fenos (approved):

src/internal/database/pg-connection.ts
GuptaManan100: This has also been fixed with https://github.com/multigres/multigres/pull/1258!

ferhatelmas: it works https://github.com/supabase/storage/pull/1228

PR #1155, opened 2026-06-15
ci: add multigres to migration idempotency check

TylerHillery:

Now that multigres supports pg dump we can include it in the migrations idempotency check

Currently multigres step doesn't include this check because it did not support pg dump

## What is the new behavior?

Now we have multigres also do the migration idempotency check. I have extracted this step into it's own common action so we can reuse it across steps. This centralizes the logic in one spot.


ferhatelmas:
FYI, multigres image that supports pg_dump isn't ready yet

TylerHillery:
> FYI, multigres image that supports pg_dump isn't ready yet

Sounds good I will leave this in draft until it's ready to test

ferhatelmas (approved):

PR #1159, opened 2026-06-16
fix: cheaper counter metrics

ferhatelmas:

Refactor for performance

Sync allocation/hashing for counter metrics. 

## What is the new behavior?

Convert counters to observables tallied by in-memory integers.
Use batch observer to traverse related states once (for example res and req size)

## Additional context

Related to https://github.com/supabase/storage/pull/1157

### DIFF
--- a/src/internal/monitoring/metrics.ts
+++ b/src/internal/monitoring/metrics.ts
@@ -114,13 +114,6 @@ const httpRequestSizeBytes = registerMetric('http_request_size_bytes', 'counter'
     unit: 'bytes',
   })
 )
-httpRequestSizeBytes.addCallback((observer) => {
-  httpSizeMetricsState.forEach((state) => {
-    if (state.requestBytes > 0) {
-      observer.observe(state.requestBytes, state.attributes)
-    }
-  })
-})
 
 const httpResponseSizeBytes = registerMetric('http_response_size_bytes', 'counter', () =>
   meter.createObservableCounter('http_response_size_bytes', {
@@ -128,13 +121,20 @@ const httpResponseSizeBytes = registerMetric('http_response_size_bytes', 'counte
     unit: 'bytes',
   })
 )
-httpResponseSizeBytes.addCallback((observer) => {
-  httpSizeMetricsState.forEach((state) => {
-    if (state.responseBytes > 0) {
-      observer.observe(state.responseBytes, state.attributes)
-    }
-  })
-})
+
+meter.addBatchObservableCallback(
+  (observer) => {
+    httpSizeMetricsState.forEach((state) => {
+      if (state.requestBytes > 0) {
+        observer.observe(httpRequestSizeBytes, state.requestBytes, state.attributes)
+      }
+      if (state.responseBytes > 0) {
+        observer.observe(httpResponseSizeBytes, state.responseBytes, state.attributes)
+      }
+    })
+  },
+  [httpRequestSizeBytes, httpResponseSizeBytes]
+)
 
 /** Records request/response bytes by bumping in-process tallies with one state lookup. */
 export function recordHttpSizes(
@@ -192,26 +192,26 @@ const fileUploadStarted = registerMetric('upload_started', 'counter', () =>
     description: 'Total uploads started',
   })
 )
-fileUploadStarted.addCallback((observer) => {
-  uploadMetricsState.forEach((state) => {
-    if (state.started > 0) {
-      observer.observe(state.started, state.attributes)
-    }
-  })
-})
 
 const fileUploadedSuccess = registerMetric('upload_success', 'counter', () =>
   meter.createObservableCounter('upload_success', {
     description: 'Total successful uploads',
   })
 )
-fileUploadedSuccess.addCallback((observer) => {
-  uploadMetricsState.forEach((state) => {
-    if (state.success > 0) {
-      observer.observe(state.success, state.attributes)
-    }
-  })
-})
+
+meter.addBatchObservableCallback(
+  (observer) => {
+    uploadMetricsState.forEach((state) => {
+      if (state.started > 0) {
+        observer.observe(fileUploadStarted, state.started, state.attributes)
+      }
+      if (state.success > 0) {
+        observer.observe(fileUploadedSuccess, state.success, state.attributes)
+      }
+    })
+  },
+  [fileUploadStarted, fileUploadedSuccess]
+)
 
 /** Records an upload start by bumping an in-process tally. */
 export function recordUploadStarted(uploadType: string): void {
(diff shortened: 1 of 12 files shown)


src/internal/monitoring/metrics.ts
```diff
@@ -57,7 +57,47 @@ export function registerMetric<T>(name: string, type: MetricType, factory: () =>
...
+  status_code: string
+}
+
+type HttpSizeMetricsState = {
+  requestBytes: number
+  responseBytes: number
+  attributes: Attributes
+}
+
+const httpSizeMetricsState = new Map<string, HttpSizeMetricsState>()
```
mcollina: This should be an LRU to prevent memory leaks and DoS attacks

src/internal/monitoring/metrics.ts
```diff
@@ -57,7 +57,47 @@ export function registerMetric<T>(name: string, type: MetricType, factory: () =>
...
+
+function getHttpSizeMetricsState(attributes: HttpMetricAttributes): HttpSizeMetricsState {
+  const key = `${attributes.method}\x00${attributes.operation}\x00${attributes.status_code}`
+  let state = httpSizeMetricsState.get(key)
+  if (!state) {
+    state = {
+      requestBytes: 0,
+      responseBytes: 0,
```
mcollina: this would mean that the metric is monotonically growing. Over time, you might problems because those numbers will become large

src/internal/monitoring/metrics.ts
ferhatelmas: States are bounded at the moment, it wouldn't be an issue in the current usage. 

I can add a cap for general just in case

src/internal/monitoring/metrics.ts
ferhatelmas: Yes, I am aware of it and taking for good enough. Do you have a suggestion how to make it more resilient?

fenos (approved):
nice implementation!

src/internal/monitoring/metrics.ts
fenos: Even the underline otel metric would keep this number ever growing. Resetting it, would results in incorrect metrics. 

Not sure how we can improve this, maybe we should consider using a `BigInt` instead of `number`?

src/internal/monitoring/metrics.ts
```diff
@@ -57,7 +58,63 @@ export function registerMetric<T>(name: string, type: MetricType, factory: () =>
...
+  createState: (attributes: HttpMetricAttributes): HttpSizeMetricsState => ({
+    requestBytes: 0,
+    responseBytes: 0,
+    attributes: {
+      method: attributes.method,
+      operation: attributes.operation,
+      status_code: attributes.status_code,
+    },
+  }),
+  observe: (observer, counters, state) => {
```
fenos: Is this actually needed? can we not iterate the counters internally and set the observer the same way by default, maybe we can keep it as an optional overwrite?

src/internal/monitoring/metrics.ts
```diff
@@ -68,34 +125,80 @@ export const httpRequestDuration = registerMetric(
...
+      observer.observe(counters.started, state.started, state.attributes)
+    }
+    if (state.success > 0) {
+      observer.observe(counters.success, state.success, state.attributes)
+    }
+  },
+})
+
+/** Records an upload start by bumping an in-process tally. */
+export function recordUploadStarted(uploadType: string): void {
```
fenos, line 265: Wondering if we should remove these helper function and achieve some type of api that can be cleanly called from the outside, for ex:

```
 uploadMetrics.state(uploadType).incGroupCounter('started')

File: src/internal/monitoring/metrics.ts (hunk 1 above)
Comment (ferhatelmas) [reply src/internal/monitoring/metrics.ts]: Added LRU cache with safe max cap, PTAL

File: src/internal/monitoring/metrics.ts (hunk 2 above)
Comment (ferhatelmas) [reply src/internal/monitoring/metrics.ts]: Added safe add to reset, this should handle the concern. I wouldn't go for big int for overhead

File: src/internal/monitoring/metrics.ts (hunk 3 above)
Comment (ferhatelmas) [reply src/internal/monitoring/metrics.ts]: For http/upload, yes possible but cache metrics don't map 1-1 cleanly

File: src/internal/monitoring/metrics.ts (hunk 1 above)
Comment (fenos) [reply src/internal/monitoring/metrics.ts]: Mmmh @ferhatelmas LRU cache would make the metrics look wrong if they get evicted since it will start from 0 again

File: src/internal/monitoring/metrics.ts (hunk 2 above)
Comment (fenos) [reply src/internal/monitoring/metrics.ts]: @ferhatelmas not sure i understood what you mean by "safe to add reset" can you explain please

File: src/internal/monitoring/metrics.ts (hunk 3 above)
Comment (ferhatelmas) [reply src/internal/monitoring/metrics.ts]: made it default and pass if only custom

File: src/internal/monitoring/metrics.ts (hunk 4 above)
Comment (ferhatelmas) [reply src/internal/monitoring/metrics.ts:265]: provided a helper for upload/http

File: src/internal/monitoring/metrics.ts (hunk 2 above)
Comment (ferhatelmas) [reply src/internal/monitoring/metrics.ts]: it will reset when instance is restarting, it will be rare and we use rate in visualization so made incrementing safe and it resets automatically instead of overflow and precision loss

File: src/internal/monitoring/metrics.ts (hunk 1 above)
Comment (ferhatelmas) [reply src/internal/monitoring/metrics.ts]: yes, but it won't hit the max states according to our usage. It's just a safety net

Comment (ferhatelmas) [reply src/internal/monitoring/metrics.ts]: LRU cache seemed to be too expensive, tracked a number and fall backed to overflow state.

File: src/internal/monitoring/metrics.ts (hunk 2 above)
Comment (ferhatelmas) [reply src/internal/monitoring/metrics.ts]: I switched it to not increasing in the final version

File: src/internal/monitoring/metrics.ts [hunk 5]
```diff
@@ -106,72 +278,71 @@ export const fileUploadedSuccess = registerMetric('upload_success', 'counter', (
...
-  getCacheMetricsState(cache).requests[outcome]++
+  const state = cacheMetrics.state(cache).requests[outcome]
+  state.count = safeAddCounter(state.count, 1)
 }
 
 /** Records a single capacity/ttl cache eviction by bumping an in-process tally. */
 export function recordCacheEviction(cache: CacheName): void {
-  getCacheMetricsState(cache).evictions++
+  const state = cacheMetrics.state(cache)
+  state.evictions = safeAddCounter(state.evictions, 1)
```
fenos: This api still seems a bit strange. Not intuitive. 

Can we just call a method that does this internally?

ferhatelmas: Cache had a different shape and made difficult but did a bit standardization and auto generated methods, should be cleaner now

fenos (approved):

PR #1160, opened 2026-06-18
fix: limit for delete objects and sign urls

ferhatelmas:

Bug fix

Delete objects/sign urls don't have a proper max limit. They are capped according to url length for the old postgrest implementation but it makes unnecessary allocation and not user friendly. 

## What is the new behavior?

Set a clear limit of 1000.

## Additional context

Internal batching is 500 because it fanouts with `.info` objects.

### DIFF
--- a/src/http/routes/s3/router.test.ts
+++ b/src/http/routes/s3/router.test.ts
@@ -1,3 +1,4 @@
+import { MAX_OBJECTS_PER_REQUEST } from '@storage/limits'
 import { vi } from 'vitest'
 import { S3ProtocolHandler } from '../../../storage/protocols/s3/s3-handler'
 import { Uploader } from '../../../storage/uploader'
@@ -242,6 +243,45 @@ describe('CompleteMultipartUpload route mapping', () => {
 })
 
 describe('DeleteObject route mapping', () => {
+  it('rejects DeleteObjects payloads over the object request cap in router validation', async () => {
+    const { default: DeleteObject } = await import('./commands/delete-object')
+    const router = new Router()
+
+    DeleteObject(router as unknown as S3Router)
+
+    const route = router
+      .routes()
+      .get('/:Bucket')
+      ?.find(
+        (candidate) =>
+          candidate.method === 'post' &&
+          candidate.querystringMatches.some((match) => match.key === 'delete')
+      )
+
+    expect(route).toBeDefined()
+
+    const validate = route!.compiledSchema()
+    const data = {
+      Params: { Bucket: 'bucket' },
+      Querystring: { delete: '' },
+      Body: {
+        Delete: {
+          Object: [...Array(MAX_OBJECTS_PER_REQUEST + 1).keys()].map((i) => ({
+            Key: `object-${i}`,
+          })),
+        },
+      },
+    }
+
+    expect(validate(data)).toBe(false)
+    expect(validate.errors).toEqual([
+      expect.objectContaining({
+        instancePath: '/Body/Delete/Object',
+        keyword: 'maxItems',
+      }),
+    ])
+  })
+
   it('returns 204 from iceberg single-object deletes', async () => {
     const previousIcebergDeleteEnabled = process.env.ICEBERG_S3_DELETE_ENABLED
     process.env.ICEBERG_S3_DELETE_ENABLED = 'true'
--- a/src/http/routes/s3/router.test.ts
+++ b/src/http/routes/s3/router.test.ts
@@ -243,13 +243,13 @@ describe('CompleteMultipartUpload route mapping', () => {
 })
 
 describe('DeleteObject route mapping', () => {
-  const getDeleteObjectsRoute = async () => {
+  it('accepts DeleteObjects payloads at the object request cap in router validation', async () => {
     const { default: DeleteObject } = await import('./commands/delete-object')
     const router = new Router()
 
     DeleteObject(router as unknown as S3Router)
 
-    return router
+    const route = router
       .routes()
       .get('/:Bucket')
       ?.find(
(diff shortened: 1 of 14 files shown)


src/http/routes/s3/router.test.ts
```diff
@@ -242,6 +243,45 @@ describe('CompleteMultipartUpload route mapping', () => {
 })
 
 describe('DeleteObject route mapping', () => {
+  it('rejects DeleteObjects payloads over the object request cap in router validation', async () => {
```
fenos, line 281: Can we add a test that also assert that 1000 objects are going to be successfully validated

src/storage/backend/s3/adapter.ts
```diff
@@ -429,17 +430,19 @@ export class S3Backend implements StorageBackendAdapter {
    */
   async deleteObjects(bucket: string, prefixes: string[]): Promise<void> {
     try {
-      const s3Prefixes = prefixes.map((ele) => {
-        return { Key: ele }
-      })
+      for (let i = 0; i < prefixes.length; i += MAX_KEYS_PER_S3_DELETE) {
```
fenos, line 435: Wondering if we can move the DelectObjectsCommand outside the loop? since now we are simply constructing the object arrays?

src/test/object.test.ts
```diff
@@ -1810,23 +1830,62 @@ describe('testing delete object', () => {
  * */
 describe('testing deleting multiple objects', () => {
   test('check if RLS policies are respected: authenticated user is able to delete authenticated resource', async () => {
```
fenos: Description of the test doesn't reflect what we are testing

src/storage/backend/s3/adapter.ts
ferhatelmas: do you want to do in parallel?

src/http/routes/s3/router.test.ts
ferhatelmas: added

src/test/object.test.ts
```diff
@@ -2452,48 +2512,91 @@ describe('testing uploading with generated signed upload URL', () => {
...
       },
     })
-    expect(response.statusCode).toBe(200)
-    const result = JSON.parse(response.body)
-    expect(result[0].error).toBe('Either the object does not exist or you do not have access to it')
+    expect(response.statusCode).toBe(400)
   })
 
-  test('user is not able to generate signedURLs without Auth header', async () => {
+  test('rejects signed URL requests over the object request cap', async () => {
```
fenos, line 2602: Mmmh we shouldn't remove the previous test, maybe adding another one is better

src/test/object.test.ts
ferhatelmas: updated

src/test/object.test.ts
ferhatelmas: added a new and kept the old

src/storage/backend/s3/adapter.ts
ferhatelmas: made it parallel

fenos:
@copilot review again the PR

src/test/object.test.ts
```diff
@@ -1809,24 +1830,63 @@ describe('testing delete object', () => {
  * DELETE /objects
  * */
 describe('testing deleting multiple objects', () => {
-  test('check if RLS policies are respected: authenticated user is able to delete authenticated resource', async () => {
```
fenos, line 1812: I think we have overwritten the previous test, we need to make sure we keep both

src/test/object.test.ts
```diff
@@ -2452,48 +2512,91 @@ describe('testing uploading with generated signed upload URL', () => {
  */
 describe('testing generating signed URLs', () => {
   test('check if RLS policies are respected: authenticated user is able to sign URLs for an authenticated resource', async () => {
```
fenos, line 2514: description of the test doesn't reflect what we are testing

src/test/object.test.ts
```diff
@@ -2452,48 +2512,91 @@ describe('testing uploading with generated signed upload URL', () => {
...
+    }
+  })
+
+  test('check if RLS policies are respected: anon user is not able to generate signedURLs for authenticated resource', async () => {
     const response = await appInstance.inject({
       method: 'POST',
       url: '/object/sign/bucket2',
       headers: {
-        authorization: `Bearer ${process.env.AUTHENTICATED_KEY}`,
+        authorization: `Bearer ${anonKey}`,
```
fenos, line 2578: is changing the key correct here? before we were using `authenticated`, now `anon`

ferhatelmas: tests are shuffled from iterations

src/test/object.test.ts
ferhatelmas: tried to minimize diff, should be clearer to read now

src/test/object.test.ts
ferhatelmas: it's added below

src/storage/object.ts
```diff
@@ -779,20 +766,13 @@ export class ObjectStorage {
...
-    for (let i = 0; i < paths.length; ) {
-      const pathsSubset = []
-      let urlParamLength = 0
-
-      for (; i < paths.length && urlParamLength < requestUrlLengthLimit; i++) {
-        const path = paths[i]
-        pathsSubset.push(path)
-        urlParamLength += encodeURIComponent(path).length + 9 // length of '%22%2C%22'
-      }
+    for (let i = 0; i < paths.length; i += MAX_OBJECTS_PER_LOOKUP_BATCH) {
```
fenos: I'm wondering if we should throw an error if we call this function with more than 1k items instead of silently ignore them, what do you think?

fenos (approved):

src/storage/object.ts
ferhatelmas: we're not ignoring but handling all of them.

Since it's already handled in validation layer, I wouldn't bring another check here.

ferhatelmas: I made a shortcut if less than batch size not to allocate array again

PR #1164, opened 2026-06-19
fix: inefficient traversal and intermediate allocations

ferhatelmas:

Refactoring for performance

We have some inefficient iteration (Object.keys) and intermediate allocations (filter then map) in hot paths.

## What is the new behavior?

Use iterative for loops to get rid of them.


fenos (approved):
Nice improvements here!

PR #1168, opened 2026-06-22
perf(s3): improve router performance

ferhatelmas:

Refactor for perf

S3 router is less efficient than it could be.

## What is the new behavior?

* precompile route matchers at registration
* ajv validator is moved out hot path
* operation is precomputed and cached on route
* prevent wrapper object allocation in match loop
* partition routes between iceberg and standard


fenos (approved):
Very nice optimisation!

PR #1169, opened 2026-06-22
perf: use fast-decode for uri component decoding

ferhatelmas:
Part of the performance work adopting patterns from fastify.

Swaps `decodeURIComponent` for [`fast-decode-uri-component`](https://github.com/fastify/fast-decode-uri-component) (the decoder fastify uses internally) at the URI-component decode sites. It returns the input unchanged when there is nothing to decode (no `%` → no allocation) and returns `null` instead of throwing on malformed input.

Changes:
- `isIpAddress` (database SSL): decode with fast-decode, falling back to the raw value on malformed input instead of throwing.
- pprof multipart filename parsing: use the `null` return to fall through to the quoted form, replacing the `try/catch`.
- Add `fast-decode-uri-component` as a direct dependency (previously transitive via fastify) plus a small ambient type declaration.

`decodeURI` in `object.ts` is intentionally left alone — its semantics differ from `decodeURIComponent` and aren't interchangeable.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,6 +49,7 @@
         "aws-sigv4-sign": "^1.2.1",
         "conventional-changelog-conventionalcommits": "^5.0.0",
         "dotenv": "^16.0.0",
+        "fast-decode-uri-component": "^1.0.1",
         "fastify": "^5.8.5",
         "fastify-plugin": "^5.1.0",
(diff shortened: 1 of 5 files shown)


ferhatelmas:
closing this since main hot path is already included in https://github.com/supabase/storage/pull/1174 and other uses are cold so not needed

PR #1171, opened 2026-06-22
perf(logger): serialize request/reply logs with fast-json-stringify

ferhatelmas:
Defines JSON schemas for the structured request and reply log payloads and compiles them with `fast-json-stringify`. The compiled serializers are wired into pino as per-key stringifiers for `req`/`res`, so the hot logging path (emitted on every request) skips pino's generic stringify for these objects.

The existing `req`/`res` serializers still run first to normalize the values; the compiled stringifier then produces the embedded JSON. Child loggers inherit the stringifiers through the prototype chain, so the whole logger tree benefits without re-registration. Log output is unchanged — `req`/`res` remain nested JSON objects.

`fast-json-stringify` was already present transitively (via fastify); it's now declared as a direct dependency since we import it.

### DIFF
--- a/src/internal/monitoring/logger.ts
+++ b/src/internal/monitoring/logger.ts
@@ -1,5 +1,6 @@
 import { resolve } from 'node:path'
 import { normalizeRawError } from '@internal/errors'
+import fastJson from 'fast-json-stringify'
 import fastQuerystring from 'fast-querystring'
 import type { FastifyBaseLogger, FastifyReply, FastifyRequest } from 'fastify'
 import pino, { Logger } from 'pino'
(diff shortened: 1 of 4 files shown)


src/internal/monitoring/logger.ts
```diff
@@ -34,6 +35,39 @@ export interface SerializedReplyLog extends SerializedReplyLogShape {
...
+} as const
+
+// JSON schemas for the structured payloads that appear on every request log.
+// They are compiled into fast-json-stringify serializers and wired into pino as
+// per-key stringifiers (see registerLogStringifiers), so the hot path avoids
+// pino's generic stringify for these objects.
+export const serializeRequestLogToJson = fastJson({
+  type: 'object',
+  properties: {
+    region: { type: 'string' },
```
fenos, line 50: Missing operation

fenos (approved):

src/internal/monitoring/logger.ts
ferhatelmas: this is only for res/req fields, rest might come according to the benefits of this

PR #1201, opened 2026-07-03
fix: improve error message for referenced resource (FK violation)

itslenny:

Bug fix

When a bucket has foreign key references that prevent it from being deleted the error message is vague and potentially misleading.

This most commonly happens because of orphan multipart uploads (storage.s3_multipart_uploads) that still reference the bucket. The old error message provides no obvious information to the end user.

**Current error message:**
```json
{
	"statusCode": "404",
	"error": "InvalidRequest",
	"message": "The related resource does not exist"
}
```

## What is the new behavior?

**New error message:**
```json
{
	"statusCode": "409",
	"error": "ResourceReferenced",
	"message": "Key (id)=(my-test-bucket) is still referenced from table \"s3_multipart_uploads\"."
}
```

## Additional context

The PostgreSQL error code `23503` is returned for all foreign key constraint violations. This change differentiates between missing related resources and resources that cannot be removed because they are still referenced by other resources


fenos (approved):

src/storage/database/errors.ts
```diff
@@ -25,6 +25,11 @@ export class DBError extends StorageBackendError implements RenderableError {
       case '23505':
         return ERRORS.ResourceAlreadyExists(pgError).withMetadata(pgErrorMetadata(pgError, context))
       case '23503':
+        if (pgError.detail?.includes('is still referenced from table')) {
+          return ERRORS.ResourceReferenced(pgError.detail, pgError).withMetadata(
```
fenos: We just need to make sure we don't return any SQL resource back to the client

ferhatelmas (approved):

fenos (approved):

PR #1207, opened 2026-07-06
fix: include resource type in IcebergMaximumResourceLimit error message

itslenny:

Bug fix

The IcebergMaximumResourceLimit error does not indicate which resource is causing the restriction.

This error message could indicate maximum number of catalogs, buckets, or tables and the end user has no way of knowing which resource is causing the restriction: `The maximum number of this resource 10 is reached`

## What is the new behavior?

Include the resource type in the error message text so the error message is more actionable.

### DIFF
--- a/src/storage/protocols/iceberg/catalog/tenant-catalog.ts
+++ b/src/storage/protocols/iceberg/catalog/tenant-catalog.ts
@@ -155,7 +155,7 @@ export class TenantAwareRestCatalog extends RestCatalogClient {
       })
 
       if (catalogCount >= this.options.limits.maxCatalogsCount) {
-        throw ERRORS.IcebergMaximumResourceLimit(this.options.limits.maxCatalogsCount)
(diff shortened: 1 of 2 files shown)


src/storage/protocols/iceberg/catalog/tenant-catalog.ts
```diff
@@ -155,7 +155,7 @@ export class TenantAwareRestCatalog extends RestCatalogClient {
       })
 
       if (catalogCount >= this.options.limits.maxCatalogsCount) {
-        throw ERRORS.IcebergMaximumResourceLimit(this.options.limits.maxCatalogsCount)
+        throw ERRORS.IcebergMaximumResourceLimit('catalogs', this.options.limits.maxCatalogsCount)
```
ferhatelmas: nit: I would prefer single

ferhatelmas (approved):

PR #1208, opened 2026-07-06
fix: update acceptance tests for staging

itslenny:

Bug fix

Acceptance tests cannot run against staging environment

## What is the new behavior?

Adjust acceptance tests so they can be run against staging

Add CDN Edge tests to test Smart CDN and cache purge behavior end to end


acceptance/specs/cdn.test.ts
```diff
@@ -0,0 +1,1085 @@
...
+      client = createRestClient()
+      bucketName = uniqueBucketName('cdn-purge')
+      await createRestBucket(bucketName, { isPublic: true })
+    })
+
+    afterAll(async () => {
+      await cleanupRestResources(bucketName, [], client)
+    })
+
+    it('purges an object cache entry', async () => {
```
fenos, line 182: i think we can use `it.skipIf(!isCdnEdge)` helper so that we can remove the ifs `isCdnEdge`

acceptance/specs/tus.test.ts
```diff
@@ -543,7 +543,13 @@ async function createTusUpload({
       method: 'OPTIONS',
     })
     expect([200, 204]).toContain(options.status)
-    expect(options.headers.get('tus-version')).toContain(tusVersion)
+    if (options.headers.has('cf-ray')) {
+      // api gateway should strip tus header
+      expect(options.headers.has('tus-version')).toBe(false)
```
fenos, line 548: it does strip it but it shouldn't 😄

itslenny: Updated the comment, created a ticket to fix this behavior at the gateway in the future.

acceptance/specs/cdn.test.ts
itslenny: I'd have to add separate tests for that to work. These tests hit the purge endpoints in both local and remote mode. Only in remote mode can we test if the purge actually worked. So, rather than having fully separate remote/local tests I am just conditionally doing the check on remote only before/after the purge.

acceptance/README.md
ferhatelmas:
```suggestion
STORAGE_BACKEND=file ACCEPTANCE_ENABLE_RENDER=true npm run acceptance -- --profile full acceptance/specs/render.test.ts
```

acceptance/specs/cdn.test.ts
```diff
@@ -0,0 +1,1085 @@
...
+
+          await client.request('GET', purgedRoute, {
+            expectedCacheStatus: 'MISS',
+            expectedStatus: 200,
+          })
+          await client.request('GET', purgedRoute, {
+            expectedCacheStatus: 'HIT',
+            expectedStatus: 200,
+            retries: CACHE_RETRIES,
+          })
```
ferhatelmas: repeated many times, this could be be a helper

acceptance/specs/cdn.test.ts
```diff
@@ -0,0 +1,1085 @@
...
+              },
+              expectedStatus: 200,
+              token: requireServiceKey(config),
+            })
+
+            // Signed requests return HIT for 60s or until the token expires
+            await delay(SIGNED_EXPIRES_IN_S * 1000)
+
+            const expectedError = isSigned
+              ? { statusCode: '400', error: 'InvalidJWT' }
```
ferhatelmas, line 780: I think we should test not expiring token to test real cache invalidation

acceptance/specs/cdn.test.ts
```diff
@@ -0,0 +1,1085 @@
...
+                expectedStatus: 200,
+                token: requireServiceKey(),
+              }
+            )
+
+            // Signed requests return HIT for 60s or until the token expires
+            await delay(SIGNED_EXPIRES_IN_S * 1000)
+
+            const expectedError = isSigned
+              ? { statusCode: '400', error: 'InvalidJWT' }
```
ferhatelmas, line 913: similar non-expiring token should be used here as well

acceptance/specs/cdn.test.ts
```diff
@@ -0,0 +1,1085 @@
...
+function getToken(route: string): string {
+  return new URL(route, 'http://abc').searchParams.get('token') || ''
+}
+
+describeAcceptance(
+  'CDN cache purge operations',
+  {
+    destructive: true,
+    profiles: ['full'],
+    requires: ['cdn'],
```
ferhatelmas, line 164: hits to render but doesn't require it

acceptance/specs/cdn.test.ts
```diff
@@ -0,0 +1,1085 @@
...
+                )
+                return { file, route, token }
+              })
+            )
+
+            await pauseForWebhookIfNeeded(accessMethod)
+
+            // check invalid token / access denied
+            if (isSigned) {
+              // const { route } = items[0]
```
ferhatelmas:
```suggestion
```

acceptance/specs/cdn.test.ts
itslenny: I did it this way so the test wouldn't take 60s+ but I guess for an acceptance test it's fine and it has no impact on the local (CI) runs. Updated.

ferhatelmas: When move/delete are done, cache invalidate and we should check miss before waiting for full expiration to see invalidation is working, right?

itslenny: No, that is not how it currently works for signed URLs. It takes 60s for a cache miss.

The reason is because on every signed URL request we read KV and check for an etag. If we find one we return the object from cache.

When the object is deleted we send the webhook which removes the KV (does not actually purge cache currently). The KV returns a stale value (with etag) for ~60s. After 60s we get a KV miss which results in a call to origin and a not found response.

The upcoming Smart CDN fixes will address this, but these acceptance tests are for the current behavior only.

itslenny: The updated test checks 2 things... one is the expired token (as before) and (per your recommendation) I added the test that waits the full 60s for the KV to invalidate so we can see an actual cache miss

acceptance/scripts/reset-project.ts
```diff
@@ -0,0 +1,276 @@
+/**
+ * Wipes every storage/analytics/vector bucket in the target project so acceptance tests
```
ferhatelmas, line 2: `ACCEPTANCE_ALLOW_DESTRUCTIVE`

Shall we check this env in addition to `yes`?

ferhatelmas (approved):

PR #1210, opened 2026-07-06
fix: simplify operation shape

ferhatelmas:

Refactor

Operation is wrapped inside into an object with type.

## What is the new behavior?

Unnecessary, make it top level.

## Additional context

Helps cleaning up compilation in S3 router.

### DIFF
--- a/src/http/routes/s3/router.test.ts
+++ b/src/http/routes/s3/router.test.ts
@@ -594,26 +594,6 @@ describe('S3 router registration precomputation', () => {
 
     expect(route!.operation).toBe('storage.s3.iceberg.object.get')
   })
-
-  it('stores a reusable operation object for request metadata', () => {
-    const router = new Router()
-
-    router.get(
-      '/:Bucket/*',
-      {
-        schema: {},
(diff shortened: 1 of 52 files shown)


src/http/routes/s3/router.test.ts
```diff
@@ -594,26 +594,6 @@ describe('S3 router registration precomputation', () => {
 
     expect(route!.operation).toBe('storage.s3.iceberg.object.get')
   })
-
-  it('stores a reusable operation object for request metadata', () => {
```
fenos, line 598: Why was this test deleted?

ferhatelmas: [operationConfig](https://github.com/supabase/storage/pull/1210/changes#diff-cb31a0761b4a43f47fcca47e84c432201b6779934b6ea5d7811536a299c569d8L128) is gone, it was for temporary compiled one until this clean up is done, `operation` is already covered by the above test

fenos (approved):

fenos (approved):

PR #1211, opened 2026-07-07
fix(S3): enforce presigned POST policy conditions

TylerHillery:

Bug fix to enforce policy conditions in S3 presigned POST url.

Currently only verifies the HMAC over the Policy blob 

## What is the new behavior?

Properly validates the conditions

### DIFF
--- a/src/http/plugins/signature-v4.ts
+++ b/src/http/plugins/signature-v4.ts
@@ -41,6 +41,7 @@ declare module 'fastify' {
     streamingSignatureV4?: ChunkSignatureV4Parser
     multiPartFileStream?: MultipartFile
     bodySha256: string
+    s3PostPolicyContentLengthMax?: number
   }
 }
 
@@ -160,6 +161,13 @@ async function authorizeRequestSignV4(
     )
   }
 
+  if (clientSignature.policy) {
+    request.s3PostPolicyContentLengthMax = signatureV4.validatePostPolicyConditions(
+      clientSignature,
+      { bucket: (request.params as Record<string, string | undefined>)?.Bucket }
+    )
+  }
+
   const wasBodyHashed = allowBodyHash && byteHasherStream && byteHasherStream.writableEnded
 
   const returnStream = wasBodyHashed
@@ -203,7 +211,7 @@ async function authorizeRequestSignV4(
   return returnStream
 }
 
-async function extractSignature(req: AWSRequest) {
+async function extractSignature(req: AWSRequest): Promise<ClientSignature> {
   if (typeof req.headers.authorization === 'string') {
     return SignatureV4.parseAuthorizationHeader(req.headers)
   }
--- a/src/http/plugins/signature-v4.ts
+++ b/src/http/plugins/signature-v4.ts
@@ -5,7 +5,12 @@ import { getJwtSecret, getTenantConfig, s3CredentialsManager } from '@internal/d
 import { ERRORS } from '@internal/errors'
 import { RequestByteCounterStream } from '@internal/streams'
 import { HashSpillWritable } from '@internal/streams/hash-stream'
-import { ClientSignature, SignatureV4, SignatureV4Service } from '@storage/protocols/s3'
+import {
+  assertPolicyConditionsSatisfied,
+  ClientSignature,
+  SignatureV4,
+  SignatureV4Service,
+} from '@storage/protocols/s3'
 import { ByteLimitTransformStream } from '@storage/protocols/s3/byte-limit-stream'
 import {
   ChunkSignatureV4Parser,
@@ -41,7 +46,6 @@ declare module 'fastify' {
     streamingSignatureV4?: ChunkSignatureV4Parser
     multiPartFileStream?: MultipartFile
     bodySha256: string
-    s3PostPolicyContentLengthMax?: number
   }
 }
 
@@ -162,10 +166,9 @@ async function authorizeRequestSignV4(
   }
 
   if (clientSignature.policy) {
-    request.s3PostPolicyContentLengthMax = signatureV4.validatePostPolicyConditions(
-      clientSignature,
-      { bucket: (request.params as Record<string, string | undefined>)?.Bucket }
-    )
+    assertPolicyConditionsSatisfied(clientSignature.policy.value, clientSignature.policy.fields, {
+      bucket: (request.params as Record<string, string | undefined>)?.Bucket,
+    })
   }
 
   const wasBodyHashed = allowBodyHash && byteHasherStream && byteHasherStream.writableEnded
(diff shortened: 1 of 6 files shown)


src/storage/protocols/s3/signature-v4.ts
```diff
@@ -301,6 +345,150 @@ export class SignatureV4 {
...
+   * The `bucket` value is taken from the request target (the URL), since AWS
+   * POST uploads carry the bucket in the path rather than as a form field.
+   */
+  validatePostPolicyConditions(
+    clientSignature: ClientSignature,
+    opts: { bucket?: string }
+  ): number | undefined {
+    const policy = clientSignature.policy
+    if (!policy) {
+      return undefined
```
fenos: should we error here? otherwise a clever user can simply omit the policies?

src/storage/protocols/s3/signature-v4.ts
```diff
@@ -301,6 +345,150 @@ export class SignatureV4 {
...
+        coveredFields.add(result.field)
+      }
+      if (result.max !== undefined) {
+        // Fold to the most restrictive upper bound across every content-length-range.
+        max = max === undefined ? result.max : Math.min(max, result.max)
+      }
+    }
+
+    // Every submitted field must be constrained by a condition
+    for (const field of Object.keys(fields)) {
```
fenos: Can we remove the above allocation and double loop by validating inline?

src/storage/protocols/s3/signature-v4.ts
```diff
@@ -301,6 +345,150 @@ export class SignatureV4 {
...
+  }
+
+  protected evaluatePolicyCondition(
+    condition: PolicyCondition,
+    fields: Record<string, string>
+  ): { field?: string; max?: number } {
+    const normalized = this.normalizePolicyCondition(condition)
+
+    if (normalized.operator === 'content-length-range') {
+      return { max: normalized.max }
```
fenos: this doesn't seem right, the `content-length-range` should be a range of min,max

src/storage/protocols/s3/signature-v4.ts
```diff
@@ -230,9 +239,22 @@ export class SignatureV4 {
 
     const xPolicy: Policy = JSON.parse(Buffer.from(policy, 'base64').toString('utf-8'))
 
-    if (xPolicy.expiration) {
-      this.checkExpiration(longDate, xPolicy.expiration)
-    }
+    // A POST policy must declare an expiration; AWS treats a missing one as
+    // invalid. Without this check a signed policy with no expiration would never
+    // expire and could be replayed forever.
+    this.checkPolicyExpiration(xPolicy.expiration)
```
fenos: wondering if we should extract policy validation and logic into a `policy.ts` file inside the `protocols/s3` to keep it tidier

src/http/plugins/signature-v4.ts
```diff
@@ -41,6 +41,7 @@ declare module 'fastify' {
     streamingSignatureV4?: ChunkSignatureV4Parser
     multiPartFileStream?: MultipartFile
     bodySha256: string
+    s3PostPolicyContentLengthMax?: number
```
fenos: we could potentially rename this to something like: `s3PostPolicies?: { range: [0, 1] } }`

src/storage/protocols/s3/signature-v4.ts
fenos: also i'm not sure this applies for post uploads, i'm wondering if it's only for download?

fenos: Ah yes, this field allow for min, max file size validation. 
I would say that we could remove it for now and mark it as not supported, and we can support it in a separate PR

src/storage/protocols/s3/signature-v4.ts
TylerHillery: removed normalizePolicyCondition, so we now parse + validate each condition inline in a single pass over conditions (no intermediate { operator, field, value } object per condition).

src/storage/protocols/s3/signature-v4.ts
TylerHillery: Removed content-length-range handling for now. I added a TODO for a follow up to add the enforcement.

src/storage/protocols/s3/signature-v4.ts
TylerHillery: I moved everything into the `policy.ts` to tidy things up

src/storage/protocols/s3/signature-v4.ts
```diff
@@ -230,9 +239,22 @@ export class SignatureV4 {
...
+    form.forEach((value, key) => {
+      if (typeof value !== 'string') {
+        return
+      }
+      const normalizedKey = key.toLowerCase()
+      if (normalizedKey in fields) {
+        throw ERRORS.InvalidSignature('Duplicate form field in POST policy request')
+      }
+      fields[normalizedKey] = value
+    })
```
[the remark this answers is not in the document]
TylerHillery: added

src/http/plugins/signature-v4.ts
TylerHillery: no longer needed now that we removed the content length range handling for now for a later PR

src/http/plugins/signature-v4.ts
```diff
@@ -160,6 +165,12 @@ async function authorizeRequestSignV4(
     )
   }
 
+  if (clientSignature.policy) {
+    assertPolicyConditionsSatisfied(clientSignature.policy.value, clientSignature.policy.fields, {
```
fenos: Wondering if we should move this check inside the signature-v4 `parseMultipartSignature` so that we validate it before returning it

TylerHillery: So the flow goes:

```
  authorizeRequestSignV4(request)
    |
    ├─ 1. extractSignature(request)
    |        └─ parseMultipartSignature(formData)
    |
    ├─ 2. verify(clientSignature)
    |   
    |
    └─ 3. assertPolicyConditionsSatisfied(...)
```
We don't verify the client signature until after it's parsed, and we can't check the policy conditions until the signature is verified, otherwise we could be checking against a tampered-with policy. So I think it makes sense to leave in `authorizeRequestSignV4`. Let me know your thoughts.

fenos: Ah! good point i agree!
How about moving it into the `verify` function then? 

The only reason i'm recommending to encaspulate this verification inside the sign-v4 class is so that if we use this class somewhere else we won't forget to call `assertPolicyConditionsSatisfied` after a verify

TylerHillery: good call, moved both `assertPolicyNotExpired` and `assertPolicyConditionsSatisfied` to the `verify` function so all verification is done in one place.

fenos (approved):

PR #1213, opened 2026-07-07
feat: add admin endpoint to cleanup orphan iceberg resources

itslenny:

feature

## What is the new behavior?

Introduces an admin endpoint that can be used to clean up orphan iceberg resources. These are items that were "soft deleted" by marking them as "deleted_at" but never actually removed.

- `GET tenants/iceberg/orphan-catalogs` - returns list of orphan catalogs to clean up (dry run)
- `DELETE tenants/iceberg/orphan-catalogs` - queues a `DeleteIcebergResources` job for each item

### DIFF
--- a/src/http/routes/admin/iceberg-admin.ts
+++ b/src/http/routes/admin/iceberg-admin.ts
@@ -1,0 +1,85 @@
+import { multitenantPgExecutor } from '@internal/database'
+import { DeleteIcebergResources } from '@storage/events/iceberg'
+import { FastifyInstance } from 'fastify'
+import { getConfig } from '../../../config'
+import { registerApiKeyAuth } from '../../plugins/apikey'
+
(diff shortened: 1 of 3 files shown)


src/http/routes/admin/iceberg-admin.ts
```diff
@@ -0,0 +1,85 @@
...
+        items: rows,
+      })
+    }
+  )
+
+  fastify.delete(
+    '/iceberg/orphan-catalogs',
+    { schema: { tags: ['iceberg'] } },
+    async (request, reply) => {
+      if (!isMultitenant) {
```
ferhatelmas: nit: might be good to guard if queue is enabled

ferhatelmas (approved):

fenos (approved):

PR #1222, opened 2026-07-09
chore: cleanup logflare onerror edge cases

ferhatelmas:

chore

Logflare error payload could be too big and could bloat logs. 
 
## What is the new behavior?

Cap the error body as defense in depth.

## Additional context

Related to #1221


ferhatelmas:
closing, just in case if real info in truncated part and until it becomes a real issue.

PR #1223, opened 2026-07-09
fix: gracefully ignore missing tenant when cleaning up iceberg resources

itslenny:

Bug fix

If the tenant is missing when `DeleteIcebergResources` runs the entire job fails and resources are not cleaned up

## What is the new behavior?

Ignore missing tenant in multitenant mode and still clean up iceberg resources and multitenant db entries


ferhatelmas:
Can we add a test to seal the behavior so that we don't change by mistake later?

ferhatelmas (approved):

PR #1226, opened 2026-07-13
feat: add continuous profiling and client

ferhatelmas:

feature

Profiling is done through admin handlers and requires Watt. 

## What is the new behavior?

It can be done manual as before but also can be done automatically and profiles are archived into the given profiling bucket. Triggers:

* health: a normal 30-second rotation. Profiling starts when the latest roughly one-second ELU sample is > 55%, remains active down to 45% due to hysteresis, and pauses above 80%.
* health-overload: preserved overload profile, used when the live profile cannot be retrieved after ELU crosses > 80%.
event-loop-delay-severe: current delay p99 is ≥ 1 second. Archives the active profile containing the stall when non-empty; otherwise captures the next window.
event-loop-delay: at least 3 of the most recent, up-to-5 one-second samples had delay p99 ≥ 150 ms. Captures the next ungated window.

Its client can be used to navigate all profiles. 
Manual still requires profiling bucket for archiving and client simplification.

## Additional context

It's controlled via env and zero cost if not triggered.
Dropped profile merging because it can be done offline if needed and it's rarely useful due to provenance and sampling timing.
Dropped multipart machinery for heartbeats to overcome load balancer idle timeouts (60s) because manual and auto uses the same approach.


src/internal/monitoring/pprof/automatic.ts
```diff
@@ -0,0 +1,139 @@
...
+  constructor(private readonly maxCapturesPerHour = isolateCaptureBudget) {}
+
+  sample(elu: number, delayP99Ms: number, now = Date.now(), canFire = true) {
+    this.elu.push(elu)
+    if (this.elu.length > 10) this.elu.shift()
+    this.delays.push(delayP99Ms >= profilingTriggerDelayP99Ms)
+    if (this.delays.length > 5) this.delays.shift()
+    this.captures = this.captures.filter((at) => now - at < 3_600_000)
+    const meanElu = this.elu.reduce((sum, value) => sum + value, 0) / this.elu.length
+    const reason =
```
fenos: Can we wrap this logic in a function and maybe use select statement for better redability

src/internal/monitoring/pprof/automatic.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (!profilingAutomaticEnabled) return
+  if (!profilingS3Bucket) {
+    logSchema.error(logger, '[Profiling] automatic profiling disabled', {
+      type: 'profiling',
+      error: new Error('PROFILING_S3_BUCKET is required when automatic profiling is enabled'),
+    })
+    return
+  }
+  // warm up the module
+  void loadPprof().catch(() => {})
```
fenos: if this fails shouldn't we abort all operations? it seems it is a required component?

src/internal/monitoring/pprof/automatic.ts
```diff
@@ -0,0 +1,139 @@
...
+  const histogram = monitorEventLoopDelay({ resolution: 20 })
+  const trigger = new AutomaticProfileTrigger()
+  let previous = performance.eventLoopUtilization()
+  histogram.enable()
+  const timer = setInterval(() => {
+    const current = performance.eventLoopUtilization()
```
fenos: The content of this interval could go into a `captureElu` function for better redability

src/internal/monitoring/pprof/controller.ts
```diff
@@ -0,0 +1,120 @@
...
+            kind: options.kind,
+            reason: options.reason,
+            startedAt,
+            durationSeconds: options.seconds,
+            service: options.service ?? serviceName,
+          },
+          body
+        )
+      )
+      if (options.class === 'auto') {
```
fenos: not really sure what this if statement is meant for, we are taking the snapshot in any case, the else condition simply swallow the error, is this intentional why?

src/internal/monitoring/pprof/automatic.ts
ferhatelmas: extracted a function with early return

src/internal/monitoring/pprof/automatic.ts
ferhatelmas: it would repeat loading again but disabling feature is better, returned here now

src/internal/monitoring/pprof/automatic.ts
ferhatelmas: sounds good, moved

src/internal/monitoring/pprof/controller.ts
ferhatelmas: waiting for auto but manual can fail in the background since we will get the body directly. Added a comment to explain

fenos: but we are still catching the exception upstream:
https://github.com/supabase/storage/pull/1226/changes#diff-9a2353f3ec8bfcae08c7c493f6c07c9304c84d754527b0a5b764dd833e362938R136

It seems redundant to swallow the exception here

src/internal/monitoring/pprof/controller.ts
```diff
@@ -0,0 +1,122 @@
...
+            pprof.heap.stop()
+          }
+        }
+      }
+      body = await pprof.encode(profile!)
+    } finally {
+      this.active = false
+    }
+
+    if (profilingS3Bucket) {
```
fenos: i would recommend making the ProfileController configurable with archive configuration in the constructor, this way is more "pure" and doesn't rely on the `getProfileStore()`, and `profilingS3Bucket` but is injected in the constructor.

Something like:

```ts
interface Archiver {
   archive(key: ProfileKey, body: Readable.Stream): Promise<void>
}

class S3Archiver {
    constructor(store: S3Store) {}

    archive(...) { ... }
}

new ProfileController({
  archiver?: new S3Archiver(getProfileStore()),
})
```

This way it can be used as standalone

src/internal/monitoring/pprof/automatic.ts
```diff
@@ -0,0 +1,155 @@
+import { monitorEventLoopDelay, performance } from 'node:perf_hooks'
```
fenos: maybe rename the file to `deamon` or `worker`?

src/internal/monitoring/pprof/automatic.ts
```diff
@@ -0,0 +1,155 @@
...
+  profilingMaxCapturesPerHour,
+  numWorkers,
+  profilingS3Bucket,
+  profilingSevereDelayP99Ms,
+  profilingSevereElu,
+  profilingTriggerDelayP99Ms,
+  profilingTriggerElu,
+} = getConfig()
+
+export function captureBudgetForIsolate(
```
fenos: nit: rename `captureBudgetForIsolate` => `captureBudgetForWorker`

src/internal/monitoring/pprof/automatic.ts
```diff
@@ -0,0 +1,155 @@
...
+    numericWorkerId >= workerCount
+  ) {
+    return maxCapturesPerHour
+  }
+
+  const capturesPerWorker = Math.floor(maxCapturesPerHour / workerCount)
+  return capturesPerWorker + (numericWorkerId < maxCapturesPerHour % workerCount ? 1 : 0)
+}
+
+const isolateCaptureBudget = captureBudgetForIsolate(
```
fenos: nit: `captureWorkerBudget`

src/internal/monitoring/pprof/automatic.ts
```diff
@@ -0,0 +1,155 @@
...
+  profilingMaxCapturesPerHour,
+  numWorkers,
+  resolveRuntimeIdentity().workerId
+)
+
+interface AutomaticProfilingOptions {
```
fenos: nit: `WorkerProfileOptions` or `DeamonProfileOptons`

src/internal/monitoring/pprof/automatic.ts
```diff
@@ -0,0 +1,155 @@
...
+  service: string
+  signal: AbortSignal
+}
+
+export class AutomaticProfileTrigger {
```
fenos: nit: `ProfileTrigger` no need for `Automatic`

src/internal/monitoring/pprof/automatic.ts
```diff
@@ -0,0 +1,155 @@
...
+      this.captures.length >= this.maxCapturesPerHour
+    )
+      return
+    this.captures.push(now)
+    this.cooldownUntil = now + profilingCooldownSeconds * 1000
+    return reason
+  }
+}
+
+export async function startAutomaticProfiling(options: AutomaticProfilingOptions) {
```
fenos: nit: `startProfilingWorker`

src/internal/monitoring/pprof/client-http.ts
```diff
@@ -39,97 +29,149 @@ export function resolvePprofAdminUrl(
...
+      adminUrl: options.adminUrl,
+      apiKey: options.apiKey,
+      path: `/debug/pprof/${options.type}`,
+      params: options.type === 'heap-snapshot' ? undefined : { seconds: options.seconds },
+      accept: options.type === 'heap-snapshot' ? 'application/json' : 'application/gzip',
+    })
+  )
+}
+
+export async function fetchStoredProfiles(options: {
```
fenos: nit: consistent wording, `fetchArchivedProfiles`

src/internal/monitoring/pprof/client-http.ts
```diff
@@ -39,97 +29,149 @@ export function resolvePprofAdminUrl(
...
+      date: options.date,
+      limit: options.limit,
+      cursor: options.cursor,
+    },
+    accept: 'application/json',
+  })
+  return (await response.json()) as PprofStoredProfileList
+}
+
+export async function fetchStoredProfile(options: {
```
fenos: same as above: nit consistent working `archivedProfiles`

src/internal/monitoring/pprof/client-http.ts
```diff
@@ -39,97 +29,149 @@ export function resolvePprofAdminUrl(
...
+    adminUrl: options.adminUrl,
+    apiKey: options.apiKey,
+    path: '/debug/pprof/profiles/detail',
+    params: { id: options.id },
+    accept: 'application/json',
+  })
+  return (await response.json()) as PprofStoredProfile
+}
+
+export async function downloadStoredProfile(options: {
```
fenos: same as above

src/internal/monitoring/pprof/store.ts
```diff
@@ -0,0 +1,345 @@
...
+export interface ProfileIdentity {
+  class: ProfileClass
+  kind: ProfileKind
+  service: string
+  reason: string
+  startedAt: Date
+  durationSeconds: number
+}
+
+export interface StoredProfile extends ProfileIdentity {
```
fenos: nit: `ArchivedProfile`

src/internal/monitoring/pprof/types.ts
```diff
@@ -1,86 +1,25 @@
-import type { Readable } from 'stream'
```
fenos, line 1: let's try to avoid generic `types.ts` files better to have the interfaces and types in their own owning files

src/internal/monitoring/pprof/store.ts
```diff
@@ -0,0 +1,345 @@
...
+const DAY_MILLISECONDS = 24 * 60 * 60 * 1000
+const {
+  profilingS3Bucket,
+  profilingS3Endpoint,
+  profilingS3ForcePathStyle,
+  profilingS3Region,
+  version,
+} = getConfig()
+
+function trimDashes(value: string) {
```
fenos: Wondering if we can move these helper functions in a store-key.ts file and encapsulate this logic there

src/internal/monitoring/pprof/store.ts
```diff
@@ -0,0 +1,345 @@
...
+}
+
+export class ProfileStore {
+  readonly client: S3Client
+  readonly bucket: string
+
+  constructor() {
+    if (!profilingS3Bucket) throw new Error('PROFILING_S3_BUCKET is not configured')
+    this.bucket = profilingS3Bucket
+    this.client = new S3Client({
```
fenos: if the S3Client is using global config variables, we can simply move it out from the class, and accept the `S3Client` in the constructor, better reusability too

src/internal/monitoring/pprof/types.ts
ferhatelmas: dropped

src/internal/monitoring/pprof/client-http.ts
```diff
@@ -39,97 +52,147 @@ export function resolvePprofAdminUrl(
...
-          sourceMaps: options.sourceMaps,
-        }
+async function request(options: {
+  adminUrl: string
+  apiKey: string
+  path: string
+  params?: Record<string, PprofQueryValue>
+  accept: string
+}) {
   const response = await fetch(
```
mcollina: Given you are converting this from a webstream to a node stream anyway, you are better served by using `undici.request`.

src/internal/monitoring/pprof/runtime-extension.ts
```diff
@@ -0,0 +1,623 @@
...
+interface ProfileCapturedEvent extends WorkerLifecycleEvent {
+  id: string
+  type: string
+  timestamp: number
+}
+
+interface HealthMetricsEvent extends WorkerLifecycleEvent {
+  id: string
+  healthSignals: Array<{ type: string; p99?: number; timestamp?: number }>
+}
```
mcollina: I'm adding the types for those events. https://github.com/platformatic/platformatic/pull/5032

src/config.ts
```diff
@@ -445,6 +458,46 @@ export function getConfig(options?: { reload?: boolean }): StorageConfigType {
...
+      30,
+      1,
+      300
+    ),
+    profilingCpuIntervalMicros: envIntegerInRange(
+      getOptionalConfigFromEnv('PROFILING_CPU_INTERVAL_MICROS'),
+      10_000,
+      1_000,
+      1_000_000
+    ),
```
mcollina, line 485: The default we use is 33ms. This is significantly higher than pprof own default 1ms. This is the value we found in our testing to lead to the best data with the lowest overhead.

src/internal/monitoring/pprof/client-http.ts
ferhatelmas: updated

src/config.ts
ferhatelmas: set default to 33ms

src/internal/monitoring/pprof/runtime-extension.ts
ferhatelmas: bumped to the latest for sample count and dropped custom code, when types are released, I will bump version again

ferhatelmas:
@claude review always

src/internal/monitoring/pprof/runtime-extension.ts
ferhatelmas: updated to 3.65.0 and used the new types

fenos (approved):

PR #1228, opened 2026-07-14
chore: cleanup multigres workaround for statement timeout

ferhatelmas:

chore

Multigres keeps returning zero for statement timeout via `current_setting` and required top level `show` instead.

## What is the new behavior?

https://github.com/multigres/multigres/pull/1258 fixed the problem so dropping the workaround and making it consistent for engines.

## Additional context

Related to https://github.com/supabase/storage/pull/1154#discussion_r3570608315 where the original bug was found.


fenos (approved):

GuptaManan100:
Awesome! Thank you!

PR #1229, opened 2026-07-14
fix: retry for backup object event

ferhatelmas:

Bug fix

Backup object handler doesn't throw error so failure doesn't retry.
There are some typos in logs fields.

## What is the new behavior?

Rethrow the error so queue can retry (count 5 can work).
Fix typos.

## Additional context

We should extend `logSchema.error` to extend `EventLog` and cut duplication between error and success but it's for another PR.


itslenny (approved):

fenos (approved):

ferhatelmas:
@claude review always

PR #1239, opened 2026-07-16
fix: add acceptance test for s3 objects sanitization

itslenny:
Adds an acceptance test to protect against S3 route sanitization breaking signing

### DIFF
--- a/acceptance/specs/s3.test.ts
+++ b/acceptance/specs/s3.test.ts
@@ -339,6 +339,64 @@ describeAcceptance(
         client.destroy()
       }
     })
+
+    it('correctly handles objects with special characters in key', async () => {
+      const client = createAcceptanceS3Client()
+      const bucketName = uniqueBucketName('s3chars')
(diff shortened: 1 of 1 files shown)


acceptance/specs/s3.test.ts
```diff
@@ -339,6 +339,48 @@ describeAcceptance(
         client.destroy()
       }
     })
+
+    it('puts, gets, lists, and deletes an object whose key contains a colon', async () => {
+      const client = createAcceptanceS3Client()
+      const bucketName = uniqueBucketName('s3colon')
+      const key = uniqueObjectKey('board:doc', 'bin')
```
ferhatelmas: `! $ & ' ( ) * + , : ; = @ [ ] ^ |`

let's add whole bunch

ferhatelmas (approved):

acceptance/specs/s3.test.ts
itslenny: Added all but `[ ] ^ |` which fail our object key validation. Added a separate test for those as they still pas signature verification, but fail due to the key name

ferhatelmas (approved):

PR #1243, opened 2026-07-17
fix: bump watt to latest and align config to control via env

ferhatelmas:

kinda chore but better to have a release so labeled fix

## What is the new behavior?

Bump watt to latest for leveraging [continuous profiling](https://docs.platformatic.dev/docs/guides/capture-flamegraphs-on-health-events).
Update watt config to latest schema and split config so that it can be reused for multiple applications while enablement controlled via env and keeping start order.  

## Additional context

Related to #1226
Related to #1230


src/internal/monitoring/pprof/runtime.test.ts
```diff
@@ -675,7 +675,7 @@ describe('pprof runtime helpers', () => {
 
       expect(failedToStart.message).toBe(
         'Failed to start profiling for service "storage:0": ' +
-          'Worker 0 of application storage not found. Available applications are: 4, 5'
+          'Worker 0 of application storage not found. Available workers are: 4, 5'
```
ferhatelmas, line 678: upstream error message fix

TylerHillery (approved):

PR #1244, opened 2026-07-17
fix: add acceptance test to ensure cache access isolation

itslenny:

Add acceptance check to ensure private accesses are not available via public endpoint.


acceptance/specs/cdn.test.ts
```diff
@@ -51,8 +51,8 @@ const onePixelPng = new Uint8Array(
   )
 )
 
-const SIGNED_EXPIRES_IN_S = 10
-const CACHE_RETRIES = 10
+const SIGNED_EXPIRES_IN_S = 20
+const CACHE_RETRIES = 15
```
ferhatelmas: these budgets aren't exactly compatible with each other and test timeout budget so it might expire or timeout while retrying

itslenny: I removed this change from this PR as the new test in this PR cannot be merged until the gateway fix is deployed.

ferhatelmas (approved):

PR #1246, opened 2026-07-20
fix: add finite number validation

ferhatelmas:

Bug fix

AJV coercion (i.e. `Infinity`, `1e999`) skips limit checks. 

## What is the new behavior?

Add finite check and reject them.
It's mostly relevant for S3 routes but hardened it in all routes. 
Strip finite keyword from from OpenAPI since it's a runtime specific and not compatible.

## Additional context

Related to #1237


fenos (approved):

ferhatelmas:
@claude review always

PR #1248, opened 2026-07-20
fix(config): vector s3 buckets empty env parsing

TylerHillery:

`VECTOR_S3_BUCKETS` gets set to an empty string for any region with no vector buckets configured. `''.split(',')` returns `['']` rather than `[]`, so the existing `|| []` fallback never applied. On startup this seeded a `shard_key: ''` row in the shard table for `kind: 'vector'` in every such region. This was harmless while those regions didn't have vector buckets but once a region gets enabled, the empty shard key can still get selected.

This PR filters out empty strings after splitting, so an empty/unset env var always resolves to `[]`.

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -215,3 +215,53 @@ describe('tenant pool cache config parsing', () => {
     expect(config.tenantPoolCacheMissLogSampleRate).toBe(0)
   })
 })
+
+describe('vectorS3Buckets config parsing', () => {
+  const originalValue = process.env.VECTOR_S3_BUCKETS
+
+  afterEach(() => {
+    if (originalValue === undefined) {
+      delete process.env.VECTOR_S3_BUCKETS
+    } else {
+      process.env.VECTOR_S3_BUCKETS = originalValue
+    }
+
+    vi.resetModules()
+  })
+
+  test('defaults to an empty array when VECTOR_S3_BUCKETS is unset', async () => {
+    delete process.env.VECTOR_S3_BUCKETS
+
+    const { getConfig } = await import('./config')
+    const config = getConfig({ reload: true })
+
+    expect(config.vectorS3Buckets).toEqual([])
+  })
+
+  test('defaults to an empty array when VECTOR_S3_BUCKETS is an empty string', async () => {
+    process.env.VECTOR_S3_BUCKETS = ''
+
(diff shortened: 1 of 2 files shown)


ferhatelmas (approved):

src/config.ts
```diff
@@ -675,7 +675,8 @@ export function getConfig(options?: { reload?: boolean }): StorageConfigType {
     vectorEnabled: getOptionalConfigFromEnv('VECTOR_ENABLED') === 'true',
     vectorBucketProvider: (getOptionalConfigFromEnv('VECTOR_BUCKET_PROVIDER') ||
       's3') as VectorBucketProvider,
-    vectorS3Buckets: getOptionalConfigFromEnv('VECTOR_S3_BUCKETS')?.trim()?.split(',') || [],
+    vectorS3Buckets:
+      getOptionalConfigFromEnv('VECTOR_S3_BUCKETS')?.trim()?.split(',').filter(Boolean) || [],
```
ferhatelmas: nit: might be good to trim again before filter

src/config.test.ts
```diff
@@ -215,3 +215,53 @@ describe('tenant pool cache config parsing', () => {
...
+    process.env.VECTOR_S3_BUCKETS = 'bucket-0,bucket-1,bucket-2'
+
+    const { getConfig } = await import('./config')
+    const config = getConfig({ reload: true })
+
+    expect(config.vectorS3Buckets).toEqual(['bucket-0', 'bucket-1', 'bucket-2'])
+  })
+
+  test('ignores a trailing comma', async () => {
+    process.env.VECTOR_S3_BUCKETS = 'bucket-0,bucket-1,'
```
ferhatelmas:
```suggestion
    process.env.VECTOR_S3_BUCKETS = 'bucket-0, bucket-1,'
```

ferhatelmas: re: https://github.com/supabase/storage/pull/1248/changes#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R679

src/config.ts
TylerHillery: done

src/config.test.ts
TylerHillery: done

PR #1251, opened 2026-07-21
fix: fold search path into set config

ferhatelmas:

refactor for performance

If pool is external, we do an extra round trip for search path.

## What is the new behavior?

Fold it into scope query similar to statement timeout (#1154).

## Additional context

Also added a fast path for no signal to reduce GC overhead. 
Related to #1154


ferhatelmas:
@claude review always

fenos (approved):

PR #1253, opened 2026-07-22
fix: replace xmljs with fxp and hoist builder

ferhatelmas:

Refactor for performance while improving spec compliance

xmljs is used for xml parsing/building

## What is the new behavior?

fast-xml-parser and fast-xml-builder are used instead.

## Additional context

Build 3-7x faster, parse is up to 30% faster while being more GC friendly (up to 120x in build). 
Added `text/xml` into accepted xml types.
Fixed double nesting [get bucket location](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketLocation.html). 
Added S3 namespace.
Hoist the builder.
Disabling coercing for delete, otherwise it can delete wrong keys.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,6 +50,8 @@
         "fast-decode-uri-component": "^1.0.1",
         "fast-json-stringify": "^6.3.0",
         "fast-querystring": "^1.1.2",
+        "fast-xml-builder": "^1.1.8",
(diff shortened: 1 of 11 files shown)


ferhatelmas:
@claude review always

fenos (approved):

PR #1254, opened 2026-07-22
fix: refresh on-request migration context

ferhatelmas:

Bug fix

Latest migration is set on request so that it can be used in the life cycle of the request.
On request migration runner can run migrations but it doesn't refresh the property on request.
Downstream can see old migration value.

## What is the new behavior?

Migration version is refreshed correctly.

## Additional context

This makes latest migration property more useful (skips LRU cache lookups).


ferhatelmas:
@claude review always

fenos (approved):

PR #1257, opened 2026-07-23
feat: add support for asymmetrical url signing keys, auth style key rotation and well-known

itslenny:

Feature

- Only supports symmetric (HMAC) url signing keys
- No way to create standby keys or swap them (similar to how auth key rotation works)
- No way to list tenant jwks
- JWK/JWT includes the mutable kind as a prefix in the kid - this makes swapping keys problematic

## What is the new behavior?

- Remove kind prefix from kid
- Support generating/using asymmetrical signing keys (ECDSA)
- Add endpoints to support "auth style" key rotation
  - Create standby signing keys
  - Swap standby signing keys to active (used to sign URLs via api)
  - List jwks
- Add public `.well-known/jwks.json` so asymmetric keys can be validated at the gateway


src/internal/auth/jwt.ts
```diff
@@ -86,6 +94,53 @@ export function isDownloadScopedToken(payload: { scope?: SignedUrlScope }): bool
...
+
+function getKidId(kid: string): string {
+  return kid.split('_').pop() as string
+}
+
+// A jwk's kid is formatted as "<kind>_<id>" (see JWKSManager.createJwkKid) - kind is
+// mutable (eg a standby/active swap flips it) while id is the permanent identity of the
+// underlying key. Comparing kids by their id suffix instead of requiring an exact string
+// match means a swap can never invalidate a jwt that still embeds the pre-swap kid
+function kidsReferToSameKey(keyKid: string | undefined, headerKid: string | undefined): boolean {
```
ferhatelmas: This works for us but I think this is not enough for external validators, they will require kid to be exact match. We should make kid immutable

src/internal/auth/jwt.ts
```diff
@@ -86,6 +94,53 @@ export function isDownloadScopedToken(payload: { scope?: SignedUrlScope }): bool
...
+    if (value !== undefined) {
+      publicJwk[field] = value
+    }
+  }
+  return publicJwk as unknown as JwksConfigKey
+}
+
+function getKidId(kid: string): string {
+  return kid.split('_').pop() as string
+}
```
ferhatelmas: re: immutable (https://github.com/supabase/storage/pull/1257/changes#r3720537198), if we address that comment, this won't be an issue but otherwise, we should do this only for url signing keys

src/internal/auth/jwks/manager.ts
ferhatelmas, line 64: we have similar tenant pool problem here as well, inflight can repopulate old keys and age refresh can keep them indefinitely, needs a separate fix. I will do it

src/internal/database/tenant.ts
```diff
@@ -368,10 +363,44 @@ export async function getJwtSecret(tenantId: string): Promise<{
...
   return { secret, urlSigningKey, jwks }
 }
 
+/**
+ * Get the public jwks for a tenant, suitable for exposing on a public jwks endpoint.
+ * Symmetric (oct) keys have no public form and are excluded; asymmetric keys are reduced
+ * to their public components only.
+ * @param tenantId
+ */
+export async function getPublicJwks(tenantId: string): Promise<JwksConfigKey[]> {
```
ferhatelmas, line 398: since it's used by well known handler, shouldn't we add JWT_JWKS for single tenant?

src/app.ts
```diff
@@ -104,6 +106,9 @@ const build = (opts: buildOpts = {}): FastifyInstance => {
     reply.send(version)
   })
   app.get('/status', async (request, response) => response.status(200).send())
+  app.get('/.well-known/jwks.json', async (request, reply) => {
+    reply.send({ keys: await getPublicJwks(request.tenantId) })
+  })
```
ferhatelmas, line 113: shall we add 10min caching similar to auth?

src/app.ts
```diff
@@ -69,6 +70,7 @@ const build = (opts: buildOpts = {}): FastifyInstance => {
     '/healthcheck',
     '/version',
     '/documentation',
+    '/.well-known/jwks.json',
```
ferhatelmas: how are we going to have visibility? a simple counter might be useful

src/internal/auth/jwks/manager.ts
```diff
@@ -1,23 +1,32 @@
...
+export const JWK_KIND_STORAGE_URL_SIGNING = 'storage-url-signing-key'
+export const JWK_KIND_STORAGE_URL_STANDBY = 'storage-url-standby-key'
+export const JWK_KID_SEPARATOR = '_'
 
-const tenantJwksSingleFlight = createSingleFlightByKey<JwksConfig>()
 // Max 16,384 items. At ~2.5KB per JWKS, this uses roughly ~40MB of heap memory worst-case.
 export const TENANT_JWKS_CACHE_MAX_ITEMS = 16384
 export const TENANT_JWKS_CACHE_TTL_MS = 1000 * 60 * 60 // 1h
 
+const tenantJwksSingleFlight = createSingleFlightByKey<JwksConfig>()
```
ferhatelmas: https://github.com/supabase/storage/pull/1257/changes#r3720597931

we need to use invalidatable one here and get rid of refresh

src/internal/database/tenant.ts
itslenny: This is already included 

getPublicJwks() -> getTenantJwks() -> getSingleTenantJwtConfig() -> getConfig().jwtJWKS -> getOptionalConfigFromEnv('JWT_JWKS') || null

src/internal/auth/jwt.ts
itslenny: Changed to immutable, but we still need this as JWTs signed before this change will have the kind prefix, but it will they will always start with `storage-url-signing-key_` so the check is simpler. We never signed URLs with any other kind.

src/app.ts
itslenny: I'm just going to include it in logging for now since it's new, if it's too noisy we can change it later

src/http/routes/admin/jwks.ts
```diff
@@ -37,6 +56,36 @@ const updateSchema = {
...
+    type: 'object',
+    properties: {
+      type: {
+        type: 'string',
+        enum: URL_SIGNING_JWK_TYPES,
+      },
+    },
+    required: ['type'],
+  },
+} as const
```
ferhatelmas: they are same, one instead?

src/http/routes/admin/jwks.ts
```diff
@@ -46,29 +95,48 @@ interface JwksAddRequestInterface extends RequestGenericInterface {
...
+}
+
 type ValidationResult = { message: string } | undefined
 
 function validateAddJwkRequest({ jwk, kind }: JwksAddRequestInterface['Body']): ValidationResult {
-  if (kind.includes('_')) {
-    return { message: 'Kind cannot contain underscore characters' }
+  if (kind.includes(JWK_KID_SEPARATOR)) {
+    return { message: `Kind cannot contain restricted character "${JWK_KID_SEPARATOR}"` }
   }
```
ferhatelmas: do we still need this restriction?

src/http/routes/admin/jwks.ts
```diff
@@ -46,29 +95,48 @@ interface JwksAddRequestInterface extends RequestGenericInterface {
...
 
   if (kind.length > 50) {
     return { message: 'Kind cannot exceed 50 characters' }
   }
 
+  if (kind === JWK_KIND_STORAGE_URL_SIGNING || kind === JWK_KIND_STORAGE_URL_STANDBY) {
+    return { message: `Cannot create add a jwk using reserved kind "${kind}"` }
```
ferhatelmas: `Cannot add ...`

src/internal/auth/jwks/manager.ts
```diff
@@ -60,17 +61,49 @@ export class JWKSManager<TRX> {
...
+      JWK_KIND_STORAGE_URL_STANDBY
+    )
   }
 
   /**
    * Atomically rolls the URL signing JWK by deactivating the current key and creating a new one
    * @param tenantId
    */
-  async rollUrlSigningJwk(tenantId: string): Promise<{ oldKid: string | null; newKid: string }> {
+  async rollUrlSigningJwk(
```
ferhatelmas, line 115: to be safe, this should take the lock as well

src/internal/auth/jwks/kid.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Strips a legacy "<kind>_" prefix, but only for the (pre-existing) signing kind - other kids
+ * (custom addJwk kinds, standby kids, or the current bare-id format) are returned unchanged.
+ */
+export function normalizeUrlSigningKid(kid: string): string {
+  return kid.startsWith(LEGACY_URL_SIGNING_KID_PREFIX)
+    ? kid.slice(LEGACY_URL_SIGNING_KID_PREFIX.length)
+    : kid
+}
```
ferhatelmas, line 15: just noting, while deploying, old nodes will fail since they require the prefix but it should be quick

src/storage/limits.ts
```diff
@@ -167,8 +167,12 @@ export function parseFileSizeToBytes(valueWithUnit: string) {
   }
 }
 
+export const UUID_PATTERN =
+  '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-5][0-9a-fA-F]{3}-[089abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$'
+const UUID_REGEX = new RegExp(UUID_PATTERN)
```
ferhatelmas, line 172: that's nice

src/app.ts
```diff
@@ -104,6 +105,10 @@ const build = (opts: buildOpts = {}): FastifyInstance => {
     reply.send(version)
   })
   app.get('/status', async (request, response) => response.status(200).send())
+  app.get('/.well-known/jwks.json', async (request, reply) => {
+    reply.header('cache-control', 'public, max-age=600')
```
ferhatelmas, line 110: just in case add `Vary: X-Forwarded-Host` for caching

ferhatelmas, line 110: might be better to set headers after getting response so that error isn't cached

src/internal/auth/jwks/kid.ts
ferhatelmas: bare match then normalized check would handle

src/internal/database/tenant.ts
ferhatelmas: how about standalone url signing key?

src/http/routes/admin/jwks.ts
itslenny: nope. removed.

src/http/routes/admin/jwks.ts
itslenny: merged into one schema

src/internal/database/tenant.ts
itslenny: Do you mean the jwt_secret? If so, it is symmetrical and has no public part so it's not needed here.

src/internal/auth/jwks/manager.ts
```diff
@@ -125,10 +179,10 @@ export class JWKSManager<TRX> {
     return tenantJwksSingleFlight(tenantId, async () => {
       const data = await this.storage.listActive(tenantId)
 
-      let urlSigningKey: JwksConfigKeyOCT | undefined
+      let urlSigningKey: UrlSigningJwksConfigKey | undefined
```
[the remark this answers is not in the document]
itslenny: Regression from rebase. fixed.

src/internal/database/tenant.ts
ferhatelmas: if url signing key isn't repeated in keys because type shape totally makes sense

src/internal/database/tenant.ts
```diff
@@ -375,10 +370,44 @@ export async function getJwtSecret(tenantId: string): Promise<{
...
+  const { jwks } = await getTenantJwks(tenantId)
+
+  const publicJwks: JwksConfigKey[] = []
+  for (const jwk of jwks.keys) {
+    try {
+      publicJwks.push(toPublicJwk(jwk))
+    } catch {
+      // toPublicJwk throws for "oct" (no public form) and for any kty it doesn't recognize
+      // the error can be ignored and these keys will be excluded from the public jwk list
+    }
```
ferhatelmas: nit: I think it would be better not to use try/catch because it could hide some other hidden bugs. Checking via a utility if publicable might be better

src/internal/auth/jwt.ts
```diff
@@ -182,6 +191,50 @@ function getPreparedJWTVerificationKey(
   return importedKey
 }
 
+// RFC 7517 §6 - the JWK parameters that fully describe the PUBLIC half of each asymmetric
+// key type. Deliberately an allowlist rather than a denylist of private fields: any parameter not named here is dropped
+// "oct" (symmetric) keys have no public form at all and are intentionally absent from this map.
+const PUBLIC_JWK_FIELDS: Partial<Record<JwksConfigKey['kty'], readonly string[]>> = {
```
ferhatelmas, line 259: not for us but self host env keys could have `use` / `key_ops` if so, we shouldn't strip them. 

I don't think worth, just noting

src/internal/auth/jwt.ts
```diff
@@ -182,6 +191,50 @@ function getPreparedJWTVerificationKey(
...
+ * @throws for keys that have no public fields (e.g. "oct" symmetric keys)
+ */
+export function toPublicJwk(jwk: JwksConfigKey): JwksConfigKey {
+  const publicFields = PUBLIC_JWK_FIELDS[jwk.kty]
+  if (!publicFields) {
+    throw new Error(`Cannot derive a public jwk for kty "${jwk.kty}"`)
+  }
+
+  const publicJwk: Record<string, unknown> = { kty: jwk.kty }
+  for (const field of [...publicFields, ...PUBLIC_JWK_METADATA_FIELDS]) {
```
ferhatelmas: nit: it's small but it might be better to add metadata fields directly into public fields so that there is less garbage here

src/internal/auth/jwt.ts
```diff
@@ -182,6 +191,50 @@ function getPreparedJWTVerificationKey(
...
+// Jwk's kid was simplified to use just the tenants_jwks row's bare uuid
+// Historically we embedded the kind resulting in a kid in the format "<kind>_<id>"
+// So a header's kid must be normalized to its id suffix before comparing against a jwk's (bare) kid
+// this is to support existing JWTs that were already signed using the legacy format
+function kidsMatch(keyKid: string | undefined, headerKid: string | undefined): boolean {
+  if (keyKid === undefined || headerKid === undefined) {
+    return false
+  }
+  // Bare-to-bare is the common case (and the only one post-rollout), so check it directly first.
+  return keyKid === headerKid || keyKid === normalizeUrlSigningKid(headerKid)
```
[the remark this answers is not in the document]
ferhatelmas: this is existing, so I think it's fine to ignore here but are we going to have a separation between auth and url sign eventually?

src/internal/auth/jwks/manager.ts
ferhatelmas: addressed in https://github.com/supabase/storage/pull/1316

src/internal/auth/jwks/manager.ts
ferhatelmas: race is handled in https://github.com/supabase/storage/pull/1316 and refresh is separate since performance

src/internal/auth/jwt.ts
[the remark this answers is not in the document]
itslenny: This is/was intentional, but I think it might make sense in the future to add the ability to separate these keys. However, afaik no one has asked for this functionality yet.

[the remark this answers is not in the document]
ferhatelmas: agreed, out of scope here

src/internal/auth/jwks/kid.ts
ferhatelmas, line 15: While deploying, new instance will generate kid by id but if request hits to old instance, it will reject. It's a short time but possible. 

Shall we add similar utility (kind+sep+id or id) and deploy it then deploy this version?

src/storage/events/jwks/jwks-create-signing-secret.ts
```diff
@@ -35,10 +37,10 @@ export class JwksCreateSigningSecret extends BaseEvent<JwksCreateSigningSecretPa
   }
 
   static async handle(job: Job<JwksCreateSigningSecretPayload>) {
-    const { tenantId, sbReqId } = job.data
+    const { tenantId, sbReqId, keyType } = job.data
```
ferhatelmas, line 41: noting: controlled by us but bumping queue name would be safer

src/storage/events/jwks/jwks-create-signing-secret.ts
```diff
@@ -35,10 +37,10 @@ export class JwksCreateSigningSecret extends BaseEvent<JwksCreateSigningSecretPa
...
 
     try {
-      const { kid } = await jwksManager.generateUrlSigningJwk(tenantId)
+      const { kid } = await jwksManager.generateUrlSigningJwk(tenantId, keyType || 'HS512')
```
ferhatelmas: nit: maybe `HS512` as a fallback constant

src/internal/auth/jwks/store-pg.ts
ferhatelmas, line 199: per state machine, it won't be called concurrently or on signing key but shall we add `kind` to filter to ensure standby?

src/internal/auth/jwks/kid.ts
itslenny: I pulled the kid normalization code into it's own PR so we can deploy it first.

https://github.com/supabase/storage/pull/1331

src/internal/auth/jwks/store-pg.ts
itslenny: This can be called on a key of any kind. It is used in rollURlSigningJwk to deactivate the active key. It can also be used to deactivate custom keys that were added which can have any kind (except the reserved ones). Since the kid is unique (primary key) there is no concern of updating the wrong key.

src/storage/events/jwks/jwks-create-signing-secret.ts
itslenny: I don't think this is worth changing. This is essentially dead code. We used it to backfill jwks for all tenants when we first introduced this and will probably never use it again. I thought about just removing it, but if there is ever a situation where tenants aren't getting jwks created we might be happy this is an option. Same goes for the roll endpoint. It has never been used in production, but if we ever need to quickly rotate secrets for a leak we'll be happy to have the option. 

Only list, swap, add jwk, and generate standby key will be exposed via the platform api.

src/internal/auth/jwks/store-pg.ts
itslenny: This did make me realize that it's possible to deactivate the current url signing key which would leave the tenant with no url signing key at all. I added a check at the endpoint layer to explicitly prevent this. We can still use the roll endpoint and from an end user perspective they should create a standby, swap, and then deactivate

ferhatelmas: exactly, this was my concern

src/internal/auth/jwt.ts
ferhatelmas: this becomes statistically true with the feature, we should use the cached one unconditionally, it already handles internally

src/http/routes/admin/jwks.ts
```diff
@@ -129,23 +176,70 @@ export default async function routes(fastify: FastifyInstance) {
         params: { tenantId, kid },
         body: { active },
       } = request
+      const { urlSigningKey } = await getJwtSecret(tenantId)
+      if (typeof urlSigningKey !== 'string' && kid === urlSigningKey.kid) {
+        return reply.status(409).send({ error: 'Cannot deactivate the current url signing key' })
+      }
```
ferhatelmas: nit: shall we add `active:true` into condition as noop?

src/http/routes/admin/jwks.ts
```diff
@@ -129,23 +176,70 @@ export default async function routes(fastify: FastifyInstance) {
...
       const result = await jwksManager.toggleJwkActive(tenantId, kid, active)
```
ferhatelmas: nit: this can throw 500 if already existing active. By normal client, it won't be the case but might be better to catch and return 400/409

src/http/routes/admin/jwks.ts
```diff
@@ -46,29 +78,44 @@ interface JwksAddRequestInterface extends RequestGenericInterface {
...
 }
 
-interface JwksRollRequestInterface extends RequestGenericInterface {
+interface JwksListRequestInterface extends RequestGenericInterface {
+  Params: {
+    tenantId: string
+  }
+}
+
+interface JwksGenerateStandbyRequestInterface extends RequestGenericInterface {
```
ferhatelmas: nit: this seems same with `JwksRollRequestInterface`

src/http/routes/admin/jwks.ts
ferhatelmas: noting, this is using cached version, should we keep this check in db layer just in case?

itslenny: I was trying to avoid an extra DB call, but it does seem safer to check the db on this one. Additionally, this allows returning a 404 for not found keys and checking the kind of the key directly. Seems cleaner all around.

After thinking through this I realized there is never a valid reason to toggle a url signing key. Activating one should cause a unique constraint violation, and deactivating one causes there to be no signing key. This also resolves the [potential 500 raised here](https://github.com/supabase/storage/pull/1257#discussion_r3823289388).

However, that opens the possibility of creating a key that cannot be reactivated using the roll endpoint. Therefore, I refactored roll to use the same standby/swap pattern using the existing methods. We've never used the roll endpoint so there shouldn't be any already in this state, but if there are we can fix them manually as a one off.

src/internal/auth/jwks/manager.ts
```diff
@@ -113,16 +113,25 @@ export class JWKSManager<TRX> {
     type: UrlSigningJwkType
   ): Promise<{ oldKid: string | null; newKid: string }> {
     return this.storage.transaction(async (trx) => {
-      await this.storage.lockKeyByKind(trx, tenantId, JWK_KIND_STORAGE_URL_SIGNING)
       const currentKeys = await this.storage.listActive(tenantId, JWK_KIND_STORAGE_URL_SIGNING, trx)
       const currentKey = currentKeys[0]
```
ferhatelmas: there is a race here, current key is read outside of locked critical section

src/internal/auth/jwks/store-pg.ts
ferhatelmas: I rechecked this, endpoint check can only protect serial calls. If toggleActive is blocked on row lock of concurrent roll, it can deactivate the single signing key. We still need the kind check here to prevent it

ferhatelmas: nit: change looks good. We could also add a concurrency test but leaving it your judgment

ferhatelmas (approved):

PR #1269, opened 2026-07-27
fix: update pool reliably upon config changes

ferhatelmas:

Bug fix

Tenant config could change (URL, max connections, delete) while processes still held stale pools or stale single-flight loads. Rebalance/destroy was incomplete and racy under concurrent invalidation.

## What is the new behavior?

Pools are cached by tenant and database url so stale requests can't revert the endpoint.
Add invalidatable single flight so that an old load can't populate the cache with wrong config.

### DIFF
--- a/src/internal/database/pg-connection.ts
+++ b/src/internal/database/pg-connection.ts
@@ -143,7 +143,6 @@ const transactionSetupMaxAttempts = 11
 const transactionSetupTotalBudgetMs = 3000
 const transactionSetupRetryMinDelayMs = 50
 const transactionSetupRetryMaxDelayMs = 200
-const endedPoolErrorMessage = 'Cannot use a pool after calling end on the pool'
 
 async function waitForPoolDrains(drains: Promise<void>[]): Promise<void> {
   const results = await Promise.allSettled(drains)
@@ -179,11 +178,7 @@ export class PgPoolStrategy {
   acquire(): PgPoolExecutor {
     const pool = this.getPool()
     if (!this.executor || this.executorPool !== pool) {
-      this.executor = new PgPoolExecutor(
-        pool,
-        Boolean(this.options.isExternalPool),
-        this.endpointScope
-      )
+      this.executor = new PgPoolExecutor(pool, Boolean(this.options.isExternalPool))
       this.executorPool = pool
     }
     return this.executor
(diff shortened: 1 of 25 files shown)


src/internal/database/pg-connection.ts
```diff
@@ -141,53 +169,167 @@ export class PgPoolStrategy {
...
+    let originalPool: Pool | undefined
+
+    if (settings.configRevision !== undefined) {
+      const appliedRevision = this.options.configRevision
+      const revisionComparison = compareConfigRevisions(settings.configRevision, appliedRevision)
+
+      if (revisionComparison >= 0) {
+        const endpointChanged =
+          settings.dbUrl !== this.options.dbUrl ||
+          Boolean(settings.isExternalPool) !== Boolean(this.options.isExternalPool)
```
fenos: I'm not sure this statement is correct. 
Why we care about isExternalPool is different?

src/internal/database/pg-connection.ts
```diff
@@ -488,6 +676,10 @@ export class PgPoolExecutor implements DatabaseTransactionalExecutor {
   }
 }
 
+export interface PgPinnedExecutor extends DatabaseTransactionalExecutor {
```
fenos: Not sure what "Pinned" means here, not clear

src/internal/database/pool.ts
```diff
@@ -81,29 +97,27 @@ const multiTenantTtlConfig = {
   checkAgeOnGet: true,
 }
 
-const manuallyDestroyedPools = new WeakSet<PoolStrategy>()
+const retiringPools = new WeakSet<PoolStrategy>()
+const inFlightRetirements = new Set<Promise<void>>()
```
fenos: we could remove `inFlightRetirements` and only track `retiringPools` we can set 1 boolean flag `isRetiring` and `retiringDone: Promise<void>`

ferhatelmas: just having a flag on pool isn't enough since we need to iterate and wait for shutdown but updated it to track them together

src/internal/database/pg-connection.ts
ferhatelmas: pgvector cache was pinning the pool but since it's not practical issue, dropped it

src/internal/database/pg-connection.ts
ferhatelmas: it changes tx behavior and it's captured when strategy is created so we need to check but split/renamed it to be clearer

ferhatelmas:
@claude review always

src/internal/database/pool.ts
```diff
@@ -259,18 +268,21 @@ export abstract class PoolManager<TPool extends PoolStrategy = PoolStrategy> {
...
   getPool(settings: TenantConnectionOptions): TPool {
-    const existingPool = tenantPools.get(settings.tenantId)
+    const cacheKey = tenantPoolCacheKey(settings)
+    const existingPool = tenantPools.get(cacheKey)
     const outcome: CacheLookupOutcome = existingPool ? 'hit' : 'miss'
     recordTenantPoolCacheLookup(settings, outcome)
 
     if (existingPool) {
+      existingPool.rebalance({ maxConnections: settings.maxConnections })
       return existingPool as TPool
```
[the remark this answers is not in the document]
ferhatelmas: as documented, will leave it

fenos (approved):

PR #1270, opened 2026-07-28
fix(storage): document 200 responses for vector bucket CRUD endpoints

grdsdev:
## Context
The SDK team is starting to use the generated OpenAPI spec to generate idiomatic API clients for the Supabase SDKs. We're starting with vector buckets for testing. This PR is the first of several that will make the fastify-generated OpenAPI spec ready for client generation, split by route family to keep each one reviewable.

## Summary
First slice of #1215 (too large to review as one PR), split by route family. This PR covers only the vector bucket routes.

- Adds `vectorBucketSchema`, `getVectorBucketResponseSchema`, `listVectorBucketsResponseSchema` (mirroring `bucketSchema`/`objectSchema`), and wires them into `create-bucket`/`delete-bucket`/`get-bucket`/`list-buckets` so the exported spec has success schemas for the vector CRUD endpoints, not just error responses.
- Adds `src/http/routes/openapi-transform.ts`: derives stable `operationId`s from each route's existing `config.operation`, names swagger components by `$id` instead of `def-N`, folds duplicate trailing-slash paths, and documents a generic 4xx (doc-only, covering 403 etc.) for any route without one. Prerequisite infra for generating a usable client from any route family, starting here with vector.

Deferred to follow-up PRs: bucket/object/iceberg/render/tus/cdn/health route documentation, `error-handler.ts` formatter support, `apikey.ts`, typespec tooling additions.

## Test plan
- [x] `tsc -noEmit` — no errors
- [x] `biome check` on touched files — clean
- [x] `vitest run` on `openapi-transform.test.ts` — 5/5 passing

### DIFF
--- a/src/http/plugins/log-request.ts
+++ b/src/http/plugins/log-request.ts
@@ -39,11 +39,6 @@ declare module 'fastify' {
 
   interface FastifyContextConfig {
     operation?: string
-    // Explicit OpenAPI operationId override - takes precedence over the id derived from
-    // `operation` (see src/http/routes/openapi-transform.ts). Set this when a route's
-    // generated SDK method name needs to stay stable regardless of derivation/collision
-    // rules, or to resolve a genuine duplicate operationId.
-    operationId?: string
     resources?: BivariantHandler<[req: FastifyRequest], string[]>
     logMetadata?: BivariantHandler<[req: FastifyRequest], Record<string, unknown>>
   }
--- a/src/http/plugins/log-request.ts
+++ b/src/http/plugins/log-request.ts
@@ -39,6 +39,11 @@ declare module 'fastify' {
 
   interface FastifyContextConfig {
     operation?: string
+    // Explicit OpenAPI operationId override - takes precedence over the id derived from
+    // `operation` (see src/http/routes/openapi-transform.ts). Set this when a route's
+    // generated SDK method name needs to stay stable regardless of derivation/collision
+    // rules, or to resolve a genuine duplicate operationId.
+    operationId?: string
     resources?: BivariantHandler<[req: FastifyRequest], string[]>
     logMetadata?: BivariantHandler<[req: FastifyRequest], Record<string, unknown>>
   }
--- a/src/http/routes/openapi-transform.test.ts
+++ b/src/http/routes/openapi-transform.test.ts
@@ -1,0 +1,63 @@
+import { FastifySchema, RouteOptions } from 'fastify'
+import { createOpenApiTransform } from './openapi-transform'
+import { ROUTE_OPERATIONS } from './operations'
+
+function routeWithOperation(operation?: string): RouteOptions {
+  return {
+    method: 'POST',
+    url: '/object/:bucketName/*',
+    config: operation ? { operation } : undefined,
+  } as RouteOptions
+}
+
+describe('documentMultipartUploadBody (via transformOpenApiSchema)', () => {
+  const multipartOperations = [
+    ROUTE_OPERATIONS.CREATE_OBJECT,
+    ROUTE_OPERATIONS.UPDATE_OBJECT,
+    ROUTE_OPERATIONS.UPLOAD_SIGN_OBJECT,
+  ]
+
+  it.each(multipartOperations)('documents the multipart body for %s', (operation) => {
+    const transform = createOpenApiTransform()
+    const { schema } = transform({
+      schema: {} as FastifySchema,
+      url: '/object/:bucketName/*',
+      route: routeWithOperation(operation),
+    })
+
+    expect(schema.consumes).toEqual(['multipart/form-data'])
+    expect(schema.body).toEqual({
+      type: 'object',
+      properties: {
+        cacheControl: { type: 'string' },
+        metadata: { type: 'string' },
+        file: { type: 'string', contentEncoding: 'binary' },
+      },
+      required: ['cacheControl', 'file'],
+    })
+  })
+
+  it('leaves an unrelated operation unchanged', () => {
+    const transform = createOpenApiTransform()
+    const { schema } = transform({
+      schema: {},
+      url: '/object/:bucketName',
+      route: routeWithOperation(ROUTE_OPERATIONS.GET_AUTH_OBJECT),
+    })
+
+    expect(schema.body).toBeUndefined()
+    expect(schema.consumes).toBeUndefined()
+  })
+
+  it('leaves a route with no operation unchanged', () => {
+    const transform = createOpenApiTransform()
+    const { schema } = transform({
+      schema: {},
+      url: '/object/:bucketName',
+      route: routeWithOperation(undefined),
+    })
+
+    expect(schema.body).toBeUndefined()
+    expect(schema.consumes).toBeUndefined()
+  })
+})
--- a/src/http/routes/openapi-transform.test.ts
+++ b/src/http/routes/openapi-transform.test.ts
@@ -78,7 +78,8 @@ describe('operationId (via transformOpenApiSchema)', () => {
     expect(schema.operationId).toBe('objectGetAuthenticatedHead')
   })
 
-  it('throws on a genuine operationId collision instead of silently suffixing', () => {
+  it('warns and leaves a genuine operationId collision undocumented instead of silently suffixing', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {})
     const transform = createOpenApiTransform()
     transform({
       schema: {},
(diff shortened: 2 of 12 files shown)


src/http/routes/openapi-transform.ts
```diff
@@ -0,0 +1,249 @@
...
+
+/**
+ * OpenAPI requires operationId to be unique across the whole document. `exposeHeadRoutes`
+ * auto-derives a HEAD operation from every GET route re-using the same `config.operation`,
+ * so the id needs a per-method suffix to stay unique when that happens.
+ * Returns a fresh transform bound to its own dedup state, so main/admin specs don't
+ * leak collisions into each other when generated in the same process (see export-docs.ts).
+ */
+export function createOpenApiTransform() {
+  const seenIds = new Set<string>()
```
ferhatelmas: This makes ordering important so maintenance problem. If we change route ordering in a file, it will break SDKs. I would suggest adding explicit config directly to routes for documentation and validate for uniqueness. Auto generated head routes could receive `Head` suffix 

```js
config: {
  operation: ROUTE_OPERATIONS.GET_AUTH_OBJECT,
	openapiOperation: "yourid"
}
```

src/http/routes/openapi-transform.ts
```diff
@@ -0,0 +1,249 @@
...
+    ...schema,
+    consumes: ['multipart/form-data'],
+    body: {
+      type: 'object',
+      properties: {
+        cacheControl: { type: 'string' },
+        metadata: { type: 'string' },
+        file: { type: 'string', contentEncoding: 'binary' },
+      },
+      required: ['cacheControl', 'file'],
```
ferhatelmas: `cacheControl` isn't required, defaults to `no-cache`

src/http/routes/openapi-transform.ts
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  return {
+    ...schema,
+    consumes: ['multipart/form-data'],
+    body: {
+      type: 'object',
+      properties: {
+        cacheControl: { type: 'string' },
+        metadata: { type: 'string' },
```
ferhatelmas: it also accepts `userMetadata` and `contentType`

src/http/routes/openapi-transform.ts
```diff
@@ -0,0 +1,249 @@
...
+ * createObject/updateObject/uploadSignedObject never set `schema.body` - they read the raw
+ * multipart stream directly via `uploadFromRequest` -> `fileUploadFromRequest`
+ * (see src/storage/object.ts), without registering `@fastify/multipart`'s
+ * `attachFieldsToBody`. That means `request.body` is always `undefined` on these routes, so a
+ * real `schema.body` would make Fastify validate that `undefined` (substituted as `null`)
+ * against a required-fields schema on every real upload and fail it with a 400 - a production
+ * regression, not a docs improvement. Document the multipart shape here instead, transform-only,
+ * where - like `defaultErrorResponse` above - it can never reach live request validation.
+ */
+function documentMultipartUploadBody(schema: FastifySchema, route: RouteOptions): FastifySchema {
```
ferhatelmas: we also support raw uploads so we need to document it as well? 

I would suggest dropping multipart from this PR and could be added separately in a follow up

src/http/routes/openapi-transform.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  return {
+    ...schema,
+    consumes: ['multipart/form-data'],
+    body: {
+      type: 'object',
+      properties: {
+        cacheControl: { type: 'string' },
+        metadata: { type: 'string' },
+        file: { type: 'string', contentEncoding: 'binary' },
```
ferhatelmas:
```suggestion
        file: { type: 'string', format: 'binary' },
```

https://spec.openapis.org/oas/v3.0.3#data-types

src/http/routes/openapi-transform.ts
```diff
@@ -0,0 +1,257 @@
...
+ * message, code} body (or, for a subtree with its own formatter, whatever that subtree
+ * documents instead) - default the doc to that shape for any otherwise-undocumented 4xx.
+ * Doc-only on purpose: several handlers reply with an ad-hoc, partial error body directly
+ * (`reply.status(400).send({message: '...'})`, bypassing the formatter entirely), and an
+ * earlier version of this defaulted via a real onRoute hook that made Fastify enforce
+ * errorSchema's `required` fields during response *serialization* - which threw on exactly
+ * those ad-hoc replies (fast-json-stringify errors on a missing required property rather
+ * than dropping it). A transform can't affect request handling, so it can't cause that.
+ */
+function defaultErrorResponse(schema: FastifySchema | undefined): FastifySchema {
```
ferhatelmas: this is not aligned with iceberg https://github.com/supabase/storage/blob/c57fbf24e439c970453c82535a09199712ae7a9f/src/http/routes/iceberg/index.ts#L41

src/http/routes/openapi-transform.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+function routeWithOperation(operation?: string): RouteOptions {
+  return {
+    method: 'POST',
+    url: '/object/:bucketName/*',
+    config: operation ? { operation } : undefined,
+  } as RouteOptions
+}
+
+describe('documentMultipartUploadBody (via transformOpenApiSchema)', () => {
```
ferhatelmas: would be nice to see a few tests vector response and error envelope

src/http/routes/openapi-transform.ts
grdsdev: Fixed in 1fc0f5547b3941f58a3be72995becd0fce54d0b9

src/http/routes/openapi-transform.ts
grdsdev: Fixed in d21b0d1034435bb7481fc3e0bbbea9b786ee5131

src/http/plugins/log-request.ts
```diff
@@ -39,6 +39,11 @@ declare module 'fastify' {
 
   interface FastifyContextConfig {
     operation?: string
+    // Explicit OpenAPI operationId override - takes precedence over the id derived from
+    // `operation` (see src/http/routes/openapi-transform.ts). Set this when a route's
+    // generated SDK method name needs to stay stable regardless of derivation/collision
+    // rules, or to resolve a genuine duplicate operationId.
+    operationId?: string
```
ferhatelmas: I think we should get rid of this, and fallback to standard `schema.operationId`

src/http/routes/vector/delete-bucket.ts
ferhatelmas: seems there is a typo here: this should be delete

could you fix while we're here?

src/http/routes/vector/get-bucket.ts
ferhatelmas: yup, same thing here

src/http/routes/openapi-transform.test.ts
```diff
@@ -0,0 +1,183 @@
...
+    return { method: 'GET', url } as RouteOptions
+  }
+
+  it('defaults an undocumented route to the standard error shape', () => {
+    const transform = createOpenApiTransform()
+    const { schema } = transform({
+      schema: {},
+      url: '/object/:bucketName',
+      route: routeAt('/object/:bucketName'),
+    })
```
ferhatelmas, line 17: these tests constructing manually, good for testing transformer but they don't validate if generated contract works as expected, we need tests for the full contract, something in `app.test.ts`

src/http/routes/openapi-transform.ts
```diff
@@ -0,0 +1,238 @@
...
+        .split('_')
+        .filter(Boolean)
+        .map((word, i) => (i === 0 ? word : word[0].toUpperCase() + word.slice(1)))
+        .join('')
+    )
+    .map((part, i) => (i === 0 ? part : part[0].toUpperCase() + part.slice(1)))
+    .join('')
+}
+
+const NON_STANDARD_ERROR_SHAPE_PATH_PREFIX = '/iceberg'
```
ferhatelmas: we also need to exclude tus (`/upload/resumable`)

src/storage/schemas/vector.ts
```diff
@@ -24,3 +24,44 @@ const vectorIndex = {
...
+// OpenAPI-documented response schemas for the vector bucket CRUD endpoints
+// (src/http/routes/vector/{create,get,list,delete}-bucket.ts). Registered on
+// Fastify via app.addSchema so they appear as named components in the
+// exported spec, matching bucketSchema/objectSchema.
+export const vectorBucketSchema = {
+  $id: 'vectorBucketSchema',
+  type: 'object',
+  properties: {
+    vectorBucketName: { type: 'string' },
+    creationTime: {
```
ferhatelmas, line 37: this is optional than null

src/http/routes/openapi-transform.ts
```diff
@@ -0,0 +1,238 @@
...
+ * object) already reuse the same ROUTE_OPERATIONS constant across multiple registrations
+ * (e.g. POST / and POST /*), so this can't hard-fail doc generation for the whole app over
+ * a pre-existing duplicate it doesn't own. Warn and leave the colliding route without an
+ * operationId instead - no worse than before this transform existed, and each occurrence
+ * is a route family that should get its own config.operationId in a follow-up.
+ * Returns a fresh transform bound to its own dedup state, so main/admin specs don't
+ * leak collisions into each other when generated in the same process (see export-docs.ts).
+ */
+export function createOpenApiTransform() {
+  const seenIds = new Set<string>()
```
ferhatelmas: ` Map<operationId, "METHOD /url">` would be safer to run it for different formats (json, yaml)

PR #1273, opened 2026-07-28
fix: purge CDN cache on bucket update or delete

itslenny:

Bug fix

- If a public bucket becomes private the objects are still available via the public endpoint until the CDN cache expires
- If a bucket is deleted the objects are still available via CDN cache

## What is the new behavior?

Proactively purge CDN cache if a public bucket becomes private or a bucket is deleted.

If a private bucket becomes public the objects become public and there is no need to purge the CDN and we can just let it expire naturally.

### DIFF
--- a/src/storage/storage.ts
+++ b/src/storage/storage.ts
@@ -1,6 +1,7 @@
 import { tenantHasFeature } from '@internal/database'
 import { ERRORS, StorageBackendError } from '@internal/errors'
 import { logger, logSchema } from '@internal/monitoring'
+import { CdnCacheManager } from '@storage/cdn/cdn-cache-manager'
 import { BucketCreatedEvent, BucketDeleted } from '@storage/events'
 import { StorageObjectLocator } from '@storage/locator'
 import { InfoRenderer } from '@storage/renderer/info'
@@ -30,6 +31,8 @@ function assertNever(value: never): never {
  * to provide a rich management API for any folders and files operations
  */
 export class Storage {
+  private readonly cdnCache = new CdnCacheManager()
+
   constructor(
     public readonly backend: StorageBackendAdapter,
     public readonly db: Database,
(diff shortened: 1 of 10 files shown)


src/storage/storage.ts
```diff
@@ -30,6 +31,8 @@ function assertNever(value: never): never {
  * to provide a rich management API for any folders and files operations
  */
 export class Storage {
+  private readonly cdnCache = new CdnCacheManager()
```
ferhatelmas: make it module level, it doesn't have state

src/storage/storage.ts
```diff
@@ -232,15 +235,20 @@ export class Storage {
     }
     bucketData.allowed_mime_types = data.allowedMimeTypes
 
-    return this.db.updateBucket(id, bucketData)
+    const result = await this.db.updateBucket(id, bucketData)
+
+    // purge cache if a bucket is changing from public to private
+    if (data.public === false && result?.previous.public === true) {
+      await this.purgeBucketCache(id)
```
ferhatelmas, line 239: not sure but it might be better to make this an event so that we can retry, dropping it won't be nice

src/test/bucket.test.ts
```diff
@@ -631,6 +688,39 @@ describe('testing DELETE bucket', () => {
...
+      })
+      deleted = true
+    } finally {
+      if (!deleted) {
+        await cleanupBucket(bucketId)
+      }
+    }
+  })
+
+  test('bucket deletion succeeds even if purging the CDN cache fails', async () => {
```
ferhatelmas, line 759: maybe for PUT as well

src/storage/storage.ts
```diff
@@ -259,6 +267,32 @@ export class Storage {
...
+
+    return deleted
+  }
+
+  /**
+   * Purges the CDN cache for a bucket, tolerating failures since this is a
+   * best-effort side effect that must not block the underlying DB operation.
+   * @param bucketId
+   */
+  private async purgeBucketCache(bucketId: string) {
```
ferhatelmas, line 273: if cdn url is missing, this will be logging errors, might be better to return early

ferhatelmas (approved):

PR #1276, opened 2026-07-28
fix(docs): normalize nullable JSON Schema idioms for OpenAPI 3.0 docs

grdsdev:
## Summary
- The generated OpenAPI docs declare `"openapi": "3.0.3"`, but route schemas are authored as plain JSON Schema (`type: [X, 'null']`, or `{ anyOf: [<schema>, { type: 'null' }] }` for `$ref`-based unions). Both idioms are valid JSON Schema — AJV validates them natively — but OpenAPI 3.0's Schema Object can't represent either: it only allows a single `type` string plus a separate `nullable: true` boolean.
- This mismatch breaks strict OpenAPI 3.0 consumers (e.g. Swift's `OpenAPIKit30`-based codegen used by supabase-swift), which fail to decode the document entirely.
- Adds `src/http/openapi-nullable.ts`, a docs-only swagger `transform` (composed with the existing `finiteSwaggerTransform`) that rewrites both idioms into their OpenAPI 3.0 equivalent before the doc is emitted. Runtime request/response validation schemas are completely untouched — same clone-before-mutate approach as the existing `stripFiniteKeyword`.
- Considered instead rewriting the schema files directly to `nullable: true`, but `nullable` isn't a real JSON Schema keyword — AJV would ignore it and reject legitimate `null` values at runtime. The transform-only approach has no runtime risk and touches no schema files.

## Test plan
- [x] `npx vitest run src/http/openapi-nullable.test.ts src/http/finite.test.ts` — new + existing tests pass
- [x] `npx tsc -noEmit` — clean
- [x] `npx @biomejs/biome check` on changed files — clean

### DIFF
--- a/src/admin-app.ts
+++ b/src/admin-app.ts
@@ -6,7 +6,8 @@ import { getGlobal } from '@platformatic/globals'
 import fastify, { FastifyInstance, FastifyServerOptions } from 'fastify'
 import { getConfig } from './config'
 import { plugins, routes, setErrorHandler } from './http'
-import { finiteSwaggerTransform, withFiniteAjv } from './http/finite'
+import { withFiniteAjv } from './http/finite'
+import { docsSwaggerTransform } from './http/openapi-nullable'
 
 interface buildOpts extends FastifyServerOptions {
   exposeDocs?: boolean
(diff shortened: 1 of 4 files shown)


grdsdev:
Closing in favor of a different approach — see #1276-replacement (bumping declared openapi version to 3.1.0 instead of transforming schemas at the docs layer). Will link the new PR here.

PR #1277, opened 2026-07-28
fix: validate generated OpenAPI spec against OAS 3.1

grdsdev:
## Summary
- Declare OpenAPI `3.1.0` (up from `3.0.x`) in `src/app.ts` / `src/admin-app.ts`, since 3.1's Schema Object is a full JSON Schema 2020-12 superset — needed because route schemas use `type: [X, 'null']` instead of the OAS 3.0 `nullable: true` idiom. (Originally targeted 3.2.0; downgraded to 3.1.0 since most codegen tools don't support 3.2 yet — 3.1 already covers what we need.)
- Convert route schemas from the OAS 3.0 `nullable: true` idiom to standard JSON Schema `type: [X, 'null']`.
- Add `src/scripts/validate-openapi.ts` (`npm run docs:validate`), which validates both generated specs against the official OAS 3.1 meta-schema (`@hyperjump/json-schema/openapi-3-1`), including embedded Schema Objects.
- Add a strict-dialect pass in the same script (`unevaluatedProperties: false` layered on the official OAS 3.1 dialect) that walks every embedded Schema Object and rejects unknown keywords — catches a regression back to the OAS 3.0 `nullable` idiom, which the official dialect alone permits silently.
- Wire `docs:validate` into CI (`.github/workflows/ci.yml`), with `VECTOR_ENABLED`/`ICEBERG_ENABLED` set so those routes' schemas are included.

## Test plan
- [x] `npm run test:unit` passes (public/admin app suites green).
- [x] `npm run build` / `tsc -noEmit` clean.
- [x] `npm run docs:validate` passes against the current generated specs.
- [x] Verified the strict pass actually catches regressions: reverted a schema to the old `nullable: true` idiom locally, confirmed `docs:validate` failed with the exact offending path, then reverted the test change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,8 @@ jobs:
           cp .env.test.sample .env.test
       - name: Unit tests pass
         run: npm run test:unit:coverage
+      - name: Validate generated OpenAPI spec
+        run: npm run docs:validate
       - name: Upload coverage results to Coveralls
         uses: coverallsapp/github-action@5cbfd81b66ca5d10c19b062c04de0199c215fb6e # v2.3.7
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,9 @@ jobs:
         run: npm run test:unit:coverage
       - name: Validate generated OpenAPI spec
         run: npm run docs:validate
+        env:
+          VECTOR_ENABLED: true
+          ICEBERG_ENABLED: true
       - name: Upload coverage results to Coveralls
         uses: coverallsapp/github-action@5cbfd81b66ca5d10c19b062c04de0199c215fb6e # v2.3.7
         with:
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "migrations:types": "tsx src/scripts/migrations-types.ts",
     "docs:export": "tsx ./src/scripts/export-docs.ts",
     "docs:export-ui": "tsx ./src/scripts/export-docs-ui.ts",
+    "docs:validate": "tsx ./src/scripts/validate-openapi.ts",
     "pprof:capture": "tsx src/scripts/pprof-client.ts",
     "test:dummy-data": "tsx -r dotenv/config ./src/test/db/import-dummy-data.ts",
     "test:unit": "vitest run --config vitest.unit.config.ts",
@@ -117,6 +118,7 @@
   "devDependencies": {
     "@aws-sdk/s3-presigned-post": "^3.1023.0",
     "@biomejs/biome": "2.5.1",
+    "@hyperjump/json-schema": "^1.17.7",
     "@platformatic/runtime": "^3.63.0",
     "@types/js-yaml": "^4.0.5",
     "@types/json-bigint": "^1.0.4",
--- a/src/admin-app.test.ts
+++ b/src/admin-app.test.ts
@@ -258,14 +258,14 @@ describe('admin app', () => {
     }
   })
 
-  it('declares OpenAPI 3.2.0, matching the JSON Schema nullable idioms route schemas use', async () => {
+  it('declares OpenAPI 3.1.0, matching the JSON Schema nullable idioms route schemas use', async () => {
     getGlobal.mockReturnValue(undefined)
 
     const app = await buildAdminApp({ exposeDocs: true })
 
     try {
       const spec = app.swagger() as { openapi: string }
-      expect(spec.openapi).toBe('3.2.0')
+      expect(spec.openapi).toBe('3.1.0')
     } finally {
       await app.close()
     }
--- a/src/admin-app.test.ts
+++ b/src/admin-app.test.ts
@@ -257,4 +257,17 @@ describe('admin app', () => {
       await app.close()
     }
   })
+
+  it('declares OpenAPI 3.2.0, matching the JSON Schema nullable idioms route schemas use', async () => {
+    getGlobal.mockReturnValue(undefined)
+
+    const app = await buildAdminApp({ exposeDocs: true })
+
+    try {
+      const spec = app.swagger() as { openapi: string }
+      expect(spec.openapi).toBe('3.2.0')
+    } finally {
+      await app.close()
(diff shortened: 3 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -43,6 +43,8 @@ jobs:
           cp .env.test.sample .env.test
       - name: Unit tests pass
         run: npm run test:unit:coverage
+      - name: Validate generated OpenAPI spec
+        run: npm run docs:validate
```
ferhatelmas, line 47: this only reads `.env` through config but .env.sample doesn't enable vector/iceberg so we need to pass them explicitly

```
env:
    VECTOR_ENABLED: true
    ICEBERG_ENABLED: true
```

src/scripts/validate-openapi.ts
```diff
@@ -0,0 +1,49 @@
...
+  const specs = [
+    { name: 'public API', instance: buildApp({ exposeDocs: true }) },
+    { name: 'admin API', instance: buildAdmin({ exposeDocs: true }) },
+  ]
+
+  let hasErrors = false
+
+  for (const { name, instance } of specs) {
+    const spec = await getSpec(instance)
+    const result = validateOpenApi(spec, BASIC)
```
ferhatelmas: This validation seems permissive for unknown keywords so if there is `nullable` regression, it won't catch it

package.json
```diff
@@ -117,6 +118,7 @@
   "devDependencies": {
     "@aws-sdk/s3-presigned-post": "^3.1023.0",
     "@biomejs/biome": "2.5.1",
+    "@hyperjump/json-schema": "^1.17.7",
```
ferhatelmas, line 121: I was thinking we're using ajv, can't we do the validation without a new dependency?

.github/workflows/ci.yml
grdsdev: Fixed - 28b1f43

package.json
grdsdev: Claude helped me to come up with an answer for this, see below

> Looked into it: ajv (v8) already supports the 2020-12 dialect out of the box, so that's not the blocker. The issue is the OAS meta-schema itself (schema-base + dialect + vocab annotations) is a multi-file bundle hosted at spec.openapis.org, pulled in via `$ref`/`$dynamicRef`. Ajv has no built-in fetching/registration for that bundle, so we'd need to hand-roll a loader that fetches and `addSchema`s every referenced file, plus set `strict: false` for the `$vocabulary` entries ajv doesn't recognize.
> 
> `@hyperjump/json-schema` handles all of that already and is purpose-built for validating against OpenAPI's published meta-schemas. It's also a devDependency only (CI-time, no runtime footprint), so I'd rather keep it than reimplement that plumbing with ajv. Also downgrading the spec target to 3.1.0 in a follow-up commit since most generators don't support 3.2 yet, doesn't change this tradeoff either way.

src/scripts/validate-openapi.ts
grdsdev: Good catch, confirmed: reverted one schema to the old `nullable: true` idiom and this still reported "valid" — the OAS dialect permits unknown keywords, so a stray `nullable` is just silently ignored.

Fixed in 1f665a75: added a strict dialect overlay (`unevaluatedProperties: false` on top of the official OAS 3.1 dialect, via its `$dynamicAnchor: "meta"` so it applies recursively at every nesting level), and run it against every embedded Schema Object in the generated spec (`components.schemas` values + anything under a `schema` key). Re-tested with the same `nullable: true` regression — now fails with the exact offending path pointed out.

src/admin-app.test.ts
```diff
@@ -257,4 +257,17 @@ describe('admin app', () => {
...
   })
+
+  it('declares OpenAPI 3.1.0, matching the JSON Schema nullable idioms route schemas use', async () => {
+    getGlobal.mockReturnValue(undefined)
+
+    const app = await buildAdminApp({ exposeDocs: true })
+
+    try {
+      const spec = app.swagger() as { openapi: string }
+      expect(spec.openapi).toBe('3.1.0')
```
ferhatelmas, line 268: `expect(JSON.stringify(spec)).not.toContain('"nullable"')`

can we add this or similar that it doesn't regress?

src/app.test.ts
```diff
@@ -38,4 +38,17 @@ describe('public app', () => {
...
   })
+
+  it('declares OpenAPI 3.1.0, matching the JSON Schema nullable idioms route schemas use', async () => {
+    const app = buildApp({ exposeDocs: true })
+
+    try {
+      await app.ready()
+
+      const spec = app.swagger() as { openapi: string }
+      expect(spec.openapi).toBe('3.1.0')
```
ferhatelmas, line 49: similarly here

src/scripts/validate-openapi.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  if (typeof node !== 'object' || node === null) {
+    return
+  }
+
+  if (parentKey === 'schema') {
+    out.push({ path, schema: node })
+    return
+  }
```
ferhatelmas, line 74: Is this going to handle a mere field named `schema`? 
we don't have at the moment but I guess it could be useful for iceberg at some point

src/http/schemas/file-size-limit.ts
```diff
@@ -1,11 +1,10 @@
 export const fileSizeLimitSchema = {
   anyOf: [
-    { type: 'integer', finite: true, examples: [1000], nullable: true, minimum: 0 },
+    { type: ['integer', 'null'], finite: true, examples: [1000], minimum: 0 },
     {
-      type: 'string',
+      type: ['string', 'null'],
       pattern: '^[0-9]+(?:\\.[0-9]+)?(?:[gG][bB]|[mM][bB]|[kK][bB]|[bB])$',
       examples: ['100MB'],
-      nullable: true,
```
ferhatelmas, line 8: seems one is left behind https://github.com/supabase/storage/blob/47048ef9eb3f2385975dc5657872f24c10d0549b/src/http/finite.test.ts#L52

PR #1279, opened 2026-07-29
fix: use common cache for pgvector metrics

ferhatelmas:

Bug fix

pgvector metric cache is using a TTL cache but we're trying to drop it. 
It caches lookup failures at the moment.

## What is the new behavior?

Use common LRU cache wrapper with the same TTL, we can drop age update separately for all caches.
Don't cache metric lookup failure but still fallback to `cosine`.
Fix cross tenant contamination by scoping db identity.

## Additional context

This enables to drop TTL dependency after pool cache is refactored.


ferhatelmas:
@claude review always

fenos (approved):

ferhatelmas:
@claude review always

PR #1293, opened 2026-08-03
fix: hardening for non-json object reply

ferhatelmas:

Bug fix

Non-JSON content-type payload isn't serialized and causes uncaught exception. 

## What is the new behavior?

Convert blob payload to web streams where applicable.
Make db connection cleanup hooks sync, any escape will be caught by fastify and it's faster as well by not allocating one promise and doing one less microtask execution.

## Additional context

Bump fastify due to https://github.com/fastify/fastify/pull/6889


fenos (approved):

ferhatelmas:
@claude review

PR #1294, opened 2026-08-03
fix(tus): fail closed on incomplete uploads without discarding safe retries

ferhatelmas:

Bug fix

Upstream gracefully ends the proxy stream on request abort so the datastore sees a clean EOF and a truncated PATCH resolves a successful partial write with a phantom offset.

## What is the new behavior?

Incoming message and internal cancellation signal are carried into the datastore through async local storage. After write resolves, bytes written are compared to content length, or abort/complete/cancel for chunked bodies, and accordingly upload is destroyed than resumed to prevent silent corruption.

Mutation is tracked to stay resumable if errors came before durable store changes.
Also, harden remove for genuine upstream gaps to delete part, clear cache, handle per key delete errors.

## Additional context

* Upload-Checksum is silently ignored as before.
* Ultimate solution is in upstream, which I will follow up, and checksum extension will provide the missing guarantee even if byte arithmetric holds.

### DIFF
--- a/src/http/routes/tus/index.ts
+++ b/src/http/routes/tus/index.ts
@@ -11,7 +11,9 @@ import {
   FileStore,
   LockNotifier,
   PgLocker,
+  runWithTusRequest,
   S3Store,
+  setTusRequestCancellationSignal,
   UploadId,
 } from '@storage/protocols/tus'
 import { S3Locker } from '@storage/protocols/tus/s3-locker'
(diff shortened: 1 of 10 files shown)


ferhatelmas:
@claude review always

itslenny (approved):

ferhatelmas:
closing in favor of adding checksum support on upstream

PR #1296, opened 2026-08-03
refactor: make normalizeColumns table dependent

TylerHillery:

Refactor to make `normalizeColumns()` table dependent and consolidate all new column checking to use this function so we can have all rules defined in one spot. 

Currently all rules apply to all callers of `normalizeColumns` even though the new columns added are tied to certain tables. This has caused other normalize column like behavior elsewhere for specific functions.

## What is the new behavior?

- Made a `COLUMN_MIGRATION_RULES` object that defines all rules per table. Callers of `normalizeColumns()` now must pass in one of the keys of this object which corresponds to the table name.
- Old loop returned as soon as it hit the first rule whose migration hadn't run, so only
  one rule ever actually got applied. New version collects every pending rule for the table
  and excludes all their columns. This matters now since `s3_multipart_uploads` has two rules
  (`custom-metadata` -> `user_metadata`, `s3-multipart-uploads-metadata` -> `metadata`).
- Folded buckets' type filtering and s3_multipart_uploads' metadata filtering into the same
  rules table too, since they were doing the exact same thing by hand elsewhere. Deleted
  `hasMultipartMetadataColumn()` and `normalizeMultipartUploadColumns()` since they're not needed
  anymore.
- Made `normalizeColumns()` overloaded instead of generic so we could drop the as T casts.

### DIFF
--- a/src/storage/database/pg.ts
+++ b/src/storage/database/pg.ts
@@ -57,6 +57,20 @@ interface UnscopedQueryOptions {
   readonly timeoutMs?: number
 }
 
+const COLUMN_MIGRATION_RULES = {
+  objects: [{ migration: 'custom-metadata', newColumns: ['user_metadata'] }],
+  s3_multipart_uploads: [
+    { migration: 'custom-metadata', newColumns: ['user_metadata'] },
+    { migration: 's3-multipart-uploads-metadata', newColumns: ['metadata'] },
+  ],
+  buckets: [{ migration: 'iceberg-catalog-flag-on-buckets', newColumns: ['type'] }],
+} as const satisfies Record<
+  string,
+  ReadonlyArray<{ migration: keyof typeof DBMigration; newColumns: readonly string[] }>
+>
+
+export type ColumnRuleTable = keyof typeof COLUMN_MIGRATION_RULES
+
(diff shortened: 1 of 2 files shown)


src/storage/database/pg.ts
```diff
@@ -1736,78 +1736,50 @@ export class StoragePgDB implements Database {
...
+  ): T
+  protected normalizeColumns(
+    table: ColumnRuleTable,
+    columns: string | Record<string, unknown>
+  ): string | Record<string, unknown> {
     const latestMigration = this.latestMigration
 
     if (!latestMigration) {
       return columns
     }
```
TylerHillery, line 1753: One thing I noticed is if `lastestMigration` is `undefined` we just assume migrations have ran versus `hasMigrations()` specifically calls `tenantHasMigrations` which calls `getTenantConfig()` to get the tenant's latest migration. Should we be doing that here? I left as is for now.

```typescript
  async hasMigration(migration: keyof typeof DBMigration): Promise<boolean> {
    if (this.latestMigration !== undefined) {
      return DBMigration[this.latestMigration] >= DBMigration[migration]
    }

    return tenantHasMigrations(this.tenantId, migration)
  }
````

TylerHillery:
closing in favor of #1238

PR #1297, opened 2026-08-04
fix: mark storage.filename as IMMUTABLE

joaompneves:

Bug fix

`storage.filename(text)` is pure path parsing but was never marked `IMMUTABLE`, so PL/pgSQL leaves it `VOLATILE`. `plpgsql_check` / `supabase db lint` then warn when STABLE functions call it:

> routine is marked as STABLE, but expression is VOLATILE

`storage.foldername` and `storage.extension` were already marked `IMMUTABLE` in `0036` / `0060`; `filename` was left behind.

## What is the new behavior?

Tenant migration `0061` recreates `storage.filename` as `IMMUTABLE`, matching the sibling helpers.

## Additional context

Same body as today — only the volatility label changes.


ferhatelmas:
@joaompneves this looks good, but could you rebase and run `npm run migrations:types` and add its output as well? Added a test in #1299, it will catch the missing one if you run `npm run test:unit`.

ferhatelmas (approved):

PR #1301, opened 2026-08-05
fix: cleanup empty folder failure from concurrent upload in file driver

ferhatelmas:

Bug fix

There is a TOCTOU race while cleaning up empty directories in file driver.
Between check and deletion, a concurrent write can write a new version in the folder and error.

## What is the new behavior?

Use an atomic delete which does nothing if folder isn't empty.


ferhatelmas:
@claude review always

fenos (approved):

PR #1303, opened 2026-08-06
fix: cache crypto key

ferhatelmas:

Bug fix

Every verifyJWT gives a fresh key material to jose, causing web crypto to import the same HMAC or asymmetric key repeatedly.

## What is the new behavior?

Cache string/HMAC secrets in a LRU keyed by algorithm and secret
JWKs use a weak map keyed by JWK object identity and algorithm, tied to the lifetime of cached tenant/JWKS configuration.

## Additional context

This improves verify side, sign is a follow up.


ferhatelmas:
@claude review always

fenos (approved):

PR #1304, opened 2026-08-07
fix: add projection cache

ferhatelmas:

refactor for performance

Projection columns are split, parsed, quoted, etc. many times.

## What is the new behavior?

Compile projection into two level map (migration and then projection) so that work is done once and hot path does zero allocation.

## Additional context

It's up to 60 time faster with zero allocation (~800 byte/req vs 0).
It's similar direction with #1238 while a bit slower (due to map lookups vs array index) but simpler to understand and maintain.

### DIFF
--- a/src/storage/database/pg.test.ts
+++ b/src/storage/database/pg.test.ts
@@ -257,9 +257,9 @@ describe('StoragePgDB testPermission', () => {
 
 describe('StoragePgDB column selection', () => {
   test.each([
-    ['operation-function', 'SELECT "id", "name"'],
-    ['iceberg-catalog-flag-on-buckets', 'SELECT "id", "type", "name"'],
-  ] as const)('uses the precomputed bucket policy for recognized migration %s', async (latestMigration, expectedSql) => {
+    ['operation-function', /SELECT "id", "name"\s+FROM/],
+    ['iceberg-catalog-flag-on-buckets', /SELECT "id", "type", "name"\s+FROM/],
+  ] as const)('uses the precomputed bucket policy for recognized migration %s', async (latestMigration, expectedSqlPattern) => {
     const { storage, transaction } = createQueryCaptureStorage(latestMigration)
     const hasMigration = vi.spyOn(storage, 'hasMigration')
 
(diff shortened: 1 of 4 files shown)


ferhatelmas:
@claude review always

src/storage/database/pg.ts
```diff
@@ -111,6 +117,34 @@ export class StoragePgDB implements Database {
...
     this.latestMigration = options.latestMigration
+
+    const migrationOrdinal = this.latestMigration ? DBMigration[this.latestMigration] : undefined
+    this.supportsCustomMetadataColumns =
+      !this.latestMigration ||
+      (migrationOrdinal !== undefined && migrationOrdinal >= DBMigration['custom-metadata'])
+    this.supportsMultipartMetadataColumn =
+      !this.latestMigration ||
+      (migrationOrdinal !== undefined &&
+        migrationOrdinal >= DBMigration['s3-multipart-uploads-metadata'])
```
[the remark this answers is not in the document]
ferhatelmas: this is matching master behavior

src/storage/database/pg.ts
```diff
@@ -1736,76 +1774,25 @@ export class StoragePgDB implements Database {
...
       return columns
     }
 
-    const rules = [{ migration: 'custom-metadata', newColumns: ['user_metadata'] }]
-
-    for (const rule of rules) {
-      if (DBMigration[latestMigration] >= DBMigration[rule.migration as keyof typeof DBMigration]) {
+    const normalizedColumns: Record<string, unknown> = {}
+    for (const column in columns) {
+      if (!Object.prototype.hasOwnProperty.call(columns, column) || column === 'user_metadata') {
```
itslenny: This could use `Object.hasOwn()` instead

src/storage/database/pg.test.ts
```diff
@@ -212,6 +255,105 @@ describe('StoragePgDB testPermission', () => {
...
+    })
+  })
+
+  test('keeps all requested object columns when their migrations are available', async () => {
+    const { storage, transaction } = createQueryCaptureStorage()
+
+    await storage.findObject('bucket', 'object', 'id,user_metadata,metadata')
+
+    expect(transaction.query.mock.calls[0]?.[0]).toMatchObject({
+      text: expect.stringContaining('SELECT "id", "user_metadata", "metadata"'),
```
itslenny: The query string could contain additional keys. We should check that it contains JUST these keys.

Maybe something like:
```
text: expect.stringMatching(/SELECT "id", "user_metadata", "metadata"\s+FROM/)
```

This applies to all of these query string checks

itslenny (approved):

PR #1311, opened 2026-08-10
fix(s3): ignore non-SigV4 Authorization header on presigned requests

### LINKED ISSUE #48880: Self-hosted Storage S3 presigned URLs break with "InvalidSignature: Unsupported authorization type" when a non-empty Authorization header is present
# Bug report

- [x] I confirm I have searched the Docs, GitHub Discussions, and Discord.

## Describe the bug

The reference `kong.yml` (`docker/volumes/api/kong.yml`) ships a `storage-v1` route whose `post-function` plugin only strips the `Authorization` header when it is **empty**. If a client sends a **non-empty** `Authorization` header on a request that is otherwise authenticated via an S3 **presigned URL** (`X-Amz-Credential`/`X-Amz-Signature` query params), that header is passed straight through to Storage.

Storage's own signature extraction (`extractSignature` in `src/http/plugins/signature-v4.ts`, `supabase/storage`) gives an `Authorization` header **absolute priority** over query-string signing whenever one is present, regardless of whether valid `X-Amz-*` query params also exist:

```ts
async function extractSignature(req: AWSRequest): Promise<ClientSignature> {
  if (typeof req.headers.authorization === 'string') {
    return SignatureV4.parseAuthorizationHeader(req.headers)
  }
  if (typeof req.query['X-Amz-Credential'] === 'string') {
    return SignatureV4.parseQuerySignature(req.query)
  }
  ...
}
```

And `parseAuthorizationHeader` (`src/storage/protocols/s3/signature-v4.ts:110-119`) throws `InvalidSignature: "Unsupported authorization type"` if the header's first token isn't `AWS4-HMAC-SHA256`:

```ts
static parseAuthorizationHeader(headers: SignatureHeaders) {
  const clientSignature = headers.authorization
  if (typeof clientSignature !== 'string') {
    throw ERRORS.InvalidSignature('Missing authorization header')
  }
  const parts = clientSignature.split(' ')
  if (parts[0] !== 'AWS4-HMAC-SHA256') {
    throw ERRORS.InvalidSignature('Unsupported authorization type')
  }
  ...
}
```

`kong.yml`'s `storage-v1` route is clearly aware of this and tries to compensate — its own comment says *"When no Authorization or apikey header is present (S3 presigned URLs), the Lua expression evaluates to nil which Kong renders as empty string. The post-function strips this empty header so Storage's S3 signature verification falls through to query-parameter parsing."* But the actual check only covers the **empty** case:

```lua
local auth = kong.request.get_header("authorization")
if auth == nil or auth == "" or auth:find("^%s*$") then
  kong.service.request.clear_header("authorization")
end
```

It does nothing if the incoming request already carries a **non-empty** `Authorization` header — which the `request-transformer` plugin right above it explicitly passes through unchanged ("*passes through existing Authorization headers*"). So any non-empty, non-SigV4 `Authorization` header on a presigned-URL request reaches Storage and breaks it, even though the URL's own signature is valid.

In practice this breaks presigned-URL downloads/streaming (e.g. serving a large file straight from Storage via a `307` redirect to a presigned URL, bypassing the app server) whenever *anything* attaches a non-empty `Authorization` header to that outgoing request.

## To Reproduce

Minimal, offline reproduction — no real bucket data needed, since this only exercises the auth path (it 404s harmlessly on the object lookup once past the signature check):

1. Generate any valid S3-protocol presigned GET URL for an existing object:
   ```python
   import boto3
   from botocore.config import Config
   client = boto3.client(
       "s3",
       endpoint_url="https://<your-public-domain>/storage/v1/s3",
       aws_access_key_id="<S3_PROTOCOL_ACCESS_KEY_ID>",
       aws_secret_access_key="<S3_PROTOCOL_ACCESS_KEY_SECRET>",
       region_name="<region>",
       config=Config(signature_version="s3v4", s3={"addressing_style": "path"}),
   )
   url = client.generate_presigned_url(
       "get_object", Params={"Bucket": "<bucket>", "Key": "<existing-key>"}, ExpiresIn=3600
   )
   ```
2. Confirm it works normally:
   ```bash
   curl -i "$url"
   # -> 200/206, or a normal S3Error if the key doesn't exist — either way, signature verification passes
   ```
3. Repeat the exact same request, adding any non-empty `Authorization` header:
   ```bash
   curl -i -H "Authorization: Bearer not-a-sigv4-value" "$url"
   ```
4. See error: `400 Bad Request`, `{"code":"InvalidSignature","message":"Unsupported authorization type"}` — even though the query-string signature is untouched and still valid.

## Expected behavior

The presigned URL should still verify successfully; an incidental client-supplied `Authorization` header should not override a valid query-string signature — or, at minimum, Kong's own documented "empty-header" workaround (see comment in `kong.yml`) should also cover the non-empty case, since it explicitly exists to make presigned URLs work through this exact route.

## Screenshots

N/A (server-side/HTTP-level bug, see curl repro above).

## System information

- OS: Debian 13 (server), self-hosted Docker deployment
- Browser (where first observed): Firefox, but the root cause is server-side and reproducible with plain `curl`, independent of any browser
- `kong/kong:3.9.1`
- Self-hosted `supabase/storage` (S3 protocol enabled, `STORAGE_BACKEND=s3`)
- Reference `docker/volumes/api/kong.yml` from this repo (`storage-v1` route, unmodified `request-transformer` + `post-function` plugins)

## Additional context

Two independent places this could be fixed, either is sufficient on its own, but doing both would be more robust:

**A. `kong.yml` (`storage-v1` route `post-function`)**  strip `Authorization` whenever the request is presigned (has `X-Amz-Credential` in the query string), not just when it's already empty:

```diff
       - name: post-function
         config:
           access:
             - |
               local auth = kong.request.get_header("authorization")
-              if auth == nil or auth == "" or auth:find("^%s*$") then
+              local is_presigned = kong.request.get_query_arg("X-Amz-Credential") ~= nil
+              if is_presigned or auth == nil or auth == "" or auth:find("^%s*$") then
                 kong.service.request.clear_header("authorization")
               end
```

I reproduced the failure and verified this exact fix locally (isolated test, no production data involved): with the fix, the same spoofed `Authorization` header no longer breaks the request — it gets stripped and the request falls through to normal query-signature verification, matching the no-header baseline.

**B. `supabase/storage` (`extractSignature`)** — prefer the query-string signature when `X-Amz-Credential` is present, instead of unconditionally prioritizing the `Authorization` header:

```ts
async function extractSignature(req: AWSRequest): Promise<ClientSignature> {
  if (typeof req.query['X-Amz-Credential'] === 'string') {
    return SignatureV4.parseQuerySignature(req.query)
  }
  if (typeof req.headers.authorization === 'string') {
    return SignatureV4.parseAuthorizationHeader(req.headers)
  }
  ...
}
```

This would fix it at the source for every self-hosted deployment, regardless of what's in front of Storage (Kong, another proxy, or Storage exposed directly).

I can open a PR for (A) against this repo if that's welcome - it's a one-line, backward-compatible change to the reference `kong.yml`.

aantti:

Bug fix

`extractSignature` treated *any* `Authorization` header as an S3 signature and gave it priority over the query string. A presigned URL request that also carried an unrelated `Authorization` header (e.g. a user JWT added by a proxy or the client) therefore failed with `InvalidSignature: "Unsupported authorization type"`.

## What is the new behavior?

Only take the header path when the header is actually SigV4 (`AWS4-HMAC-SHA256`); otherwise fall through to the query-string signature. Genuine SigV4 header auth still takes priority, so this is a narrow change - the two mechanisms are alternatives and no legitimate presigned request needs a non-SigV4 header to win.

## Tests

Added regression tests in the `S3 Presigned URL` suite:
- presigned GET with an incidental `Authorization` header → still succeeds
- presigned upload with an incidental `Authorization` header → still succeeds
- non-SigV4 `Authorization` header with no query signature → still rejected (`403 AccessDenied`)

## Additional context

Fixes supabase/supabase#48880

(This covers the S3 router. The S3 Vectors `skipIfJwtToken` path has the same pattern, not analyzed.)

### DIFF
--- a/src/http/plugins/signature-v4.ts
+++ b/src/http/plugins/signature-v4.ts
@@ -205,7 +205,12 @@ async function authorizeRequestSignV4(
 }
 
 async function extractSignature(req: AWSRequest): Promise<ClientSignature> {
-  if (typeof req.headers.authorization === 'string') {
+  const authorization = req.headers.authorization
+
+  // Only treat the Authorization header as an S3 signature when it actually
+  // carries one. A presigned request can pick up an unrelated Authorization
+  // header which must not override a valid query-string signature.
+  if (typeof authorization === 'string' && authorization.split(' ')[0] === 'AWS4-HMAC-SHA256') {
     return SignatureV4.parseAuthorizationHeader(req.headers)
   }
 
--- a/src/http/plugins/signature-v4.ts
+++ b/src/http/plugins/signature-v4.ts
@@ -210,7 +210,7 @@ async function extractSignature(req: AWSRequest): Promise<ClientSignature> {
   // Only treat the Authorization header as an S3 signature when it actually
   // carries one. A presigned request can pick up an unrelated Authorization
   // header which must not override a valid query-string signature.
-  if (typeof authorization === 'string' && authorization.split(' ')[0] === 'AWS4-HMAC-SHA256') {
+  if (typeof authorization === 'string' && authorization.startsWith('AWS4-HMAC-SHA256 ')) {
     return SignatureV4.parseAuthorizationHeader(req.headers)
   }
 
--- a/src/test/s3-protocol.test.ts
+++ b/src/test/s3-protocol.test.ts
@@ -3460,6 +3460,23 @@ describe('S3 Protocol', () => {
         expect(resp.ok).toBeTruthy()
       })
 
+      it('rejects a non-SigV4 Authorization header when there is no query signature', async () => {
+        // Counterpart to the presigned cases above: with no query-string
+        // signature to fall through to, a non-SigV4 Authorization header must
+        // be rejected rather than silently accepted.
+        const bucket = await createBucket(client)
+
+        const resp = await fetch(`${baseUrl}/s3/${bucket}/test-key`, {
+          headers: {
+            Authorization: 'Bearer not-a-sigv4-value',
+          },
+        })
+
+        expect(resp.ok).toBeFalsy()
+        expect(resp.status).toBe(403)
+        expect(await resp.text()).toContain('AccessDenied')
+      })
+
       it('supports response-content-disposition override', async () => {
         const bucket = await createBucket(client)
         const key = 'test-disposition.jpg'
--- a/src/test/s3-protocol.test.ts
+++ b/src/test/s3-protocol.test.ts
@@ -3406,6 +3406,61 @@ describe('S3 Protocol', () => {
         expect(resp.ok).toBeTruthy()
       })
 
+      it('ignores an incidental non-SigV4 Authorization header on a presigned GET', async () => {
+        // Regression for https://github.com/supabase/supabase/issues/48880: a
+        // presigned URL carries its signature in the query string, so an
+        // unrelated Authorization header (e.g. a user JWT forwarded by a proxy)
+        // must not be parsed as an S3 signature and must not override it.
+        const bucket = await createBucket(client)
+        const key = 'test-1.jpg'
+
+        await uploadFile(client, bucket, key, 2)
+
+        const getUrl = await getSignedUrl(
+          client,
+          new GetObjectCommand({
+            Bucket: bucket,
+            Key: key,
+          }),
+          { expiresIn: 100 }
+        )
+
+        const resp = await fetch(getUrl, {
+          headers: {
+            Authorization: 'Bearer not-a-sigv4-value',
+          },
+        })
+
+        expect(resp.ok).toBeTruthy()
+      })
+
+      it('ignores an incidental non-SigV4 Authorization header on a presigned upload', async () => {
+        const bucket = await createBucket(client)
+        const key = 'test-1.jpg'
+        const body = Buffer.alloc(1024 * 2)
+
+        const uploadUrl = await getSignedUrl(
+          client,
+          new PutObjectCommand({
+            Bucket: bucket,
+            Key: key,
+            Body: body,
+          }),
+          { expiresIn: 100 }
+        )
+
+        const resp = await fetch(uploadUrl, {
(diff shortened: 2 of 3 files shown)


src/http/plugins/signature-v4.ts
```diff
@@ -205,7 +205,12 @@ async function authorizeRequestSignV4(
 }
 
 async function extractSignature(req: AWSRequest): Promise<ClientSignature> {
-  if (typeof req.headers.authorization === 'string') {
+  const authorization = req.headers.authorization
+
+  // Only treat the Authorization header as an S3 signature when it actually
+  // carries one. A presigned request can pick up an unrelated Authorization
+  // header which must not override a valid query-string signature.
+  if (typeof authorization === 'string' && authorization.split(' ')[0] === 'AWS4-HMAC-SHA256') {
```
ferhatelmas: `startsWith('AWS4-HMAC-SHA256 ')` would be better not to allocate

src/test/s3-protocol.test.ts
```diff
@@ -3406,6 +3406,60 @@ describe('S3 Protocol', () => {
         expect(resp.ok).toBeTruthy()
       })
 
+      it('ignores an incidental non-SigV4 Authorization header on a presigned GET', async () => {
```
ferhatelmas, line 3409: can we get a negative test (non-SigV4 header + no query signature)?

src/http/plugins/signature-v4.ts
aantti: Changed

src/test/s3-protocol.test.ts
aantti: Added

ferhatelmas (approved):

PR #1313, opened 2026-08-10
fix: handle feature not enabled for s3 xml serialization

ferhatelmas:

Bug fix

S3 routes run under xml parser/serializer and requires a single root.
Feature not enabled short circuit response isn't so xml parser validation check converts 403 to 500. 

## What is the new behavior?

Export s3 error formatter and pass it to the tenant feature plugin as an option to directly finish the request.


ferhatelmas:
@claude review always

itslenny (approved):

PR #1314, opened 2026-08-10
fix: memoize derived jwk algorithms

ferhatelmas:

Bug fix

Algorithms are derived from immutable JWKs array per verify.

## What is the new behavior?

Cache derived array by JWKs array identity.

## Additional context

This is similar to #1303 in its nature but a different knob to improve.

### DIFF
--- a/src/internal/auth/jwt.ts
+++ b/src/internal/auth/jwt.ts
@@ -287,7 +287,7 @@ function getJWTAlgorithms(jwks: JwksConfig | null) {
   return algorithms
 }
 
-function getJWTJwksFingerprint(jwks?: { keys: JwksConfigKey[] } | null): string {
+function getJWTJwksFingerprint(jwks?: JwksConfig | null): string {
   if (!jwks) {
     return 'null'
   }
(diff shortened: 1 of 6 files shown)


src/internal/auth/jwt.ts
```diff
@@ -86,6 +87,7 @@ export function isDownloadScopedToken(payload: { scope?: SignedUrlScope }): bool
 }
 
 const jwtJwksFingerprintCache = new WeakMap<object, string>()
+const jwtAlgorithmsCache = new WeakMap<JwksConfig, string[]>()
```
itslenny, line 90: JwksConfig is mutable. I don't think we mutate it anywhere but nothing protects us from mutating it either. Perhaps we should mark JwksConfig.keys as readonly for compile protection and freeze it when it's created for runtime protection?

itslenny (changes requested):

src/internal/auth/jwt.ts
ferhatelmas: updated, PTAL

itslenny (approved):

PR #1318, opened 2026-08-13
fix: absolute TTL, coarse ttl resolution, and jitter for caches

ferhatelmas:

Bug fix

Caches with long ttl have very low ttl resolution (1h vs 1ms). This creates timer arms with update age on get with hot keys. Additionally, update age on get isn't safe because staleness is unbounded for example from a dropped notification.

## What is the new behavior?

Increase ttl resolution to 30s but make it overridable.
Drop update age on get, there will slightly more backend checks but with single flight, it's protected.
Add jitter so that it doesn't cause refresh storm. 

## Additional context

Related to #1316
After this, tenant pools can be a LRU cache only capped with capacity, no TTL and close cache improvements.

### DIFF
--- a/src/internal/auth/jwks/manager.ts
+++ b/src/internal/auth/jwks/manager.ts
@@ -4,6 +4,7 @@ import {
   type CacheLookupOptions,
   createLruCache,
   DEFAULT_CACHE_PURGE_STALE_INTERVAL_MS,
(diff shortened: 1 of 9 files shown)


ferhatelmas:
@claude review always

fenos (approved):

PR #1319, opened 2026-08-13
chore(ci): setup dependency firewall for npm

staaldraad:

CI hardening.

When publishing, ensure dependencies are pulled in via the dependency firewall.

- [x] `DF_FIREWALL_TOKEN` value set
- [x] Policy in evaluation mode for teting


.github/workflows/release.yml
```diff
@@ -22,6 +22,7 @@ jobs:
         uses: ./.github/actions/setup-node-dependencies
         with:
           cache: "false"
+          firewall-token: ${{ secrets.DF_FIREWALL_TOKEN }}
```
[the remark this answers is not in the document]
staaldraad: the `release` job will fail, meaning `publish` doesn't get called. There is a window between a successful `release` and `publish` that could allow a compromised package to sneak in. The risk / chance of this happening is low

ferhatelmas, line 25: this is for deciding semantic release version but actual dependencies are installed inside docker later in this file, I believe we also want to filter/check them via passing a build secret

ferhatelmas (changes requested):
there is a gap for build dependencies

Dockerfile
ferhatelmas:
```suggestion
# Remove dev dependencies and keep native modules without downloading replacements or running scripts.
RUN --network=none npm prune --omit=dev --offline --ignore-scripts
```

Dockerfile
```diff
@@ -1,14 +1,17 @@
...
 COPY scripts/ensure-npm-version.cjs ./scripts/ensure-npm-version.cjs
-RUN node scripts/ensure-npm-version.cjs
+# The optional `npmrc` build secret lets CI route registry traffic through a
+# different registry layer.
+RUN --mount=type=secret,id=npmrc,target=/root/.npmrc node scripts/ensure-npm-version.cjs
 
 # Dependencies stage - install and cache all dependencies
 FROM base AS dependencies
-RUN npm ci
+RUN --mount=type=secret,id=npmrc,target=/root/.npmrc npm ci
```
ferhatelmas:
```suggestion
RUN --mount=type=secret,id=npmrc,target=/root/.npmrc npm ci --ignore-scripts
# Native addons can use the Node headers already included in the image.
RUN --network=none npm_package_config_node_gyp_nodedir=/usr/local npm rebuild
```

ferhatelmas (approved):

PR #1324, opened 2026-08-14
fix: dispose connection when storage setup fails

ferhatelmas:

Bug fix

If storage construction fails after acquiring a tenant database connection, no Storage instance is returned to the caller. The caller cannot dispose the connection through its normal cleanup path.

## What is the new behavior?

Now, the acquired connection is disposed when database, backend, or storage construction fails, then rethrows the original error.


ferhatelmas:
@claude review

itslenny (approved):

PR #1330, opened 2026-08-14
fix: max capped pools, drop ttl cache

ferhatelmas:

Bug fix + refactor

Tenant pool caches uses TTL-based expiry and updates age on get.

## What is the new behavior?

Tenant pool caches uses LRU capped with max items, and no timers.
Capacity evicted pools are retired until their last lease. Children share parent's lease (asSuperUser). There is 1 hour hard stop.
Explicit teardown forces immediate retirement.
In-flight retirements are tracked and observable, count and age gauges.
Custom logging is dropped since monitored cache wrapper handles hit/miss/eviction metrics.

## Additional context

Pool stat collection interval goes from 5s to 15s, chunk size from 100 to 4096 (experimental sweet spot). 
Will need #1320


ferhatelmas:
@claude review always

src/internal/database/pool.ts
```diff
@@ -92,86 +168,105 @@ function belongsToTenant(cacheKey: string, tenantId: string): boolean {
...
 }
 
-const tenantPools = createTtlCache<string, PoolStrategy>({
-  ...(isMultitenant ? multiTenantTtlConfig : { max: 1, ttl: Infinity }),
-  dispose: async (pool, _tenantId, reason) => {
-    if (!pool || manuallyDestroyedPools.has(pool)) {
+const tenantPools = createLruCache<string, PoolStrategy>(TENANT_POOL_CACHE_NAME, {
+  max: isMultitenant ? tenantPoolCacheMaxEntries : 1,
+  disposeAfter: (pool, cacheKey, reason) => {
+    if (reason !== 'evict') {
```
fenos: I'm wondering if there is a chance here that we are going to leak pools without having them closed properly

src/internal/database/pg-connection.ts
```diff
@@ -94,6 +96,11 @@ const transactionSetupRetryMaxDelayMs = 200
 export class PgPoolStrategy {
   protected pool?: Pool
   protected tlsSession?: TlsSessionSlot
+  private activeLeaseCount = 0
```
fenos: I think it might be better to encapsulate this logic as an abstract class or via a composite class.
This way we can keep the PgPoolStrategy focused on Postgres related functionality and less on the pool state

fenos: activeLeaseCount, retirementState, retirement, retirementWaiter, retirementLeaseTimeout

fenos: Potentially we can also encapsulate together with the above:

DeferredPoolRetirementRegistry,
trackPoolRetirement,
forceAndObservePoolRetirements,
retirePoolWhenReleasedInBackground,
waitForTenantPoolDestroys

Currently they sit in pool.ts, which seems that they are bloating the file with scattered logic across pg-connection and pool.ts

ferhatelmas: addressed

src/internal/database/pool.ts
ferhatelmas: we no longer destroy inline, and we have metrics and failure logs if happens. There were two cases earlier, addressed them separately before this PR

src/internal/database/pool-retirement.ts
```diff
@@ -0,0 +1,270 @@
+export type PoolRetirementReason = 'destroy' | 'evict'
```
fenos: I'm wondering if this classes should be thought as caches and so belonging to the `cache/` package.
so that we can design this in a slitghly different way:

instead of doing:

```ts
this.leaseRetirement = new PoolLeaseRetirement({
      leaseTimeoutMs: poolLeaseTimeoutMs,
      retire: (reason) => this.drainAndDestroyPool(reason),
      onLeaseTimeout: (reason, activeLeases) => this.logLeaseDrainTimeout(reason, activeLeases),
    })
```

We make it so it's a lruCache with least access eviction policy.

```ts
import { lruCache } from '@internal/cache'

const poolCache = lruCache({ policy: 'least-accessed', onDispose: (pool) => {
   pool.drainAndDestroyPool(reason)
}) })
```

This way the actual `pg-pool` doesn't need to know about leases and disposals.
We just add the pool to this lru cache which has the eviction policy least used.

ferhatelmas: this is a good idea and cleaner, updated 🙏🏻

src/internal/database/pg-connection.ts
```diff
@@ -629,31 +642,43 @@ function serializeJwtPayload(payload: object): string {
...
   public readonly role: string
+  public readonly pool: PgPoolStrategy
   private abortSignal?: AbortSignal
   private disposed = false
   private readonly headersPayload: string
   private userPayload?: string
 
   constructor(
-    public readonly pool: PgPoolStrategy,
+    private readonly poolLease: Lease<PgPoolStrategy>,
```
fenos, line 652: Probably is better to accept the `PgPoolStrategy` here directly, then wrap it with the lease in the constructor? 

This way is a bit more open for injection

ferhatelmas: It would make the ref counters in cache error prone. Injection isn't [difficult](https://github.com/supabase/storage/pull/1330/changes#diff-d624316485835ecb052dee8d429d361739ba5ebd47916f9c4c27f295737dcf76R64-R69). I would keep it otherwise we might need another callback for release

fenos: Are we able to accept both? 

```ts
private readonly poolLease: Lease<PgPoolStrategy> | PgPoolStrategy
```

Would it make sense to do it now?

fenos: the only thing is that if we are able to not have the Lease as a first class citizen in the Connection here seems more extensible and generally a more compliant spec

fenos: But i'm ok to keep it like this for now and do another pass if we ever needs to replace the implementation

fenos (approved):

src/internal/database/pg-connection.ts
ferhatelmas: sounds good, if needed, we can rework afterwards

PR #1337, opened 2026-08-18
fix(s3): reject wrong-length signatures instead of throwing

### LINKED ISSUE #1336: S3 requests with a malformed signature return 500 instead of 403
When an S3 request comes in with a signature that isn't 64 hex chars, storage returns 500 InternalError instead of 403 SignatureDoesNotMatch.

`SignatureV4.verify()` compares the client and server signatures with `crypto.timingSafeEqual(Buffer.from(clientSignature.signature), Buffer.from(serverSignature.signature))`. The server signature is always 64 chars, but the client signature is taken straight from the request with no length check. `timingSafeEqual` throws a RangeError when the two buffers differ in length, and that error is not a StorageBackendError, so the s3 error handler falls back to 500. `verifyPostPolicySignature()` has the same shape.

A wrong signature should be a 403, and since 5xx is retryable, AWS SDKs also retry these for no reason.

Repro:

```
curl -X PUT 'http://localhost:5000/s3/mybucket/key' \
  -H 'Authorization: AWS4-HMAC-SHA256 Credential=key/20240101/us-east-1/s3/aws4_request, SignedHeaders=host, Signature=abc' \
  -H 'x-amz-date: 20240101T000000Z'
```

returns 500 InternalError. A same-length but wrong signature correctly returns 403, which is what points at the length mismatch.

The streaming chunk parser already guards this by checking the per-chunk signature against `/^[0-9a-fA-F]{64}$/` before comparing; the header, query, and policy paths don't.

oniani1:
Closes #1336

A signature that isn't 64 chars long makes `verify()` throw instead of returning false, because `crypto.timingSafeEqual` needs equal-length buffers. The RangeError isn't a StorageBackendError, so it surfaces as a 500 rather than a 403.

Guarded the length before the compare in both `verify()` and `verifyPostPolicySignature()`, so a wrong-length signature returns false and the caller responds with 403. Same thing the streaming parser already does with its hex check.

Added a test that a wrong-length signature resolves to false instead of throwing.

### DIFF
--- a/src/storage/protocols/s3/signature-v4-stream.test.ts
+++ b/src/storage/protocols/s3/signature-v4-stream.test.ts
@@ -620,4 +620,37 @@ describe('ChunkSignatureV4Parser', () => {
     ).toBe(true)
     expect(signingKeySpy).toHaveBeenCalledTimes(1)
   })
+
+  test('rejects a wrong-length chunk signature instead of throwing', () => {
+    const secretKey = 'secret-key'
+    const signer = new TestSignatureV4({
+      enforceRegion: false,
+      credentials: {
+        accessKey: 'access-key',
+        secretKey,
+        region: 'us-east-1',
+        service: 's3',
+      },
+    })
+    const clientSignature = {
+      credentials: {
+        accessKey: 'access-key',
+        shortDate: '20260406',
+        region: 'us-east-1',
+        service: 's3',
+      },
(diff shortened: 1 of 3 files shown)


ferhatelmas:
@oniani1 thanks, added missing test gaps, will merge on green

ferhatelmas (approved):

PR #1339, opened 2026-08-19
fix: include search and prefix when computing level for list v1

itslenny:

Bug fix

Slashes in `search` are not included when computing level. This causes `storage.search()` to return the name from the incorrect level when the search param includes slashes

When search params are included the `name` is returned relative to the search param (instead of the prefix)

## What is the new behavior?

Include prefix+search when computing level so it matches the internal behavior of storage.search (`v_prefix := coalesce(prefix, '') || coalesce(search, '')`)

Always return object name relative to the prefix. This maintains the pattern of prefix+name = full object path

Eliminate externally passed level param

## Additional context

This issue/change does not impact `list v2` as it does not include the separate `search` param

### DIFF
--- a/src/storage/database/pg.ts
+++ b/src/storage/database/pg.ts
@@ -1438,7 +1438,7 @@ export class StoragePgDB implements Database {
             safePrefix,
             bucketId,
             options.limit || 100,
-            safePrefix.split('/').length,
+            (safePrefix + safeSearch).split('/').length,
             options.offset || 0,
             safeSearch,
             sortColumn,
--- a/src/storage/database/pg.ts
+++ b/src/storage/database/pg.ts
@@ -1430,38 +1430,21 @@ export class StoragePgDB implements Database {
         ? escapeLike(options.search || '')
         : options.search || ''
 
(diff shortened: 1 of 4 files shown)


itslenny:
@claude review always

src/storage/database/pg.ts
```diff
@@ -1430,21 +1430,38 @@ export class StoragePgDB implements Database {
...
+                safePrefix,
+                bucketId,
+                options.limit || 100,
+                (safePrefix + safeSearch).split('/').length,
+                options.offset || 0,
+                safeSearch,
+                sortColumn,
+                options.sortBy?.order ?? 'asc',
+              ],
+            },
```
ferhatelmas: This won't handle old app + new db, can fail during rolling deploy. 
How about keeping `levels` parameter but ignoring inside the function?
Then we can also drop this condition

itslenny: I actually was going in that direction originally, but I felt weird about the unused level param in the search function. I put it back to make it backward compatible and removed the migration check.

ferhatelmas (approved):

PR #1340, opened 2026-08-19
feat: accept tenantId as SigV4 signing secret for session token auth

fenos:
## What

S3 session token requests can now be signed with the **tenantId as both `accessKeyId` and `secretAccessKey`**. The anon key remains supported as the signing secret for backwards compatibility:

| Scheme | accessKeyId | secretAccessKey | sessionToken |
|---|---|---|---|
| New | tenantId | tenantId | user JWT |
| Legacy (still works) | tenantId | anon key | user JWT |

Since both schemes use the same `accessKeyId`, the server cannot tell from the request which secret the client signed with, so `SignatureV4` now accepts multiple candidate secrets: the canonical request and string-to-sign are computed once, one HMAC is derived per candidate, and each is compared with a length-guarded `timingSafeEqual`. Streaming chunk signatures and POST policies are validated with the secret that matched the seed signature.

As a side effect, the `Missing tenant anon key` rejection is gone: **tenants without an anon key can now use session token auth** by signing with the tenantId secret.

## Security

No change in posture. The tenantId is public, but the anon key is also effectively public (it ships in every frontend bundle) — the SigV4 signature in this flow never proved possession of a secret. The real authentication gate is unchanged: the session token is verified as a JWT against the tenant's JWT secret/JWKS after the signature check. The signature keeps providing request integrity binding and the x-amz-date expiry window.

## Tests

- Protocol unit tests: verify with primary/fallback secret, reject unknown secret, chunk-signature lock-in to the matched secret, POST policy with fallback secret
- Plugin unit tests: session token signed with tenantId and with anon key both pass, unknown secret → 403 `SignatureDoesNotMatch`, multitenant tenant **without** an anon key authenticates via tenantId
- Integration tests (`s3-protocol.test.ts`, real `S3Client`): both secrets authenticate, unknown secret rejected, valid signature with an invalid session token JWT still fails with 403 `AccessDenied`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/storage/protocols/s3/signature-v4.ts
+++ b/src/storage/protocols/s3/signature-v4.ts
@@ -17,7 +17,7 @@ interface SignatureV4Options {
   allowBodyHashing?: boolean
   nonCanonicalForwardedHost?: string
   publicUrl?: URL
-  credentials: Omit<Credentials, 'shortDate'> & { secretKey: string }
+  credentials: Omit<Credentials, 'shortDate'> & { secretKey: string | string[] }
 }
 
 export interface ClientSignature {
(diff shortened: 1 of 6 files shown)


src/storage/protocols/s3/signature-v4.ts
```diff
@@ -397,13 +408,18 @@ export class SignatureV4 {
...
+        selectedRegion,
+        serverCredentials.service,
+        index
+      )
 
-    return { signature: this.hmac(signingKey, stringToSign).toString('hex'), canonicalRequest }
+      return this.hmac(signingKey, stringToSign).toString('hex')
+    })
+
+    return { signature: signatures[0], signatures, canonicalRequest }
```
ferhatelmas: I would only keep `signatures` and it would be more consistent, right now a bit footgun

src/storage/protocols/s3/signature-v4.ts
```diff
@@ -628,6 +653,20 @@ export class SignatureV4 {
...
+    for (const [index, candidate] of serverSignatures.entries()) {
+      const serverSig = Buffer.from(candidate)
+      if (clientSig.length === serverSig.length && crypto.timingSafeEqual(clientSig, serverSig)) {
+        this.matchedSecretIndex = index
+        return true
+      }
+    }
+
+    return false
+  }
```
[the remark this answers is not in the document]
ferhatelmas: invalid

src/storage/protocols/s3/signature-v4.ts
ferhatelmas: made the change in the last commit

ferhatelmas (approved):

PR #1341, opened 2026-08-21
feat: support addressing objects by versionId

TylerHillery:

Feature

Currently we don't allow users to specify version id

## What is the new behavior?

Can now pin get, head, info, sign, copy, move, delete, and deleteObjectets to a specific object version via versionId. Also added additional indexes.

## Additional context

Left out  S3 API changes for now as well. Will be done in a follow up PR.

Only one version can exist per object today, so `versionId` just reselects the same object you'd already get by default but this lays groundwork for real multi-version support later.

### DIFF
--- a/migrations/tenant/0067-objects-null-version-index.sql
+++ b/migrations/tenant/0067-objects-null-version-index.sql
@@ -1,0 +1,4 @@
+-- postgres-migrations disable-transaction
+
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS idx_objects_null_version
+    ON storage.objects (bucket_id, name) WHERE NOT is_versioned;
--- a/src/http/routes/object/getObject.ts
+++ b/src/http/routes/object/getObject.ts
@@ -22,6 +22,7 @@ const getObjectQuerySchema = {
   type: 'object',
   properties: {
     download: { type: 'string', examples: ['filename.jpg', null] },
+    versionId: { type: 'string', examples: ['eaa8bdb5-2e00-4767-b5a9-d2502efe2196'] },
   },
 } as const
 
@@ -34,7 +35,7 @@ type GetObjectRequest = FastifyRequest<getObjectRequestInterface>
 
 async function requestHandler(request: GetObjectRequest, response: FastifyReply) {
   const { bucketName } = request.params
-  const { download } = request.query
+  const { download, versionId } = request.query
   const objectName = request.params['*']
 
   // send the object from s3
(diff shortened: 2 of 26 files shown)


src/storage/database/pg.ts
```diff
@@ -977,17 +978,27 @@ export class StoragePgDB implements Database {
 
     const result = await this.runQuery('UpdateObject', async (db, signal) => {
       const update = buildUpdate(objectData)
+      const conditions = [
+        `bucket_id = $${update.values.length + 1}`,
+        `name = $${update.values.length + 2}`,
+      ]
+      const values = [...update.values, bucketId, name]
+
+      if (currentVersion) {
```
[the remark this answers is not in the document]
TylerHillery: fixed

src/storage/object.ts
```diff
@@ -757,10 +790,16 @@ export class ObjectStorage {
...
-      { ...metadata, url: urlToSign, scope: SIGNED_URL_SCOPE_DOWNLOAD },
+      {
+        ...metadata,
+        url: urlToSign,
+        scope: SIGNED_URL_SCOPE_DOWNLOAD,
+        ...(versionId ? { versionId } : {}),
+      },
       urlSigningKey,
       expiresIn
     )
```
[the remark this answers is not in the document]
TylerHillery: fixed

src/storage/database/pg.ts
```diff
@@ -1170,9 +1189,17 @@ export class StoragePgDB implements Database {
...
-    filters?: FindObjectFilters
+    filters?: FindObjectFilters,
+    version?: string
   ) {
     const selectedColumns = selectColumns(columns, this.objectColumnPolicy)
+    const conditions = ['name = $1', 'bucket_id = $2']
+    const values: unknown[] = [objectName, bucketId]
+
+    if (version !== undefined) {
+      values.push(version)
```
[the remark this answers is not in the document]
TylerHillery: fixed

migrations/tenant/0067-objects-null-version-index.sql
```diff
@@ -0,0 +1,4 @@
+-- postgres-migrations disable-transaction
+
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS idx_objects_null_version
```
fenos, line 3: Do we need to drop also existing unique index on `bucket,name`?

fenos (approved):

migrations/tenant/0067-objects-null-version-index.sql
TylerHillery: Will be dropped in the second PR in this stack #1346

src/http/routes/object/getObject.ts
```diff
@@ -22,6 +22,7 @@ const getObjectQuerySchema = {
   type: 'object',
   properties: {
     download: { type: 'string', examples: ['filename.jpg', null] },
+    versionId: { type: 'string', examples: ['eaa8bdb5-2e00-4767-b5a9-d2502efe2196'] },
```
ferhatelmas, line 25: realized we're missing it in the other endpoint https://github.com/supabase/storage/pull/1350

src/storage/database/pg.ts
ferhatelmas, line 1140: this isn't updated for version but it's dead code so fine, I think it's better to drop, will do

ferhatelmas: https://github.com/supabase/storage/pull/1351

Issue #1268, opened 2026-07-24
Local Storage TUS rejects signed upload token generated by the same stack with "Invalid Compact JWS"


tmprrodrigues:
# Summary

`createSignedUploadUrl` generates a signed upload token in a clean local Supabase stack. The token is structurally a valid three-segment compact JWS using HS256, but the initial TUS request is rejected with `Invalid Compact JWS`.

The same result occurs through the local API gateway and directly against the Storage container. The reproducer contains no application code.

The attached results were captured during an earlier isolated diagnostic run. This package-preparation pass did not re-run the matrix.

## Environment

- Windows x64
- Docker 29.6.2
- Bun 1.3.14
- Node 24.18.0
- Supabase CLI 2.109.1
  - storage-api v1.62.5
  - image digest `1dbe962d9862ef12e20357f9d7ba5431989c1daf4a556d6cb20ee4efd1c57320`
- Supabase CLI 2.109.0
  - storage-api v1.61.7
  - image digest `7dae5013474c1b5f816d138605e33434d83b87df1716032cf588a15561dcd937`

## Reproduction

1. Start a clean isolated local stack using one of the exact CLI versions.
2. Create a private bucket through the official Storage API.
3. Call `createSignedUploadUrl` for a random temporary object.
4. Verify only that the token is a decodable three-segment compact JWS with `alg = HS256`.
5. Send the initial resumable upload request with standard TUS metadata, `x-signature`, and `x-upsert: false`.
6. Observe HTTP 400 and `Invalid Compact JWS`.
7. Send the same request and token directly to the Storage container through stdin.
8. Observe the same result.

Relevant files:

- `src/reproduce.ts`
- `src/run-cli-matrix.ts`
- `src/sanitize.ts`
- `src/verify-no-secrets.ts`

## Minimal reproduction command

```powershell
bun install --frozen-lockfile
bun run matrix
bun run verify
```

An optional trusted CA file can be supplied with `--cafile <path-to-trusted-ca.pem>` when local HTTPS inspection is enabled.

## Expected behavior

The initial TUS request should accept the signed upload token generated by the same local Storage stack and create a resumable upload session with a session location.

## Actual behavior

During the earlier isolated diagnostic run:

- Gateway: HTTP 400, `Invalid Compact JWS` / `invalid_compact_jws`
- Direct Storage container: HTTP 400, `Invalid Compact JWS` / `invalid_compact_jws`
- No TUS session was created

## Control observations

- The token was a compact JWS.
[...]

PR #1342, opened 2026-08-22
fix(auth): return clear error on missing auth header and handle configurable tus sign path

Adityakk9031:

Bug fix / Developer Experience improvement

Related to #1268

1. When a client sends a request to an authenticated endpoint without an `Authorization` header, the JWT authentication plugin (`src/http/plugins/jwt.ts`) extracted an empty string `""` and attempted to verify it with `jose.jwtVerify`. This caused `jose` to throw an internal `JWSInvalid: Invalid Compact JWS` parsing error instead of an explicit error indicating that the authorization header is missing.
2. In `src/http/routes/tus/lifecycle.ts`, the signed URL check was hardcoded to `/upload/resumable/sign` rather than using the configured `tusPath` and `SIGNED_URL_SUFFIX` constants.

## What is the new behavior?

1. `jwtPlugin` now checks for missing `jwt` and returns `ERRORS.AccessDenied('Missing authorization header')` instead of querying the database for JWT secrets and invoking `verifyJWT` with an empty string.
2. `lifecycle.ts` uses `tusPath` and `SIGNED_URL_SUFFIX` constants to support configured TUS URL paths in addition to the default `/upload/resumable/sign`.

### DIFF
--- a/src/http/plugins/jwt.ts
+++ b/src/http/plugins/jwt.ts
@@ -46,6 +46,12 @@ const jwtPlugin = fastifyPlugin<JWTPluginOptions>(
         return
       }
 
+      if (!request.jwt) {
+        request.jwtPayload = { role: 'anon' }
+        request.isAuthenticated = false
+        throw ERRORS.AccessDenied('Missing authorization header')
+      }
+
       const { secret, jwks } = await getJwtSecret(request.tenantId)
 
       try {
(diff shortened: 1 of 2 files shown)


Adityakk9031:
@ferhatelmas  have a look

ferhatelmas:
@Adityakk9031 both change are in good direction. 

To be able to merge, better to keep them targeted because they are quite independent so better to split it them. We also need to regression coverage

ferhatelmas:
I also updated description to link `related to` and not to close the issue. We need corresponding documentation update as well.

PR #1345, opened 2026-08-23
fix: settle contended TUS lock waits on abort

### LINKED ISSUE #1344: Aborting a contended PostgreSQL TUS lock leaves acquisition pending
## Description

When PostgreSQL-backed TUS lock acquisition encounters an existing lock, aborting the acquisition can leave the lock promise pending instead of stopping promptly. The request can remain tied to the surrounding transaction timeout.

## Reproduction

1. Start PgLock acquisition with a database adapter whose mustLockObject call reports ResourceLocked.
2. Wait until the locker publishes its release request and enters the 500 ms retry delay.
3. Abort the acquisition signal.
4. Observe that acquireLock remains pending.

A deterministic unit reproduction resolves a sentinel on the next event-loop turn; current master returns the pending sentinel rather than settling acquisition.

## Expected behavior

Aborting a contended TUS lock acquisition should stop the retry loop and settle promptly without acquiring the lock.

## Root cause

The abort handlers for both the contention delay and the competing lock timeout clear their timers but do not resolve their promises. After abort, both branches of Promise.race can therefore remain pending. The outer stop-signal listener is also not removed when acquisition finishes.

## Environment

- supabase/storage master at 2a7e36f5cd0aa55949df8d6b73cc7695a57f6766
- Node.js 24.9.0
- npm 11.12.1

hsusul:
Fixes #1344

Bug fix

When PostgreSQL-backed TUS lock acquisition is waiting on a contended resource, aborting the acquisition clears both retry timers without resolving either promise. Both branches of the acquisition race can remain pending until the surrounding transaction timeout.

## What is the new behavior?

Abort-aware lock waits now settle and remove their listeners on both timeout and abort. The contention backoff reuses that behavior, already-aborted signals are honored, and the outer stop-signal listener is removed when acquisition finishes.

## Tests

Added a deterministic regression test that enters the ResourceLocked retry path, aborts after the release request is published, and verifies acquisition settles on the next event-loop turn.

Validated with:

- Focused PostgreSQL TUS locker tests: 4 passed
- Full unit suite: 137 files, 1,740 tests passed
- npm run lint
- npm run build
- git diff --check

## Additional context

#1115 also touches this file as part of a broader database transaction refactor, but its contended retry wait retains the same abort-time pending-promise behavior.

### DIFF
--- a/src/storage/protocols/tus/postgres-locker.test.ts
+++ b/src/storage/protocols/tus/postgres-locker.test.ts
@@ -1,5 +1,13 @@
+import { ERRORS } from '@internal/errors'
 import { PubSubAdapter } from '@internal/pubsub'
-import { LockNotifier } from './postgres-locker'
+import { Database } from '../../database'
+import { LockNotifier, PgLock } from './postgres-locker'
+
+class TestPgLock extends PgLock {
+  acquire(db: Database, id: string, signal: AbortSignal) {
+    return this.acquireLock(db, id, signal)
+  }
+}
 
 class FakePubSub implements PubSubAdapter {
   readonly startSpy = vi.fn(async () => undefined)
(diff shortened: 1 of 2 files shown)


fenos (approved):

ferhatelmas (approved):
we could simplify a bit and improve the test but can do as a follow up

PR #1347, opened 2026-08-24
feat: add bucket versioning status and version-aware object writes

TylerHillery:

Feature to factor in the bucket versioning status into object write and delete operations.

Currently, a bucket's versioning status can only be `DISABLED` and is not factored into operations such as PUT or DELETE.

## What is the new behavior?

All object write and delete operations now factor in the bucket's versioning status. The status can be `DISABLED`, `ENABLED`, or `SUSPENDED`.

Once versioning is enabled, the bucket can never return to `DISABLED` because enabling it permanently changes how certain operations are performed. Distinguishing a bucket that has never had versioning enabled is important because it requires less bookkeeping.

For example, in the `SUSPENDED` state, a PUT on an existing key still preserves the current object if it has `is_versioned` set to `true`, meaning it was created while the bucket was `ENABLED`. In the `DISABLED` state, we know the current object could never be versioned, so we can continue using the normal upsert behavior instead of archiving and inserting.

Here are the main changes for each status:

### ENABLED

**PUT**: We update the current version's `archived_at` and insert a new row with a different `version`. I refer to this as archive and insert.

**DELETE**: This follows the same archive-and-insert flow as PUT, but the newly inserted row has `is_delete_marker` set to `true`.

### SUSPENDED

**PUT**: This is similar to the archive-and-insert flow, with one key distinction. We archive the current row and then upsert on `(bucket_id, name) WHERE NOT is_versioned`. If a null version already exists for this object, it is replaced rather than retained as another historical version.

**DELETE**: This follows the same archive-and-upsert flow as PUT, but the null version becomes a delete marker. This is another important distinction between `SUSPENDED` and `DISABLED`, since `DISABLED` does not create delete-marker rows.

## Additional context

- To handle the question, "What happens if the bucket versioning status changes in the middle of a PUT or DELETE?", object writes take a shared lock on the bucket row, while versioning-status transitions require an exclusive lock.
- Another concept worth calling out is **promotion**. GET requests return the current version of an object. If a DELETE with a version ID hard-deletes the current version, we check for a previous version and promote the newest remaining one by setting its `archived_at` to `null`.
- It is important to reiterate that whether an object is protected by versioning depends on the bucket's status when that version was created, not its status during the current operation. An object with `is_versioned` set to `false` is not preserved when another PUT or DELETE occurs unless the bucket is currently `ENABLED`.


src/storage/object.ts
```diff
@@ -140,44 +193,70 @@ export class ObjectStorage {
    * @param objectName
    */
   async deleteObject(objectName: string, versionId?: string) {
-    const obj = await this.db.withTransaction(async (db) => {
-      const obj = await db.asSuperUser().findObject(
+    const eventObject = await this.db.withTransaction(async (db) => {
+      const superUserDb = db.asSuperUser()
+      await superUserDb.waitObjectLock(this.bucketId, objectName, undefined, { timeout: 5000 })
```
fenos: I think we might want to pass the versionId to the `waitObjectLock` otherwise we serialize all deletes for versions here

src/storage/object.ts
```diff
@@ -196,30 +275,146 @@ export class ObjectStorage {
...
-          ...(versionedEntries.length > 0
-            ? await db.deleteObjectVersions(this.bucketId, versionedEntries)
-            : []),
-        ]
+        const superUserDb = db.asSuperUser()
+        // Lock names from both target forms in one consistent order.
+        const namesToLock = [
+          ...new Set([...plainNames, ...versionedEntries.map((entry) => entry.name)]),
+        ].sort()
+        for (const name of namesToLock) {
```
fenos: this is going to be very inefficient, i don't think we need advisory locks here, the findForUpdate below would do optimistic locks on these rows

src/storage/object.ts
```diff
@@ -196,30 +275,146 @@ export class ObjectStorage {
...
+          await superUserDb.waitObjectLock(this.bucketId, name, undefined, { timeout: 5000 })
+        }
+
+        const lockedPlainObjects =
```
fenos: We could run these 2 promises in parallel and not sequetially

src/storage/object.ts
fenos: probably best refactoring this function to multiple smaller functions for easy of use and reasoning

src/storage/object.ts
```diff
@@ -491,33 +718,95 @@ export class ObjectStorage {
...
+
+    let sourceVersioningStatus: BucketVersioningStatus | undefined
+    let destinationVersioningStatus: BucketVersioningStatus | undefined
 
+    // Authorize before copying backend data, then repeat under the final locks
+    // to ensure the authorized rows did not change while the copy was prepared.
     await this.db.testPermission(async (db) => {
-      await db.findObject(this.bucketId, sourceObjectName, 'id', undefined, sourceVersionId)
-      return db.updateObject(
+      for (const key of objectKeys) {
```
fenos: We need to try and reduce serial locks like this ones

src/storage/object.ts
```diff
@@ -491,33 +718,95 @@ export class ObjectStorage {
...
+        await db.waitObjectLock(key.bucketId, key.name, undefined, { timeout: 5000 })
+      }
+
+      const superUserDb = db.asSuperUser()
+      const statuses = new Map<string, BucketVersioningStatus>()
+      for (const bucketId of [...new Set([this.bucketId, destinationBucket])].sort()) {
+        const bucket = await superUserDb.findBucketById(bucketId, 'versioning_status', {
```
fenos: we can batch select here saving serial queries

src/storage/object.ts
```diff
@@ -538,44 +827,156 @@ export class ObjectStorage {
...
-        })
+      return this.db.withTransaction(async (db) => {
+        const superUserDb = db.asSuperUser()
+
+        for (const key of objectKeys) {
+          await superUserDb.waitObjectLock(key.bucketId, key.name, undefined, { timeout: 5000 })
+        }
+
+        const lockedStatuses = new Map<string, BucketVersioningStatus>()
+        for (const bucketId of [...new Set([this.bucketId, destinationBucket])].sort()) {
```
fenos: serial query, avoid

src/storage/object.ts
```diff
@@ -538,44 +827,156 @@ export class ObjectStorage {
...
             dontErrorOnEmpty: false,
+            excludeDeleteMarkers: true,
           },
           sourceVersionId
         )
 
-        await db.updateObject(
-          this.bucketId,
-          sourceObjectName,
+        if (sourceObject.id !== sourceObj.id || sourceObject.version !== sourceObj.version) {
```
fenos: wondering if we should start considering adding assertion function helpers in limits.ts

src/storage/uploader.ts
```diff
@@ -69,32 +69,45 @@ export class Uploader {
     private readonly location: StorageObjectLocator
   ) {}
 
+  async authorizeUpload(
+    db: Database,
+    options: CanUploadOptions,
+    currentObjectIsDeleteMarker: boolean
```
fenos: probably best to add it to the CanUploadOptions

src/storage/protocols/s3/s3-handler.ts
```diff
@@ -48,11 +48,21 @@ export class S3ProtocolHandler {
    *
    * Reference: https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketVersioning.html
    */
-  async getBucketVersioning() {
+  async getBucketVersioning(bucket: string) {
+    const bucketRecord = await this.storage.db
+      .asSuperUser()
+      .findBucketById(bucket, 'versioning_status')
+    const versioningStatus = bucketRecord.versioning_status ?? 'DISABLED'
```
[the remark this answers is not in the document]
TylerHillery: S3 support is planned for follow up PR

src/storage/object.ts
```diff
@@ -79,6 +85,53 @@ export class ObjectStorage {
...
+      user_metadata: userMetadata,
+      version: newVersion,
+    })
+
+    const authorizedDelete = await db.deleteObject(
+      this.bucketId,
+      sourceObjectName,
+      sourceObject.version,
+      { skipPromotion: true }
+    )
```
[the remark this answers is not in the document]
TylerHillery: This mismatch is intentional. We check for DELETE permission because this is a logical deletion, then use the privileged role to carry out the actual storage semantics. We check permissions on the logical operation, not the underlying storage semantics used to perform it.

src/storage/versioning/guards.ts
```diff
@@ -0,0 +1,54 @@
...
+import { ERRORS } from '@internal/errors'
+import { getConfig } from '../../config'
+import type { Database } from '../database'
+import type { BucketType } from '../limits'
+
+interface VersioningStorageContext {
+  db: Pick<Database, 'hasMigration' | 'tenantId'>
+}
+
+function assertVersioningApiEnabled(bucketId: string): void {
```
fenos, line 11: i like this, maybe we can create a folder hierarchy like this: `storage/guards/versioning.ts,storage/guards/limit.ts`

src/storage/database/pg.ts
```diff
@@ -160,15 +162,23 @@ export class StoragePgDB implements Database {
       !this.latestMigration ||
       (migrationOrdinal !== undefined &&
         migrationOrdinal >= DBMigration['s3-multipart-uploads-metadata'])
-    this.objectColumnPolicy = this.supportsCustomMetadataColumns
-      ? SelectColumnPolicy.none
-      : SelectColumnPolicy.objectWithoutUserMetadata
+    this.objectColumnPolicy =
```
fenos, line 186: We need to keep in mind that we need to refactor this, maybe in this PR or following ones

src/storage/uploader.ts
```diff
@@ -202,87 +221,121 @@ export class Uploader {
...
+            }
+          )
 
-        // schedule the deletion of the previous file
-        if (currentObj && currentObj.version !== version) {
-          events.push(
-            ObjectAdminDelete.send({
+          if (!isUpsert && currentObj && !currentObj.is_delete_marker) {
+            throw ERRORS.KeyAlreadyExists(objectName)
+          }
```
ferhatelmas, line 264: TUS can reply the same version. If it's committed, catch will delete bytes of an object and lose data

migrations/tenant/0002-storage-schema.sql
```diff
@@ -77,7 +77,24 @@ CREATE TABLE IF NOT EXISTS "storage"."objects" (
...
+    IF NOT EXISTS (
+        SELECT 1
+        FROM pg_catalog.pg_proc p
+        JOIN pg_catalog.pg_namespace n ON n.oid = p.pronamespace
+        WHERE n.nspname = 'storage'
+          AND p.proname = 'enforce_versioning_status_transition'
+    ) THEN
+        CREATE UNIQUE INDEX IF NOT EXISTS "bucketid_objname" ON "storage"."objects" USING BTREE ("bucket_id","name");
+    END IF;
+END$$;
```
ferhatelmas, line 97: is it safe to update old migrations?

src/storage/database/pg.ts
```diff
@@ -1080,66 +1305,174 @@ export class StoragePgDB implements Database {
...
+    }
 
-      return this.query<Obj>(
+    const result = await this.runQuery(operationName, async (db, signal) => {
+      if (versioningStatus !== 'DISABLED') {
+        await this.query(
+          db,
+          {
+            text: `
+              UPDATE storage.objects SET archived_at = now()
```
ferhatelmas: this is tx start time, not monotonic, it can change ordering

src/storage/database/pg.ts
```diff
@@ -1234,19 +1567,48 @@ export class StoragePgDB implements Database {
...
+          if (versioningStatus === 'DISABLED') {
+            return undefined
+          }
+          return db.writeCurrentVersion('DeleteObject', bucketId, objectName, versioningStatus, {
+            version: randomUUID(),
+            is_delete_marker: true,
+            metadata: null,
+            user_metadata: null,
+          })
+        },
```
ferhatelmas, line 1596: owner is missing

migrations/tenant/0002-storage-schema.sql
fenos: Yes, the hash will simply update to the new version but not run

ferhatelmas: one hazard, it checks function introduced in 73 but index is dropped in 72. Isn't it better to guard against its replacement index 66?

src/storage/object.ts
```diff
@@ -138,123 +368,396 @@ export class ObjectStorage {
...
+          versionId
+        )
 
-      const deleted = await db.deleteObject(this.bucketId, objectName, versionId)
+        const authorized = await db.testPermission((permissionDb) =>
+          permissionDb.deleteObject(this.bucketId, objectName, obj?.version, {
+            skipPromotion: true,
+            versioningStatus,
+          })
+        )
```
ferhatelmas, line 394: noting, there is an inconsistency here according to existence (obj.version),  probe checks delete or insert but if enabled, underlying operation is always  an insert

src/storage/object.ts
```diff
@@ -372,65 +882,100 @@ export class ObjectStorage {
...
+              metadata: {
+                ...destinationMetadata,
+                lastModified: copyResult.lastModified,
+                eTag: copyResult.eTag,
+              },
+              user_metadata: destinationUserMetadata,
+              version: newVersion,
+            },
+            { versioningStatus }
+          )
```
ferhatelmas, line 977: we're spreading pre-copy values here

src/storage/object.ts
```diff
@@ -491,137 +1042,297 @@ export class ObjectStorage {
...
+      this.bucketId === destinationBucket && sourceObjectName === destinationObjectName
+    const isSamePathNoop = isSamePath && sourceVersionId === undefined
+    const objectKeys = [
+      { bucketId: this.bucketId, objectName: sourceObjectName },
+      { bucketId: destinationBucket, objectName: destinationObjectName },
+    ]
+
+    // Authorize before copying backend data. This pass takes no locks: the
+    // pass after the copy re-reads everything under the final locks and
+    // rejects the move if the status snapshot or the source row changed.
```
ferhatelmas, line 1071: I think no locks claim is wrong due to `authorizeMove`, so I wonder a case where deadlock is possible. How about taking locks here?

migrations/tenant/0002-storage-schema.sql
fenos: good catch, fixed in 0b2408b0. The guard now checks for idx_objects_current_version from 0066, which never gets dropped or renamed by a later migration. I reproduced the window on a scratch db: migrated to 0072, rewound the migrations table, replayed 0002. The old guard rebuilt the index, the new one skips it, and a fresh install still creates it.

src/storage/object.ts
fenos: this one is intentional, the probe authorizes what the user is doing rather than the SQL we end up running. Deleting an existing key stays governed by the DELETE policy, so enabling versioning doesn't change who can delete (the marker insert is superuser bookkeeping). Deleting a missing key writes a marker where nothing existed, so that case is gated by the INSERT policy instead. Added a comment at the probe site in 596f0045.

src/storage/object.ts
fenos: you're right, the probes do take row locks so two opposite moves could deadlock. Fixed in fd12f4ef: the authorize pass now takes the advisory locks inside the rolled back transaction, same gate as every other writer, and they're gone before the backend copy starts. Reworded the comment too.

src/storage/object.ts
fenos: removed in 596f0045, the upsert input now lists exactly the fields the write reads.

src/storage/uploader.ts
fenos: fixed, a replayed completion of the version that's already current returns early instead of writing again (the currentObj?.version === version guard in completeUpload), so the catch can't delete live bytes anymore.

src/storage/database/pg.ts
fenos: fixed, owner is now passed through both the single delete marker write and writeDeleteMarkers, and recorded as the marker's owner.

src/storage/uploader.ts
```diff
@@ -202,99 +221,176 @@ export class Uploader {
...
+            }
+          )
 
-        // schedule the deletion of the previous file
-        if (currentObj && currentObj.version !== version) {
-          events.push(
-            ObjectAdminDelete.send({
+          if (currentObj?.version === version) {
+            // A replayed completion of a version that already committed, for
+            // example a retried final TUS PATCH. The row is in place and its
```
ferhatelmas, line 255: are we handling ongoing write in this case?

src/storage/uploader.ts
```diff
@@ -202,99 +221,176 @@ export class Uploader {
...
+            events.push(
+              ObjectAdminDelete.send({
+                name: objectName,
+                bucketId,
+                tenant: this.db.tenant(),
+                version: replaced.version ?? undefined,
+                reqId: this.db.reqId,
+                sbReqId: this.db.sbReqId,
+              })
+            )
```
ferhatelmas: what if tx is rolled back?

src/storage/object.ts
```diff
@@ -138,123 +374,412 @@ export class ObjectStorage {
...
+          await this.backend.deleteObject(
+            this.location.getRootLocation(),
+            this.location.getKeyLocation({
+              tenantId: this.db.tenantId,
+              bucketId: this.bucketId,
+              objectName,
+            }),
+            freed.version ?? undefined
+          )
+        }
```
ferhatelmas: similar concern here https://github.com/supabase/storage/pull/1347/changes#r4004709340

src/storage/database/pg.ts
```diff
@@ -1080,66 +1306,179 @@ export class StoragePgDB implements Database {
...
+    let conflictClause = ''
+    if (versioningStatus === 'SUSPENDED') {
+      conflictClause = `
+        ON CONFLICT (bucket_id, name COLLATE "C") WHERE NOT is_versioned
+        DO UPDATE SET ${[...updateClauses, 'archived_at = NULL'].join(', ')}
+      `
+    } else if (versioningStatus === 'DISABLED') {
+      const conflictTarget = (await this.hasMigration('objects-current-version-index'))
+        ? 'ON CONFLICT (bucket_id, name COLLATE "C") WHERE archived_at IS NULL'
+        : 'ON CONFLICT (name, bucket_id)'
```
ferhatelmas, line 1403: there is a gotcha here and critical deploy order, if tenant starts getting migration 73, it's not known by master's migrations and this check will be false and will try to use a dropped index and fail

src/storage/object.ts
```diff
@@ -491,137 +1063,305 @@ export class ObjectStorage {
...
         s3DestinationKey,
-        newVersion
+        move.newVersion
       )
 
-      return this.db.asSuperUser().withTransaction(async (db) => {
-        await db.waitObjectLock(this.bucketId, destinationObjectName, undefined, {
-          timeout: 5000,
-        })
+      return this.db.withTransaction((db) =>
```
ferhatelmas: no await so its catch isn't running

src/http/routes/bucket/getBucket.ts
```diff
@@ -36,10 +36,12 @@ export default async function routes(fastify: FastifyInstance) {
...
       const { bucketId } = request.params
 
-      const results = await request.storage.findBucket(
-        bucketId,
-        'id, name, owner, public, created_at, updated_at, file_size_limit, allowed_mime_types'
-      )
+      const hasVersioningColumns = await request.storage.db.hasMigration('object-versioning-core')
+      const columns =
+        'id, name, owner, public, created_at, updated_at, file_size_limit, allowed_mime_types' +
+        (hasVersioningColumns ? ', versioning_status' : '')
```
ferhatelmas, line 42: we add it conditionally in getAllBuckets but unconditionally here, I think we should do the same. In another note, it's already inconsistent for type, not sure if we should drop the condition at this point since sdk is so old

src/storage/object.ts
```diff
@@ -372,65 +904,99 @@ export class ObjectStorage {
...
-        this.location.getRootLocation(),
-        s3SourceKey,
-        originObject.version,
-        s3DestinationKey,
-        newVersion,
-        destinationMetadata,
-        conditions,
-        { copyMetadata }
-      )
+      const { result: copyResult } = await this.copySourceContent({
```
ferhatelmas, line 915: this can retry but we discard it but actually we should use it for metadata

src/storage/protocols/s3/s3-handler.ts
```diff
@@ -1070,7 +1080,7 @@ export class S3ProtocolHandler {
     }
 
     try {
-      await this.storage.from(Bucket).deleteObject(Key)
+      await this.storage.from(Bucket).deleteObject(Key, undefined, { owner: this.owner })
```
ferhatelmas, line 1083: since writers are enabled now, I think we should guard version id for s3

src/storage/database/pg.ts
```diff
@@ -1422,22 +1984,107 @@ export class StoragePgDB implements Database {
...
+      values.push(
+        versions.map((target) => target.name),
+        versions.map((target) => target.version)
+      )
+      targetPredicates.push(
+        `(name COLLATE "C", version) IN (SELECT * FROM unnest($${values.length - 1}::text[], $${values.length}::text[]))`
+      )
+    }
+
+    const conditions = ['bucket_id = $1', `(${targetPredicates.join(' OR ')})`]
```
ferhatelmas: it seems this OR can force seq scan if mixed

src/storage/database/pg.ts
```diff
@@ -1258,20 +1627,146 @@ export class StoragePgDB implements Database {
...
+            text: `
+              UPDATE storage.objects
+              SET archived_at = NULL
+              WHERE id = (
+                SELECT id
+                FROM storage.objects
+                WHERE bucket_id = $1 AND name COLLATE "C" = $2
+                ORDER BY archived_at DESC NULLS LAST, created_at DESC, version DESC
+                LIMIT 1
+                FOR UPDATE
```
ferhatelmas, line 1653: we don't have the index for ordering, this can go over working memory while sorting and spill into disk for keys with many versions. Do we want an index or limit the versions possible?

src/storage/object.ts
```diff
@@ -138,123 +374,412 @@ export class ObjectStorage {
...
+        if ((await probe([name])).length > 0) {
+          authorized.push(name)
+        }
+      } catch (e) {
+        if (!isStorageError(ErrorCode.AccessDenied, e)) {
+          throw e
         }
+      }
+    }
+    return authorized
```
ferhatelmas, line 694: This sequential probe loop is a bit concerning since locks are held and its timeout is 5s. Is it worth of moving it to a SQL function to prevent round trips or doing binary splits the names into chunks?

src/storage/object.ts
```diff
@@ -138,123 +374,412 @@ export class ObjectStorage {
...
+    superUserDb: Database,
+    targets: DeleteTargets
+  ): Promise<LockedDeleteTargets> {
+    const names = [
+      ...new Set([...targets.plainNames, ...targets.versionedEntries.map((entry) => entry.name)]),
+    ]
+    await superUserDb.waitObjectLocks(
+      names.map((objectName) => ({ bucketId: this.bucketId, objectName })),
+      { timeout: 5000 }
+    )
```
ferhatelmas, line 558: Compared to single delete, for noncurrent, lock key doesn't include version so they aren't serialized at advisory lock. Is it intentional not to stall writes?

src/storage/database/pg.ts
```diff
@@ -1412,6 +1972,8 @@ export class StoragePgDB implements Database {
             SELECT ${selectedColumns}
             FROM storage.objects
             WHERE ${conditions.join(' AND ')}
+            ${shouldLock ? 'ORDER BY name' : ''}
```
ferhatelmas: collation is missing

src/storage/database/pg.ts
```diff
@@ -1422,22 +1984,107 @@ export class StoragePgDB implements Database {
...
+      return this.query<Obj>(
         db,
         {
           text: `
-            SELECT name, version
+            SELECT ${selectedColumns}
             FROM storage.objects
             WHERE bucket_id = $1
               AND (name COLLATE "C", version) IN (${placeholders})
+            ${shouldLock ? 'ORDER BY name, version' : ''}
```
ferhatelmas: collation here as well

src/storage/database/pg.ts
```diff
@@ -1258,20 +1627,146 @@ export class StoragePgDB implements Database {
...
+      const object = deleted.rows[0]
+      // Promotion locks its candidate row: a candidate hard-deleted by a
+      // concurrent transaction is skipped in favour of the next one instead of
+      // the UPDATE matching nothing and leaving the key without a current row.
+      if (
+        object?.archived_at === null &&
+        version !== undefined &&
+        !options.skipPromotion &&
+        (await this.hasMigration('object-versioning-core'))
+      ) {
```
ferhatelmas, line 1640: why don't we also use versioning hint to short cut?

src/storage/uploader.ts
```diff
@@ -202,99 +221,176 @@ export class Uploader {
...
-          metadata: objectMetadata,
-          user_metadata: userMetadata,
-          version,
-          owner,
-        })
-
-        const events: Promise<unknown>[] = []
+      this.db.connection.setAbortSignal(abController.signal)
+
+      return await this.db.withTransaction((scopedDb) =>
```
ferhatelmas: it seems like if we start with superuser and then probe with user, we will be able to save one save point

src/storage/uploader.ts
```diff
@@ -69,32 +76,44 @@ export class Uploader {
...
+    let currentObjectIsDeleteMarker = false
+    if (await this.db.hasMigration('object-versioning-core')) {
+      const currentObject = await this.db
+        .asSuperUser()
+        .findObject(options.bucketId, options.objectName, 'is_delete_marker', {
+          dontErrorOnEmpty: true,
         })
-      })
+      currentObjectIsDeleteMarker = currentObject?.is_delete_marker === true
     }
```
ferhatelmas, line 114: maybe skip the lookup, try to create and retry only if key exists to make successful case faster

ferhatelmas:
shall we expose version on writes?

src/storage/versioning/guards.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
+
+async function assertVersioningTenantEnabled(
+  storage: VersioningStorageContext,
+  bucketId: string
+): Promise<void> {
+  const { isMultitenant, storageVersioningEnabled } = getConfig()
+
+  if (isMultitenant && !storageVersioningEnabled && !(await tenantHasFeature(storage.db.tenantId, 'objectVersioning'))) {
```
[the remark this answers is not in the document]
ferhatelmas: I think this is valid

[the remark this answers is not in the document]
ferhatelmas: if it's kept false, then it will fallback tenant config so it can work, just not very intuitive

src/storage/object.ts
```diff
@@ -491,140 +1085,330 @@ export class ObjectStorage {
...
-        bucketId: destinationBucket,
-        tenant: this.db.tenant(),
-        version: newVersion,
-        reqId: this.db.reqId,
-        sbReqId: this.db.sbReqId,
-      })
+      // The copied destination bytes are unreferenced only while nothing
+      // committed; a post-commit failure must not delete the live row's
+      // content.
+      if (!committed) {
```
ferhatelmas: I think this client side flag isn't safe even if rare. If committed but ack is lost, we will delete a live row. Better to do a real probe similar to upload (`isCommittedVersion`)

src/storage/database/pg.ts
```diff
@@ -1354,18 +1895,28 @@ export class StoragePgDB implements Database {
...
     const conditions = ['name COLLATE "C" = $1', 'bucket_id = $2']
     const values: unknown[] = [objectName, bucketId]
 
-    if (version !== undefined) {
+    const hasVersioning = await this.hasMigration('object-versioning-core')
+    if (version === null) {
+      conditions.push('version IS NULL')
+    } else if (version !== undefined) {
       values.push(version)
       conditions.push(`version = $${values.length}`)
```
ferhatelmas, line 1909: there is an edge case for old schemas where version can be null. Client will still pass string `null` but this condition can't match it

src/storage/uploader.ts
```diff
@@ -202,55 +221,102 @@ export class Uploader {
...
           events.push(
             ObjectAdminDelete.send({
               name: objectName,
               bucketId,
               tenant: this.db.tenant(),
-              version: currentObj.version,
+              version: replaced.version ?? undefined,
               reqId: this.db.reqId,
               sbReqId: this.db.sbReqId,
             })
```
ferhatelmas, line 330: we need a catch handler for this

PR #1348, opened 2026-08-25
fix: add subscription_tier to tenant admin endpoints and database

itslenny:

The tenant config admin api ignores the `subscription_tier` value if present

## What is the new behavior?

Accept `subscription_tier` in tenant body and persist along with other tenant config


fenos (approved):

src/http/routes/admin/tenants.ts
```diff
@@ -644,6 +655,10 @@ export default async function routes(fastify: FastifyInstance) {
         tenantInfo.tracing_mode = tracingMode
       }
 
+      if (subscriptionTier) {
+        tenantInfo.subscription_tier = subscriptionTier
+      }
```
ferhatelmas, line 660: nit: fine due to its caller today but for completeness, undefined check might be better

PR #1364, opened 2026-09-03
feat: topK 10_000 support

ferhatelmas:

feat

Vector queries support up to 100 results per request.

## What is the new behavior?

S3 vector queries support up to 10,000 results per request, no change for pgvector.
Pagination is supported through nextToken for S3-based adapter.

## Additional context

https://aws.amazon.com/about-aws/whats-new/2026/06/s3-vectors-supports-10000-search-results-per-query/
https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_QueryVectors.html


TylerHillery (approved):

acceptance/specs/vector.test.ts
```diff
@@ -444,6 +445,52 @@ describeAcceptance(
...
+                float32: [1, 0],
+              },
+              topK: 2,
+              vectorBucketName,
+            },
+            expectedStatus: 400,
+            token,
+          })
+        }
+
```
itslenny, line 536: Might be nice to add a test that actually uses a nextToken against real S3 backend to confirm nextToken is being forwarded correctly.

src/storage/protocols/vector/limits.ts
```diff
@@ -2,7 +2,9 @@ import type { PutVectorsInput } from '@aws-sdk/client-s3vectors'
 import { ERRORS } from '@internal/errors'
 
 export const MIN_VECTOR_DIMENSIONS = 1
-export const MAX_QUERY_TOP_K = 100
+export const MAX_QUERY_TOP_K = 10_000
```
itslenny: nit: maybe this should be `MAX_S3_QUERY_TOP_K` or something since it's no longer THE limit, but the limit for a specific provider

itslenny (approved):

src/storage/protocols/vector/limits.ts
ferhatelmas: good idea, done

acceptance/specs/vector.test.ts
ferhatelmas: if we had vendoring in acceptance tests, could add here, otherwise, better separately. Let me see

ferhatelmas: I realized, we don't enable vector by default so added the test since won't create noise

PR #1365, opened 2026-09-04
fix: complete empty TUS multipart uploads when Upload-Length is 0

### LINKED ISSUE #1300: Resumable (tus) upload creation with Upload-Length: 0 returns 404 with no explanation
## Summary

The resumable upload endpoint responds `404 Not Found` to a creation POST
whose `Upload-Length` is `0`. A zero-byte object is a degenerate but valid
tus upload; I'd expect either `201 Created` (with the upload immediately
completable) or a meaningful `4xx` with an error body naming the constraint.

The bare 404 is the real problem: it reads as "wrong endpoint" or "transient",
not "this request can never succeed". Retrying tus clients (aiotus in my
case) treat it as retryable and loop forever.

## Repro

Against hosted storage (also reproduces on the local development stack):

    curl -si -X POST "https://<project>.storage.supabase.co/storage/v1/upload/resumable" \
      -H "Authorization: Bearer <service-role-key>" \
      -H "Tus-Resumable: 1.0.0" \
      -H "Upload-Length: 0" \
      -H "Upload-Metadata: bucketName cmVjb3JkaW5ncw==,objectName emVyby5vZ2c=" \
      # -> HTTP/2 404, no error body explaining the rejection

The identical request with `Upload-Length: 1` (and one byte PATCHed) returns
`201` and completes normally.

## How this bit me

I upload call recordings via tus. Occasionally an upstream media failure
produces a zero-byte file, and my uploader — which bounds its shutdown flush
to 60 seconds so restarts don't lose recordings — spent that entire window
retrying a request that could never succeed. Nothing in the 404 indicates
permanence, so it took a while to trace the "storage outage" back to a
zero-length edge case. I now skip zero-byte uploads client-side, but the
failure mode was avoidable.

## Ask

Either accept the zero-byte creation (`201`) or reject it with a descriptive
`4xx` (e.g. `422` and a body saying zero-length uploads are unsupported), and
document the constraint in the resumable-upload docs if it's intentional.
Either lets clients fail fast instead of retrying forever.

hsusul:
## Summary

- Fixes #1300: resumable upload creation with `Upload-Length: 0` returned a bare `404` because TUS treats size-0 as immediately final (no `write()`), while `@tus/s3-store` `create()` only starts a multipart upload. Our `onUploadFinish` then `HeadObject`s a key that does not exist yet.
- Override `S3Store.create` to call upstream `finishMultipartUpload(metadata, [])` (which already uploads a zero-byte part when `parts` is empty), then complete metadata and clear cache, so the object exists before finish runs.
- Unit tests cover size `0` finalization and that non-zero / deferred sizes are unchanged.

## Test plan

- [x] `npm run test:unit -- src/storage/protocols/tus/s3-store.test.ts`
- [ ] Manual: `POST /storage/v1/upload/resumable` with `Upload-Length: 0` and valid `Upload-Metadata` → `201`, object exists as empty
- [ ] Manual: same with `Upload-Length: 1` + PATCH still succeeds

### DIFF
--- a/src/storage/protocols/tus/s3-store.test.ts
+++ b/src/storage/protocols/tus/s3-store.test.ts
@@ -1,4 +1,6 @@
 import { HttpResponse } from '@smithy/protocol-http'
+import { S3Store as TusS3Store } from '@tus/s3-store'
+import type { Upload } from '@tus/server'
 import { S3Store } from './s3-store'
 
 class TestS3Store extends S3Store {
@@ -26,7 +28,24 @@ function createStore(
   })
 }
 
+function baseUpload(overrides: Partial<Upload> = {}): Upload {
+  return {
+    id: 'tenant/bucket/empty.txt/version-1',
+    offset: 0,
+    size: 0,
+    sizeIsDeferred: false,
+    metadata: { contentType: 'text/plain' },
+    creation_date: undefined,
+    storage: undefined,
+    ...overrides,
+  } as Upload
+}
+
 describe('S3Store', () => {
+  afterEach(() => {
+    vi.restoreAllMocks()
+  })
+
   test('removes the no-op logger middleware from the internal TUS client', () => {
     const store = createStore(vi.fn())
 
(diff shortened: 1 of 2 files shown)


ferhatelmas:
Thanks, we need to fix it in [upstream](https://github.com/tus/tus-node-server), not here. Could you direct your patch there? I can give look for review

hsusul:
Thanks — agreed this belongs upstream.

Opened the fix in tus-node-server: https://github.com/tus/tus-node-server/pull/876

Closing this PR in favor of that.

PR #1368, opened 2026-09-08
fix: free iceberg shards slots by catalog id, not name

ferhatelmas:

Bug fix

Iceberg shard slots are allocated by id but freed by name so leaks.

## What is the new behavior?

It's now always using id consistently.
Add admin triggered reclaim job that only does when slot identity matches tenant/resource, no local table metadata, warehouse is reachable, table is gone.


ferhatelmas:
@claude review always

PR #1369, opened 2026-09-08
feat: allow setting stale-while-revalidate and stale-if-error headers

itslenny:

feature

A `stale-while-revalidate` header is only added for mismatched etag and `stale-if-error` is never set

## What is the new behavior?

- Set `stale-while-revalidate` and `stale-if-error` headers based on env config
- For mismatched etags use the configured value if set (otherwise fallback to existing behavior - 30s hard coded)
- Correctly merge and de-dupe cache control directives

This PR should not change any behavior unless `RESPONSE_STALE_WHILE_REVALIDATE` or `RESPONSE_STALE_IF_ERROR` are set.


TylerHillery (approved):

src/storage/renderer/renderer.ts
```diff
@@ -198,7 +203,9 @@ export abstract class Renderer {
   }
 
   protected setCacheControlHeader(response: FastifyReply, values: Array<string | undefined>) {
-    const cacheControl = values.filter((value) => typeof value === 'string' && value.length > 0)
+    const cacheControl = values.filter(
+      (value): value is string => typeof value === 'string' && value.length > 0
+    )
```
ferhatelmas: I guess this diff is unncessary

src/storage/renderer/cache-control.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Combines a base set of Cache-Control directives with additional ones,
+ * skipping any addition whose directive name is already present in the base
+ */
+export function mergeCacheControlDirectives(
+  base: Array<string | undefined>,
+  additions: readonly string[]
+): string[] {
+  const directives = base.filter(
```
ferhatelmas, line 67: let's add an early return if `additions` is empty

src/storage/renderer/cache-control.ts
```diff
@@ -0,0 +1,72 @@
...
+      current = ''
+      continue
+    }
+
+    current += char
+  }
+
+  directives.push(current.trim())
+
+  return directives.filter((directive) => directive.length > 0)
```
ferhatelmas: can we prevent this not to allocate again?

src/storage/renderer/cache-control.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function cacheControlDirectiveName(directive: string): string {
+  return directive.split('=')[0].trim().toLowerCase()
```
ferhatelmas: if we slice, could prevent interim array

src/storage/renderer/cache-control.test.ts
```diff
@@ -0,0 +1,86 @@
+import { mergeCacheControlDirectives } from './cache-control'
+
+describe('mergeCacheControlDirectives', () => {
```
ferhatelmas, line 3: would be nice to have a no-cache test

src/storage/renderer/cache-control.ts
```diff
@@ -0,0 +1,86 @@
...
+  base: Array<string | undefined>,
+  additions: readonly string[]
+): string[] {
+  const directives = base.filter(
+    (value): value is string => typeof value === 'string' && value.length > 0
+  )
+
+  if (additions.length === 0) {
+    return directives
+  }
```
ferhatelmas: nit: filtering is already done by header setter, could be before

src/storage/renderer/cache-control.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  it('should append additions to a bare no-cache base directive', () => {
+    // Storage sets `no-cache` as the default object Cache-Control when no
+    // max-age is provided at upload time, so this is the Cache-Control
+    // value most objects actually have.
+    expect(
+      mergeCacheControlDirectives(
+        ['no-cache'],
+        ['stale-while-revalidate=120', 'stale-if-error=300']
+      )
+    ).toEqual(['no-cache', 'stale-while-revalidate=120', 'stale-if-error=300'])
```
ferhatelmas: may additions here aren't harmful but extra, do you think if it's worth of not emitting them?

PR #1378, opened 2026-09-10
fix: url-decode x-amz-copy-source in CopyObject and UploadPartCopy

### LINKED ISSUE #1377: S3 CopyObject/UploadPartCopy do not URL-decode x-amz-copy-source
### Bug report

The S3 protocol handler never URL-decodes the `x-amz-copy-source` header, so `CopyObject` and `UploadPartCopy` fail for any source key that requires percent-encoding.

Per the [CopyObject reference](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html), the copy source is a URL-encoded path, and clients encode it: botocore/AWS CLI percent-encode the key (`safe='/~'`) before sending the header, and the other AWS SDKs do the same. This repository does it too when it talks to a real S3 backend — `src/storage/backend/s3/backup.ts` sends `CopySource: encodeURIComponent(...)`.

But the gateway side parses the header raw:

https://github.com/supabase/storage/blob/master/src/storage/protocols/s3/s3-handler.ts — `copyObject()` and `uploadPartCopy()` both do

```ts
const sourceKey = (CopySource.startsWith('/') ? CopySource.replace('/', '') : CopySource)
  .split('/')
  .slice(1)
  .join('/')
```

so a compliant client copying `folder/my file+1.txt` sends `bucket/folder/my%20file%2B1.txt` and the server looks up the literal key `folder/my%20file%2B1.txt`, which does not exist.

**Impact**

- `aws s3 cp s3://bucket/"my file.txt" s3://bucket/dest.txt` → `NoSuchKey`, for any key containing a space, `+`, `#`, `?`, `%`, or non-ASCII characters.
- Same failure for server-side multipart copies (`UploadPartCopy`), which is how large-object copies are performed.
- A source key containing an encoded `%2F` is currently split as a path separator, so the wrong bucket/key pair is derived.

**Expected**

Decode the header the way S3 does: split on the unencoded `/` separators, then percent-decode the bucket and each key segment (so `%2F` stays a literal slash inside the key).

**Notes**

Decoding has to tolerate legacy callers that send raw keys containing a stray `%` (e.g. `100%-done.txt`), which is not a valid escape sequence — those should be left as-is rather than raising.

I have a fix with unit-test coverage and will open a PR referencing this issue.

hsusul:
## What

URL-decode `x-amz-copy-source` in `CopyObject` and `UploadPartCopy`.

Fixes #1377.

## Why

The copy source is a URL-encoded path per the [CopyObject reference](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html), and clients encode it — botocore/AWS CLI percent-encode the key with `safe='/~'` before sending the header, and this repo does the same when it talks to a real S3 backend (`src/storage/backend/s3/backup.ts` sends `CopySource: encodeURIComponent(...)`).

The protocol handler parsed the header raw, so `aws s3 cp s3://bucket/"my file.txt" s3://bucket/dest.txt` sent `bucket/my%20file.txt` and the server looked up the literal key `my%20file.txt` → `NoSuchKey`. This affected every key containing a space, `+`, `#`, `?`, `%` or non-ASCII character, for both plain copies and server-side multipart copies. An encoded `%2F` was additionally treated as a path separator, deriving the wrong bucket/key pair.

## How

Both commands duplicated the same parsing, so it is extracted into one `parseCopySource()` helper that:

- strips the optional leading `/`,
- splits on the unencoded `/` separators **first**, then percent-decodes the bucket and each key segment — so `%2F` stays a literal slash inside the key, matching S3,
- leaves a segment untouched when it is not a valid escape sequence, so existing callers that send raw keys containing a stray `%` (e.g. `100%-done.txt`) keep working instead of hitting a `URIError`.

## Tests

Added to `src/storage/protocols/s3/s3-handler.test.ts`:

- `CopyObject` decodes the source key (`folder/my%20file%2B1.txt` → `folder/my file+1.txt`)
- `CopyObject` decodes the source bucket (`/my%20bucket/...` → `my bucket`)
- `CopyObject` keeps an encoded `%2F` as a literal slash in the key
- `CopyObject` leaves an undecodable key (`100%-done.txt`) untouched — regression guard for legacy raw callers
- `UploadPartCopy` decodes the source key before the object lookup

All four decoding tests fail on `master` and pass with this change; the legacy-raw guard passes both before and after.

## Validation

Run locally on macOS / Node v24.9.0:

- `npm run test:unit` — 147 files, 2119 tests, all passing
- `npm run build` (`tsc -noEmit` + bundle) — passing
- `npm run lint` (biome + prettier) — clean
- `git diff --check` — clean

Not run: `npm run test:integration`. It needs the Docker Postgres stack and port 5432 was already occupied on this machine, so I did not start it. The change is confined to header parsing and is covered by the unit tests above.

### DIFF
--- a/src/storage/protocols/s3/s3-handler.test.ts
+++ b/src/storage/protocols/s3/s3-handler.test.ts
@@ -373,3 +373,109 @@ describe('S3ProtocolHandler.abortMultipartUpload', () => {
     expect(deleteMultipartUpload).not.toHaveBeenCalled()
   })
 })
+
+describe('S3ProtocolHandler CopySource decoding', () => {
+  it('url-decodes the copy source key for CopyObject', async () => {
+    const copyObject = vi.fn().mockResolvedValue({
+      eTag: '"etag"',
+      lastModified: new Date('2026-06-25T00:00:00.000Z'),
+    })
+    const storage = {
+      from: vi.fn(() => ({ copyObject })),
+    }
+    const handler = new S3ProtocolHandler(storage as never, 'tenant-id')
+
+    await handler.copyObject({
+      Bucket: 'dest',
+      Key: 'copied.txt',
+      CopySource: 'source-bucket/folder/my%20file%2B1.txt',
+    })
+
+    expect(storage.from).toHaveBeenCalledWith('source-bucket')
+    expect(copyObject).toHaveBeenCalledWith(
+      expect.objectContaining({ sourceKey: 'folder/my file+1.txt' })
+    )
+  })
+
+  it('url-decodes the copy source bucket for CopyObject', async () => {
+    const copyObject = vi.fn().mockResolvedValue({
+      eTag: '"etag"',
+      lastModified: new Date('2026-06-25T00:00:00.000Z'),
+    })
+    const storage = {
+      from: vi.fn(() => ({ copyObject })),
+    }
+    const handler = new S3ProtocolHandler(storage as never, 'tenant-id')
+
+    await handler.copyObject({
+      Bucket: 'dest',
+      Key: 'copied.txt',
+      CopySource: '/my%20bucket/file.txt',
+    })
+
+    expect(storage.from).toHaveBeenCalledWith('my bucket')
+  })
+
+  it('keeps an encoded slash inside the copy source key for CopyObject', async () => {
+    const copyObject = vi.fn().mockResolvedValue({
+      eTag: '"etag"',
+      lastModified: new Date('2026-06-25T00:00:00.000Z'),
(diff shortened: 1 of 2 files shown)


ferhatelmas (changes requested):
Thanks for PR, this mostly looks good. 

Two things: 
1. can we encode outgoing multipart copy source? It's broken now with the parser correction here
2. add an integration that exercises 1's backend (ideally + copyObject)

ferhatelmas (approved):

PR #1382, opened 2026-09-10
fix: ignore media type parameters when checking allowed_mime_types

### LINKED ISSUE #1381: allowed_mime_types rejects uploads whose Content-Type carries a charset parameter
### Bug report

A bucket's `allowed_mime_types` list is compared against the raw `Content-Type` header text, so any media type parameter (`; charset=...`, `; boundary=...`) makes an otherwise allowed upload fail with `invalid_mime_type` / 415.

`validateMimeType` in `src/storage/uploader.ts` splits the header on `/` and compares the halves verbatim:

```ts
const requestedMime = mimeType.split('/')
...
if (allowedType === type && allowedExtension === ext) {
  return true
}
```

For `Content-Type: text/plain;charset=UTF-8` the subtype becomes `plain;charset=UTF-8`, which never equals the allow-list entry `plain`.

**Why this shows up in ordinary use**

supabase-js sends exactly that by default — `DEFAULT_FILE_OPTIONS` in `storage-js/src/packages/StorageFileApi.ts` is `contentType: 'text/plain;charset=UTF-8'`, used whenever a caller uploads a `Blob`/`File`/`FormData` without passing `contentType`. So a bucket restricted to `text/plain` rejects the first-party client's default upload. The same applies to any client sending `application/json; charset=utf-8`, and to the S3 protocol path (`createMultiPartUpload` → `validateMimeType(command.ContentType, ...)`) and the TUS path (`http/routes/tus/lifecycle.ts`), which all share this function.

Per RFC 9110 §8.3, parameters are not part of the media type's identity, and the value stored/served for the object should keep its parameters — only the allow-list comparison should ignore them.

**Reproduction**

Bucket with `allowed_mime_types: ['text/plain']`:

```
POST /object/<bucket>/note.txt
Content-Type: text/plain;charset=UTF-8
```

→ `415 invalid_mime_type: mime type text/plain;charset=UTF-8 is not supported`

Expected: the upload succeeds, and the object keeps `text/plain;charset=UTF-8` as its stored content type.

**Note**

A subtype wildcard entry (`text/*`) happens to work today, because that branch never looks at the subtype — so the failure only affects exact-match allow-list entries, which is the common configuration.

I have a fix with unit-test coverage and will open a PR referencing this issue.

hsusul:
## What

Ignore media type parameters when matching a bucket's `allowed_mime_types`.

Fixes #1381.

## Why

`validateMimeType` compared the raw `Content-Type` header against the allow list:

```ts
const requestedMime = mimeType.split('/')
...
if (allowedType === type && allowedExtension === ext) return true
```

For `text/plain;charset=UTF-8` the subtype half is `plain;charset=UTF-8`, which never equals the allow-list entry `plain`, so the upload fails with `invalid_mime_type` / 415.

This is the default path for the first-party client: `DEFAULT_FILE_OPTIONS` in storage-js is `contentType: 'text/plain;charset=UTF-8'`, used whenever a caller uploads a `Blob`/`File`/`FormData` without passing `contentType`. A bucket restricted to `text/plain` therefore rejects supabase-js's default upload. The S3 protocol path (`createMultiPartUpload`) and the TUS path (`http/routes/tus/lifecycle.ts`) call the same function and were affected identically.

Per [RFC 9110 §8.3](https://www.rfc-editor.org/rfc/rfc9110#section-8.3) parameters are not part of the media type's identity.

## How

A small `mediaTypeOf()` helper strips everything from the first `;` and trims, applied to both the requested type and each allow-list entry (so an entry written as `text/plain;charset=UTF-8` also matches a bare `text/plain`).

The value stored on the object and served back on download is untouched — only the allow-list comparison is normalized.

One extra line: the guard inside the loop re-tested `requestedMime.length`, which had already been rejected above, instead of the `allowedMime` it had just parsed. It is now `allowedMime.length < 2`. Behavior is identical either way (a malformed entry could not match before and cannot match now); it just makes the guard mean what it reads.

## Tests

New `src/storage/uploader.test.ts`:

- exact match, subtype wildcard — unchanged behavior
- `text/plain;charset=UTF-8` matches `text/plain` **(fails on master)**
- `application/json; charset=utf-8` with leading whitespace matches `application/json` **(fails on master)**
- allow-list entry carrying its own parameter matches a bare media type **(fails on master)**
- `text/plain;charset=UTF-8` matches the `text/*` wildcard — guard for the branch that already worked
- a disallowed type with parameters is still rejected
- a malformed media type is still rejected

## Validation

Run locally on macOS / Node v24.9.0:

- `npm run test:unit` — 149 files, 2132 tests, all passing (3 of the new cases fail on clean `master`, verified before the fix)
- `npm run build` (`tsc -noEmit` + bundle) — passing
- `npm run lint` (biome + prettier) — clean
- `git diff --check` — clean

Not run: `npm run test:integration`. It needs the Docker Postgres stack and port 5432 was already occupied on this machine, so I did not start it. The change is confined to one pure, exported function that is covered by the unit tests above.


src/storage/uploader.test.ts
```diff
@@ -0,0 +1,47 @@
...
+function expectInvalidMimeType(fn: () => unknown) {
+  try {
+    fn()
+  } catch (error) {
+    expect(error).toMatchObject({ code: ErrorCode.InvalidMimeType })
+    return
+  }
+
+  throw new Error('expected an invalid mime type error')
+}
```
ferhatelmas: `expect(fn).toThrow(expect.objectContaining({ code }))`

src/storage/uploader.ts
```diff
@@ -302,8 +302,16 @@ export class Uploader {
  * @param mimeType
  * @param allowedMimeTypes
  */
```
ferhatelmas: we need this on the right function, it's moved now

ferhatelmas (approved):

PR #1398, opened 2026-09-16
fix: close connection on pre body error

ferhatelmas:

Bug fix

If a request is rejected before the body is read (bad JWT, missing S3 signature, 404, feature disabled, etc.), response is still keep alive and the advertised body is never drained. A pooling proxy (Kong) reuses that upstream socket and the next request bytes are treated as leftover body, so that next caller hangs until the proxy times out (~60s).

Bodyless errors are fine.

Please link any relevant issues here.

## What is the new behavior?

If the request declared a body that has not fully arrived, the response is closed and the socket is finished after the response and the next request uses a new connection. 

Explicit `shouldCloseConnection` errors still force close even after a consumed body.

Bodyless errors still stay fine.

## Additional context

Related to #1323


ferhatelmas:
@claude review always

fenos (approved):

PR #1403, opened 2026-09-17
fix: clamp to up to date when a tenant's migration is unrecognized

fenos:

Bug fix / deploy-safety hardening for the multitenant migration bookkeeping.

`areMigrationsUpToDate` compares a tenant's recorded migration name against this binary's own `DBMigration` ordinal map (`DBMigration[tenant.migrationVersion]`). If the tenant's recorded migration name isn't a key this binary knows about at all, that lookup is `undefined`, and the ordinal comparison silently evaluates as "the tenant is behind."

This can happen whenever two branches are developed in parallel and each adds a new migration: whichever merges first claims the next number, and during a rolling deploy an older pod can see a tenant a newer pod already migrated past, using a migration name the older pod has never heard of. It isn't hypothetical — it already happened once during development: `0073` was independently claimed by `revoke-grants-to-unused-operations` on `master` and by another migration on a parallel branch, which had to renumber to `0074` to avoid colliding.

When an older pod concludes (incorrectly) that a tenant is "behind", it attempts a migration run it can't reason about, and can overwrite the tenant's recorded migration state with its own, older value.

## What is the new behavior?

`areMigrationsUpToDate` explicitly detects when the tenant's recorded migration name isn't recognized by this binary and clamps to "nothing left for me to run" instead of comparing against `undefined`, logging a warning so the condition is observable. `resolveLatestMigration`'s existing fallback to its own local latest migration in this situation is documented as intentional.

This is meant to ship ahead of any PR introducing a new migration ordinal, so the "old" side of that rollout's mixed-version window already tolerates a migration name it doesn't recognize.

## Additional context

Three new unit tests cover: the unrecognized-name clamp + warning, a tenant genuinely behind a known migration, and a tenant already caught up.


src/http/plugins/db.ts
ferhatelmas, line 164: this snapshot value goes to adapter directly, related to places where hasMigration isn't used

src/internal/database/migrations/migrate.ts
```diff
@@ -93,21 +93,34 @@ export function startAsyncMigrations(signal: AbortSignal) {
...
+
+  // Unrecognized by this binary: a newer version already moved the tenant
+  // past what this one knows. Clamp to this binary's own latest instead of
+  // comparing against undefined, which would otherwise report every
+  // migration - including ones this binary does know about - as not applied
+  // (this is what PROGRESSIVE and other strategies that don't resolve
+  // request.latestMigration through resolveLatestMigration fall back to).
+  const ordinal = Object.hasOwn(DBMigration, migrationVersion)
+    ? DBMigration[migrationVersion]
+    : DBMigration[localLatest]
```
ferhatelmas: Last local respects freeze at. When the recorded migration is unrecognized, this clamps its capabilities to the frozen target even though its schema could already advance further (i.e. deploying after freeze to remove freeze). Could we clamp to the highest migration this binary knows instead?

PR #1404, opened 2026-09-17
fix: improve request local sigv4 signing key cache

ferhatelmas:

Bug fix (perf)

Sigv4 signing key cache does composite string building and hashing.

## What is the new behavior?

Use simple array slots and skip it. 
Construction 10x and lookup 15x are faster.

## Additional context

Related to #1340


src/storage/protocols/s3/signature-v4-stream.test.ts
```diff
@@ -621,6 +621,40 @@ describe('ChunkSignatureV4Parser', () => {
...
   })
 
+  test.each([
+    { name: 'date', changedScope: { shortDate: '20260407' } },
+    { name: 'region', changedScope: { region: 'eu-west-1' } },
+    { name: 'service', changedScope: { service: 's3vectors' } },
+  ])('validates chunks after the signing $name changes and changes back', ({ changedScope }) => {
+    const secretKey = 'secret-key'
+    const initialScope = { shortDate: '20260406', region: 'us-east-1', service: 's3' }
+    const signer = new SignatureV4({
```
itslenny: if you used `TestSignatureV4` here, you could spy on signingKey and assert it's called 3 times to prove the cache actually misses/recomputes on each scope change instead of serving a stale key. Currently the test only checks that validateChunkSignature returns true, which would pass even if the cache were deleted entirely.

itslenny (approved):

src/storage/protocols/s3/signature-v4-stream.test.ts
ferhatelmas: good call, applied

PR #1408, opened 2026-09-17
fix: improve lifecycle log accuracy

itslenny:

Bug fix

The lifecycle event logs used for usage tracking have accuracy issues:

- `ObjectCreated:Put` / `ObjectCreated:Copy` does not include old object size when overwriting an existing object
- Not logged if webhooks are disabled for the tenant causing usage tracking to be inaccurate
- Logs emitted in queue job which can cause duplicate entries if the job is retried

## What is the new behavior?

- `ObjectCreated:Put` / `ObjectCreated:Copy` add an oldObject field with the replaced object's size. This lets usage be correctly attributed.
- All lifecycle events:
  - log once at dispatch time, before the event is queued. This prevents double counted usage.
  - emit lifecycle logs if webhooks are disabled

## Lifecycle log change

This is an example of a lifecycle log diff after this change including `oldObject` if there is an object being overwritten

```diff
{
  "tenant": {
    "ref": "PROJECT_REF"
  },
  "name": "obj-name.txt",
  "version": "UUID",
  "bucketId": "bucket-name",
  "metadata": {
    "httpStatusCode": 200,
    "cacheControl": "no-cache",
    "eTag": "ETAG",
    "mimetype": "text/plain",
    "contentLength": 56,
    "lastModified": "TIMESTAMP",
    "size": 56
  },
  "reqId": "REQID",
  "uploadType": "standard",
+  "oldObject": {
+    "name": "obj-name.txt",
+    "bucketId": "bucket-name",
+    "version": "UUID",
+    "metadata": {
+      "eTag": "ETAG",
+      "size": 30,
+      "mimetype": "text/plain",
+      "cacheControl": "no-cache",
+      "lastModified": "TIMESTAMP",
+      "contentLength": 30,
+      "httpStatusCode": 200
+    },
    "reqId": "REQID"
  }
}
```


src/storage/uploader.ts
```diff
@@ -251,41 +249,57 @@ export class Uploader {
           )
         }
 
-        const event = isUpsert && !isNew ? ObjectCreatedPutEvent : ObjectCreatedPostEvent
+        const oldObject =
+          isUpsert && currentObj
```
ferhatelmas: this condition doesn't protect for retries (tus/multipart), not currentObj or not same version would be fine

ferhatelmas:
two notes:
1. rolling deploy but brief: old producers with new workers will lose some events
2. duplication due to tx rollback. I would suggest revisiting after #1347 (because it has relevant changes) or targeting its head

itslenny:
Closing in favor of opening a new PR built on top of object versioning changes.

PR #1427, opened 2026-09-21
chore: move to rustfs for local dev

TylerHillery:

Chore to replace minio with rustfs for local development.

Currently we use minio

## What is the new behavior?

Now we use rustfs


src/storage/backend/s3/adapter.ts
```diff
@@ -556,7 +552,7 @@ export class S3Backend implements StorageBackendAdapter {
       const data = await this.client.send(command)
       return {
         cacheControl: data.CacheControl || 'no-cache',
-        mimetype: data.ContentType || 'application/octet-stream',
+        mimetype: data.ContentType || 'binary/octet-stream',
         eTag: data.ETag || '',
```
TylerHillery, line 556: The default fallback value for S3 is `binary/octet-stream`. You can verify this against a real S3 bucket:

```bash
echo "hello" > /tmp/s3-src.txt
aws s3api put-object --bucket "$BUCKET" --key "$KEY_SRC" --body /tmp/s3-src.txt
aws s3api head-object --bucket "$BUCKET" --key "$KEY_SRC"
aws s3api copy-object --bucket "$BUCKET" --key "$KEY_DST" \
  --copy-source "$BUCKET/$KEY_SRC" \
  --metadata-directive REPLACE --cache-control "max-age=2009"
aws s3api head-object --bucket "$BUCKET" --key "$KEY_DST"
```

Both `head-object` calls return `binary/octet-stream`

This fallback only fires when the backend returns no `ContentType` at all. MinIO and real AWS S3 always set one, so we never hit our own fallback there but RustFS has a bug where `CopyObject` with `MetadataDirective=REPLACE` doesn't set a content type when none is passed through, which is what surfaced this. I'm working on an upstream fix to RustFS, but the fallback here should be updated, since it's correct behavior.

TylerHillery: here is the upstream fix: https://github.com/rustfs/rustfs/pull/8073

.docker/docker-compose-infra.yml
```diff
@@ -129,38 +129,44 @@ services:
...
     entrypoint: >
       /bin/sh -c "
-      /usr/bin/mc alias set supa-minio http://minio:9000 supa-storage secret1234;
-      /usr/bin/mc mb supa-minio/supa-storage-bucket;
-      /usr/bin/mc mb supa-minio/warehouse--table-s3;
-      /usr/bin/mc policy set public supa-minio/warehouse--table-s3;
-      /usr/bin/mc mb supa-minio/supa-storage-profiles;
+      aws --endpoint-url http://rustfs:9000 s3 mb s3://supa-storage-bucket;
+      aws --endpoint-url http://rustfs:9000 s3 mb s3://warehouse--table-s3;
+      aws --endpoint-url http://rustfs:9000 s3api put-bucket-policy --bucket warehouse--table-s3 --policy '{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Principal\":\"*\",\"Action\":[\"s3:GetObject\",\"s3:PutObject\",\"s3:DeleteObject\",\"s3:ListBucket\"],\"Resource\":[\"arn:aws:s3:::warehouse--table-s3\",\"arn:aws:s3:::warehouse--table-s3/*\"]}]}';
```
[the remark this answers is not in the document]
TylerHillery: MinIO's public policy is read+write, not read-only. mc anonymous set (the current implementation of the old mc policy set) maps public directly to the readwrite IAM policy internally: https://github.com/minio/mc/blob/master/cmd/anonymous-main.go#L236-L243

src/storage/backend/s3/adapter.ts
```diff
@@ -133,7 +133,7 @@ export class S3Backend implements StorageBackendAdapter {
     return {
       metadata: {
         cacheControl: data.CacheControl || 'no-cache',
-        mimetype: data.ContentType || 'application/octet-stream',
+        mimetype: data.ContentType || 'binary/octet-stream',
```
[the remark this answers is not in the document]
TylerHillery: This is what S3 actually defaults to and it only uses this fallback when the backend doesn't provide one. This isn't related to RustFS.

docker-compose-multi-tenant.yml
```diff
@@ -76,17 +76,17 @@ services:
...
 
-  minio_setup:
+  rustfs_setup:
     extends:
-      service: minio_setup
+      service: rustfs_setup
       file: ./.docker/docker-compose-infra.yml
     depends_on:
-      - minio
+      - rustfs
```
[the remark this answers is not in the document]
TylerHillery: fixed

src/storage/backend/s3/adapter.ts
```diff
@@ -137,7 +133,7 @@ export class S3Backend implements StorageBackendAdapter {
...
```
ferhatelmas, line 136: whatever the value, we need consistency between s3 and file backends

.docker/docker-compose-infra.yml
```diff
@@ -129,38 +129,44 @@ services:
...
+      RUSTFS_ADDRESS: ":9000"
+      RUSTFS_CONSOLE_ADDRESS: ":9001"
+      RUSTFS_CONSOLE_ENABLE: "true"
+      RUSTFS_SERVER_DOMAINS: "rustfs"
+    command: /data
 
-  minio_setup:
-    image: minio/mc
+  rustfs_setup:
+    image: amazon/aws-cli
```
ferhatelmas: might be better to pin it

.docker/docker-compose-infra.yml
```diff
@@ -129,38 +129,44 @@ services:
...
-      MINIO_ROOT_PASSWORD: secret1234
-      MINIO_DOMAIN: minio
-    command: server --console-address ":9001" /data
+      RUSTFS_ACCESS_KEY: supa-storage
+      RUSTFS_SECRET_KEY: secret1234
+      RUSTFS_ADDRESS: ":9000"
+      RUSTFS_CONSOLE_ADDRESS: ":9001"
+      RUSTFS_CONSOLE_ENABLE: "true"
+      RUSTFS_SERVER_DOMAINS: "rustfs"
+    command: /data
```
ferhatelmas: seems redundant https://github.com/rustfs/rustfs/blob/d47f54bfb2f39f48bd1adda334bd27e151fe85b8/Dockerfile#L114

.docker/docker-compose-infra.yml
ferhatelmas: another thing is that it's bigger than rustfs itself, would be nicer to use curl to create buckets and not pull this image at all

ferhatelmas:
```rustfs_setup:
    image: rustfs/rustfs:1.0.0
    depends_on:
      rustfs:
        condition: service_healthy
    entrypoint: >
      /bin/sh -c "
      for b in supa-storage-bucket warehouse--table-s3 supa-storage-profiles; do
      curl -fsS --aws-sigv4 aws:amz:us-east-1:s3 -u supa-storage:secret1234 -X PUT http://rustfs:9000/$$b;
      done;
      curl -fsS --aws-sigv4 aws:amz:us-east-1:s3 -u supa-storage:secret1234 -X PUT 'http://rustfs:9000/warehouse--table-s3?policy' -d '{\"Version\":...same policy...}';
      exit 0;
      "
```

.docker/docker-compose-infra.yml
TylerHillery: removed

.docker/docker-compose-infra.yml
TylerHillery: fixed

src/storage/backend/s3/adapter.ts
TylerHillery: updated file backends and found one more spot with multipart uploads that needed to be fixed as well

.docker/docker-compose-infra.yml
```diff
@@ -129,38 +129,39 @@ services:
...
       /bin/sh -c "
-      /usr/bin/mc alias set supa-minio http://minio:9000 supa-storage secret1234;
-      /usr/bin/mc mb supa-minio/supa-storage-bucket;
-      /usr/bin/mc mb supa-minio/warehouse--table-s3;
-      /usr/bin/mc policy set public supa-minio/warehouse--table-s3;
-      /usr/bin/mc mb supa-minio/supa-storage-profiles;
+      for b in supa-storage-bucket warehouse--table-s3 supa-storage-profiles; do
+      curl -fsS --aws-sigv4 'aws:amz:us-east-1:s3' -u supa-storage:secret1234 -X PUT http://rustfs:9000/\$$b;
+      done;
+      curl -fsS --aws-sigv4 'aws:amz:us-east-1:s3' -u supa-storage:secret1234 -X PUT 'http://rustfs:9000/warehouse--table-s3?policy' -d '{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Principal\":\"*\",\"Action\":[\"s3:GetObject\",\"s3:PutObject\",\"s3:DeleteObject\",\"s3:ListBucket\"],\"Resource\":[\"arn:aws:s3:::warehouse--table-s3\",\"arn:aws:s3:::warehouse--table-s3/*\"]}]}';
```
ferhatelmas, line 164: is this policy needed?

src/storage/backend/file.ts
```diff
@@ -371,7 +368,7 @@ export class FileBackend implements StorageBackendAdapter {
       httpStatusCode: 200,
       size: data.size,
       cacheControl: cacheControl || 'no-cache',
-      mimetype: contentType || 'application/octet-stream',
+      mimetype: contentType || 'binary/octet-stream',
```
ferhatelmas, line 371: seems like upload case is missing

ferhatelmas: and it also means we need a small test becase there is no test update which is missing

